PR #27, opened 2014-10-12
Fix thread synchronization bug

Livven:
In the `Pool.ToPool` extension method, `StringBuilder` objects were being pushed to the common pool before their results were retrieved outside the lock. This could lead to `Clear` and `ToString` being called on the same `StringBuilder` object across two different threads (by `Pool.NewStringBuilder` and `Pool.ToPool`, respectively), resulting in a corrupted return value containing a bunch of `\0` characters.


FlorianRappl:
Yep could happen. Good catch there!

PR #41, opened 2014-11-27
Element ToHtml method should be standard when it is selfclosing.

h82258652:
before ToHtml example:
&lt;p&gt;&lt;img src="xxx"&gt;&lt;/p&gt;

after ToHtml example:
&lt;p&gt;&lt;img src="xxx"<span style="color:red;">/<span>&gt;&lt;/p&gt;

I think now will better.

### DIFF
--- a/AngleSharp/DOM/Element.cs
+++ b/AngleSharp/DOM/Element.cs
@@ -677,9 +677,15 @@
             foreach (var attribute in _attributes)
                 sb.Append(Specification.Space).Append(attribute.ToString());
 
+            bool isSelfClosing = Flags.HasFlag(NodeFlags.SelfClosing);
+            if (isSelfClosing)
+            {
+                sb.Append(Specification.Solidus);
+            }
+
             sb.Append(Specification.GreaterThan);
 
-            if (!Flags.HasFlag(NodeFlags.SelfClosing))
+            if (!isSelfClosing)
             {
                 if (Flags.HasFlag(NodeFlags.LineTolerance) && FirstChild is IText)
                 {


FlorianRappl:
Unfortunately you are not right. This is not XML. Elements that are self-closing by nature (e.g. `meta`, `img` and others) are serialized like this (compare behavior in browsers).

See specification: http://www.w3.org/html/wg/drafts/html/master/syntax.html#serializing-html-fragments
Note "If current node is an area, base, basefont, bgsound, br, col, embed, frame, hr, img, input, keygen, link, menuitem, meta, param, source, track or wbr element, then continue on to the next child node at this point.". This means that no solidus (or content) is being emitted.

h82258652:
Thanks for response. I'm finding a method to convert a normal html document to a xhtml document.
By the way, do you know any library to do this in WinRT? I search on nuget, but I can not found.

FlorianRappl:
You mean for HTML to XHTML? Hm, I don't know. AngleSharp also works on WinRT so writing an extension that takes the DOM and serializes it with XML conformance is certainly possible.

Maybe I should introduce another method called `ToXml()`, which would work similar to `ToHtml()`, however, providing XML conformance.

h82258652:
Yes, I mean Html to XHtml. I have try to code extension methods for the nodes, Just like this:

<pre><code>
public static class XHtmlExtension
{
    public static string ToXHtml(this IDocument document)
    {
        return document.ChildNodes.ToXHtml();
    }
    public static string ToXHtml(this INodeList nodeList)
    {
        var sb = Pool.NewStringBuilder();
        foreach (var node in nodeList)
        {
            // as you know, no matter what element it is,
            // it always call ToXHtml(this INode node).
            sb.Append(node.ToXHtml());
        }
        return sb.ToPool();
    }
    public static string ToXHtml(this INode node)
    {
        // always call this method from
        // ToXHtml(this INodeList nodeList).
        // TODO
        // Here is hard to code.
    }
    public static string ToXHtml(this IComment comment)
    {
        // never called from
        // ToXHtml(this INodeList nodeList).
        if(comment.Data.Contains("-"))
        {
            // In XHtml, comment could not contains '-'
            throw new DOMException(ErrorCode.XXXX);
        }
        return comment.ToHtml();
    }
    // Other extension methods.
}
</code></pre>

`ToXml()` or `ToXHtml()` is both OK.
or overload the `ToHtml()` method is a good way too.

<pre><code>
public string ToHtml()
{
    return ToHtml(XXXConfiguration.Default);
}
public string ToHtml(XXXConfiguration config)
{
    // TODO
}
</code></pre>

PR #57, opened 2015-02-20
Make the methods of HtmlMarkupFormatter virtual

thorn0:
to make it easier to inherit from this class


FlorianRappl:
I thought about it, too, but I did not do that on purpose. Here composition should be favored over inheritance.

PR #60, opened 2015-02-20
Add the .editorconfig file to enforce the coding style

thorn0:
see http://editorconfig.org


FlorianRappl:
Yes, should be a good help for people using the plugin with their favorite IDE.

PR #117, opened 2015-07-25
Changed element attributes collection to be a NamedNodeMap as per spec

miroslav22:
Hi,

I've changed the structure of element attributes to be a new collection 'NamedNodeMap' rather than a simple list.

With this change (and a couple more in the scripting library) I've managed to get jQuery to parse and execute over the DOM which is pretty cool!

I've modified the unit tests as well to account for this change


AngleSharp/Dom/Collections/NamedNodeMap.cs
```diff
@@ -0,0 +1,85 @@
...
+        
+        private readonly List<IAttr> _items = new List<IAttr>();
+
+        #endregion
+
+        #region Index
+
+        public IAttr this[string name]
+        {
+            get { return _items.FirstOrDefault(i => i.Name == name); }
```
FlorianRappl, line 24: This may be used a lot, so using a `for` loop instead of LINQ will be better for performance reasons.

AngleSharp/Dom/Collections/NamedNodeMap.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return _items.GetEnumerator();
+        }
+
+        public IAttr GetNamedItemNS(string namespaceUri, string localName)
+        {
+            return _items.FirstOrDefault(attribute => attribute.NamespaceUri == namespaceUri && attribute.LocalName == localName);
```
FlorianRappl, line 69: Same reasoning as above here. LINQ should be replaced with a simple `for` loop.

AngleSharp/Dom/Collections/NamedNodeMap.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void SetNamedItem(IAttr item)
+        {
+            RemoveNamedItem(item.Name);
+            _items.Add(item);
+        }
+
+        public void RemoveNamedItem(string name)
+        {
+            _items.RemoveAll(i => i.Name == name);
```
FlorianRappl, line 54: Definitely not as severe as the other two issues here, but since we know `Name` is unique a `for` loop that breaks after the condition has been met and the the entry has been removed, could be better.

AngleSharp/Dom/Element.cs
```diff
@@ -582,16 +584,16 @@ public void SetAttribute(String name, String value)
                 if (_namespace == Namespaces.HtmlUri)
                     name = name.ToLowerInvariant();
 
-                for (int i = 0; i < _attributes.Count; i++)
-                {
-                    if (_attributes[i].Prefix == null && _attributes[i].LocalName == name)
+                foreach(var attribute in _attributes)
```
FlorianRappl, line 587: Could be moved to `NameNodeMap`. Especially the `AttributeChanged` is essential (see total discussion).

AngleSharp/Extensions/AttrExtensions.cs
```diff
@@ -20,15 +23,9 @@ static class AttrExtensions
...
+        public static Boolean Has(this INamedNodeMap attributes, String name)
         {
-            for (int i = 0; i < attributes.Count; i++)
-            {
-                if (attributes[i].Name == name)
-                    return true;
-            }
-
-            return false;
+            return attributes.Any(attribute => attribute.Name == name);
```
FlorianRappl, line 28: As far as I remember very crucial for perf to keep this as lightweight as possible. So LINQ should definitely be removed here.

AngleSharp/Interfaces/INamedNodeMap.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using AngleSharp.Attributes;
+using AngleSharp.Dom;
+
+namespace AngleSharp.Interfaces
```
FlorianRappl, line 6: These interfaces live in the same namespace as the implementations, `AngleSharp.Dom`. They are only placed in a different folder for overview / organization reasons.

FlorianRappl:
I think those are great changes, but some things need definitely to be cleaned up. I will do that actually.

FlorianRappl:
One thing that definitely needs to be integrated is attribute changing. Beforehand it was guaranteed that any change to attributes will propagate to trigger an event. Now that can be circumvented by using the `NamedNodeMap`.

FlorianRappl:
Also if `IAttr` becomes that exposed again, I feel like `Document` requires `createAttribute`. Otherwise how can `IAttr` instances be created by users?

FlorianRappl:
I will therefore implement `createAttribute` as seen in http://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html#ID-1084891198.

miroslav22:
Hi,

Many thanks for the feedback!  If you want me to look at implementing triggering the event when an attribute is changed I can do that.  Let me know if so.

FlorianRappl:
Thanks a lot for your efforts. I think I've implemented all the required steps, but what is missing are tests for the `NamedNodeMap`. Maybe you can supply a few. Right now there are already some tests to cover that changing the attribute triggers an update. But all those use the API on the `Element`, not the `NamedNodeMap`.

So the following tests would be wonderful:
- Getting an attribute via `Attributes` (i.e. `NamedNodeMap`)
- Adding an attribute via `Attributes` (i.e. `NamedNodeMap`) [change should be triggered]
- Removing an attribute via `Attributes` (i.e. `NamedNodeMap`) [change should be triggered]
- Changing an attribute via `Attributes` (i.e. `NamedNodeMap`) [change should be triggered]

FlorianRappl:
(Maybe you also know some more tests that need to be passed or would be worth writing)

miroslav22:
Hi, yes no probs, I'll add in some unit tests over the next few days.  Thanks :)

FlorianRappl:
Wonderful! Much appreciated :+1:

Issue #129, opened 2015-08-16
Split the creation and execution of the form DocumentRequest


joelverhagen:
Currently `HtmlFormElement.Submit()` creates a `DocumentRequest` and submits it to the underlying `IDocumentLoader` via `BrowsingContextExtensions.OpenAsync(...)` as two consecutive steps.

For my use case, it is beneficial to split these two distinct actions into two methods. I would like to serialize and persist the intermediate `DocumentRequest` to allow deferred navigation. For example, imagine the submission of a form as the continuation to a subsequent web page. This `DocumentRequest` instance can be considered as a continuation token. There are some small details like restoring cookie state that are out of scope for the form itself.

Pull request forthcoming.

FlorianRappl:
Yes, I guess this definitely makes sense. I'll have to think about it (and potential implications) in more detail, but generally speaking, making the whole process easier to extend and therefore providing several smaller steps is certainly beneficial. 

I'll look at the PR right away and feedback there! Thanks already for the good thoughts and enhancements.

PR #130, opened 2015-08-16
Add GetSubmission to IHtmlFormElement

joelverhagen:
1. Add `GetSubmission` to `IHtmlFormElement` and associated implementation.
2. Add `Headers` to `DocumentRequest`.
3. Add potentially missing logical negation operator `!NoValidation` (@FlorianRappl, can you confirm)?


joelverhagen:
#129

FlorianRappl:
In general I think this is great. It makes it possible to use the form submission more easily with another (possibly non-registered) requester. Or to read the contents of the request more easily.

The only problem I have / see are the changes to the `DocumentRequest`. This is now quite close to the `Request`. I wonder if these changes are all necessary here. I'll need to have a closer look.

Issue #141, opened 2015-09-02
CSS Unicode escapes not properly parsed


mganss:
Whitespace other than a single space after Unicode escapes in CSS is parsed incorrectly. Example:

``` C#
var html = @"<DIV STYLE='background: \000075
rl(javascript:alert(""foo""))'>";
var parser = new HtmlParser(new Configuration().WithCss(e => e.Options = new CssParserOptions
{
    IsIncludingUnknownDeclarations = true,
    IsIncludingUnknownRules = true,
    IsToleratingInvalidConstraints = true,
    IsToleratingInvalidValues = true
}));
var dom = parser.Parse(html);
var div = (IHtmlElement)dom.QuerySelector("div");
var bg = div.Style.First().Value; // "u\nrl(...)", should be "url(...)"
```

The relevant specification is at http://www.w3.org/TR/CSS2/syndata.html#escaped-characters
The code is at https://github.com/AngleSharp/AngleSharp/blob/0f56203a4ee28b3ce7dffaf9d8a5891daf91a243/AngleSharp/Parser/Css/CssTokenizer.cs#L1472-L1473

I can submit a PR if you want.

FlorianRappl:
Yes, that would be welcome (but please to the `devel` branch). Instead of the comparison with a space, I recommend you to use the `IsSpaceCharacter` char extension method. It does the right thing according to the spec (space, tab, lf, cr and ff are covered).

PR #142, opened 2015-09-02
Fix AngleSharp/AngleSharp#141

mganss:


AngleSharp/Parser/Css/CssTokenizer.cs
```diff
@@ -1469,12 +1469,17 @@ String ConsumeEscape(Char current)
                         break;
                 }
 
-                if (current != Symbols.Space)
+                if (current == Symbols.CarriageReturn)
```
FlorianRappl, line 1472: This branch will never be taken, since the base tokenizer will already normalize the input stream (transforming `\r\n` to `\n`). I'll fix it.

FlorianRappl:
Very well done - thanks!

mganss:
Sorry, I missed the part about the \r\n. Thanks for merging so quickly.

FlorianRappl:
No problem at all - its not obvious. It's an optimization that made my life easier.

mganss:
So I believe my code was not only too complicated, it was also buggy, because it consumed too much whitespace, in particular `\r\r\n`. But the current code seems to do that as well, e.g. `\\000062\r\r\nlack` results in "black", too. Is that correct? Shouldn't only the first `\r` be consumed?

FlorianRappl:
The tokenizer skips all `\r` characters. It is an assumption that is debatable, but considering that `\r` is non-printable and that its only purpose is in conjunction with `\n` I feel that it is alright.

Is this a theoretical concern or does it have any practical consequences?

mganss:
Purely theoretical so far :smile: But if this were in a browser context, I believe you could craft input that's harmless according to spec but is turned into an exploit by the parser, e.g. `background-image: \u000075\r\r\nrl(javascript:...)`

FlorianRappl:
Hm I think allowing javascript urls in CSS is much more of an issue. I do see your point, however. I'll try to improve the situation.

FlorianRappl:
I experimented a little bit and it seems that browsers treat sequences such as `\r\r\n` as `\n\n` and `\r\r\r\n` as `\n\n\n`. Hence `\r\n` is transformed to `\n` and `\r` is transformed to `\n`.

I guess I will change the tokenizer to show the same behavior.

mganss:
Excellent :+1:

PR #167, opened 2015-10-01
Fix Node.CompareDocumentPosition

Livven:
`Node.CompareDocumentPosition` didn't handle nested elements correctly. Example:

```
<div id='parent-1'>
    <span id='child-1'>1</span>
</div>
<div id='parent-2'>
    <span id='child-2'>2</span>
</div>
```

Both `child1.CompareDocumentPosition(child2)` and `child2.CompareDocumentPosition(child1)` would result in `DocumentPositions.Following`.

I added a few test cases as well though they're probably not comprehensive.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -225,18 +225,15 @@ If you have an idea how to improve the API (or what is missing) then posts / mes
 
 The project is always searching for additional contributors. Even if you do not have any code to contribute, but rather an idea for improvement, a bug report or a mistake in the documentation. These are the contributions that keep this project active. 
 
-Some legal stuff
-----------------
+License
+-------
 
-Copyright (c) 2013-2015, Florian Rappl and collaborators.
-All rights reserved.
+The MIT License (MIT)
 
-Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


FlorianRappl:
Hi Livven looks like a great change, but I don't accept PR to master any more. Can you please PR into an appropriate feature branch? 

It is all outlined in the [Wiki](https://github.com/AngleSharp/AngleSharp/wiki/Contributing). Thanks!

Livven:
Thanks, I created a new pull request (#168) on another branch.

BTW I don't think Github lets you assign yourself to issues unless you're a collaborator, as suggested in the [development workflow](https://github.com/AngleSharp/AngleSharp/wiki/Contributing):

> If no issue already exists for the work you'll be doing, create one to document the problem(s) being solved and self-assign.

FlorianRappl:
Yes that is true, but is completely unrelated. Assigning somebody to an issue and really working / committing are still two different things (one is just organizational / GitHub related, while the other is the actual work / git related).

Until this point in time no one asked / wanted to be a collaborator (even though many people post issues / PR on a more regular basis). So if it would be more convenient for you with the "collaborator" status I can add you to the team.

PR #168, opened 2015-10-01
Fix Node.CompareDocumentPosition

Livven:
Same as #167 but on a different branch.

---

`Node.CompareDocumentPosition` didn't handle nested elements correctly. Example:

```
<div id='parent-1'>
    <span id='child-1'>1</span>
</div>
<div id='parent-2'>
    <span id='child-2'>2</span>
</div>
```

Both `child1.CompareDocumentPosition(child2)` and `child2.CompareDocumentPosition(child1)` would result in `DocumentPositions.Following`.

I added a few test cases as well though they're probably not comprehensive.

### DIFF
--- a/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -113,6 +113,7 @@
     <Compile Include="Html\Adoption.cs" />
     <Compile Include="Html\BasicRuleSet.cs" />
     <Compile Include="Html\CData.cs" />
+    <Compile Include="Html\CompareDocumentPosition.cs" />
     <Compile Include="Html\Doctype.cs" />
     <Compile Include="Html\DomManipulation.cs" />
     <Compile Include="Html\HtmlConstruction.cs" />
--- a/AngleSharp.Core.Tests/Html/CompareDocumentPosition.cs
+++ b/AngleSharp.Core.Tests/Html/CompareDocumentPosition.cs
@@ -1,0 +1,56 @@
+﻿using AngleSharp.Dom;
+using NUnit.Framework;
+
+namespace AngleSharp.Core.Tests
+{
+    [TestFixture]
+    public class CompareDocumentPositionTests
+    {
+        IDocument doc;
+
+        IElement parent1, parent2, child1, child2;
+
+        [SetUp]
+        public void SetUp()
+        {
+            doc = "<!DOCTYPE html><html><head><title>Title</title></head><body><div id='parent-1'><span id='child-1'>1</span></div><div id='parent-2'><span id='child-2'>2</span></div></body></html>".ToHtmlDocument();
+
+            parent1 = doc.QuerySelector("#parent-1");
+            parent2 = doc.QuerySelector("#parent-2");
+            child1 = doc.QuerySelector("#child-1");
+            child2 = doc.QuerySelector("#child-2");
+
+        }
+
+        [Test]
+        public void SameParent()
+        {
+            Assert.AreEqual(DocumentPositions.Following, parent1.CompareDocumentPosition(parent2), "Parent 2 follows parent 1.");
+            Assert.AreEqual(DocumentPositions.Preceding, parent2.CompareDocumentPosition(parent1), "Parent 1 precedes parent 2.");
+        }
+
+        [Test]
+        public void NestedChildren()
+        {
+            Assert.AreEqual(DocumentPositions.Following, child1.CompareDocumentPosition(child2), "Child 2 follows child 1.");
+            Assert.AreEqual(DocumentPositions.Preceding, child2.CompareDocumentPosition(child1), "Child 1 precedes child 2.");
+        }
+
+        [Test]
+        public void Containing()
+        {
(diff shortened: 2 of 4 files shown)


FlorianRappl:
I don't want to be pedantic, but I would highly prefer if a new branch is chosen. That allows me to have a closer look without having it in the "currently ready for next release" (devel) branch. In the current model master is always pointing to the last release, while devel has everything currently set for the next release.

For your implementation, which specification did you pick? Did you choose https://dom.spec.whatwg.org/#dom-node-comparedocumentposition?

Livven:
Sorry, I'm not familiar with the workflow here but I don't think I can create new branches in your repo. I just followed your guidelines which say to "submit a pull request against `devel`". If you want to check out the PR locally it seems you should be able to use the following command:

```
git pull origin pull/168/head
```

---

I didn't follow or even know about the specification, so thanks for the link. As I said, my tests are probably not comprehensive so my implementation may not be 100% perfect, however the existing implementation, specifically the `NodeExtensions.IsPreceding` method, is completely wrong for two reasons:
1. The recursion via `return parent.IsPreceding(after)` only walks up the ancestor tree of the `before` element, so unless `after` or `after.Parent` contain `before` the method will always reach `parent == null` and thus return `false`.
2. The condition `parent == after` should return `false` instead of `true`, though this case was already handled correctly in `Node.CompareDocumentPosition` via `otherNode.IsDescendantOf(this)` so it didn't matter.

---

In the end, if you correctly check for precedence (step 6 from the linked spec) you will inevitably also satisfy all the other steps, so calling a dedicated `IsPreceding` method from `CompareDocumentPosition` doesn't really make sense.

In my implementation, the private static `Node.ComparePosition` method should satisfy all the constraints outlined in the spec, even if it uses a different algorithm to arrive at the same result. `Node.CompareDocumentPosition` merely adds a few shortcut checks to hopefully improve performance.

FlorianRappl:
I see - well, I guess its alright. Usually the flow would have been to notify about the issue in advance (by creating an issue), thus allowing a new branch to be created. On the other side if the issue is already handled then submitting the PR to `devel` (as outlined in the guidelines) is definitely alright. In this case, however, there are still some open questions.

While I agree on the general issue I have problems with the following:
- Why remove the `IsPreceding` method instead of fixing it? I don't think it would be a huge fix. I think the method outlined there could / should have worked. Obviously I never tested it and it doesn't, but it should not be off by far.
- Why not introduce the new functions as extension methods? If applicable (i.e., no `private` information required) that should be preferred. Makes the function much more usable in the future. Granted, currently it is only used within a `Node`, but following SOLID principles we shouldn't be required to touch this in the future.

Also I would love to have more tests. Of course you don't need to provide them, but if we do it right, we should do it as careful as possible.

Livven:
Well if you'd let me know how to do it then I'd be happy to, I just haven't [found](https://www.google.com/#q=github+pull+request+to+new+branch) any way myself.

EDIT: I see you clarified it. I worked around the bug in my app instead of creating an issue and waiting for someone else to fix it and a new NuGet package to come out. Unfortunate ;)

---
1. You answered your own question. Now, you could replace all instances of `return DocumentPositions.Preceding` with `return true` and everything else with `return false`, but that would simply make the code harder to understand without adding comments, for no benefit at all.
2. Why not make everything extension methods? Your previous `Node.CompareDocumentPosition` method was not an extension method. I don't care either way, I just went with the most restrictive accessibility.

Sure I'd like more tests too, but right now there are none.

FlorianRappl:
I think you confuse things. Of course  `CompareDocumentPosition` (and a lot more from the DOM) would be nice in extension method form, but (and please read them yourself) since AngleSharp tries to follow the W3C we need to follow the given interface definitions. Additionally this allows scripting engines (such as the one in _AngleSharp.Scripting.JavaScript_) to just take the classes / interfaces and create everything on the fly. With extension methods it would be much harder to read and execute.

However, for auxiliary methods we can follow best practices and provide extension methods. This way we can create a nicely decoupled, easy to read, and flexible code.

Livven:
I meant so you could do something like this:

```
public DocumentPositions CompareDocumentPosition(INode otherNode)
{
    // call an extension method which contains all the logic (and would need a different name)
    this.CompareDocumentPosition(otherNode);
}
```

The new private static methods aren't really useful in any other context which is why I made them private. Again, I don't really care, I just want to see the bug fixed. If you want to create a new branch go ahead and I can open another PR and then you can change my code however you see fit before merging it to devel, or tell me exactly what you want changed and I guess I can do that as well.

Or even come up with your own fix. As long as it gets fixed, I'm happy.

FlorianRappl:
Ah yes, actually that would also be a good option.

Yes the behavior will definitely be fixed with the next release. I'll leave that PR open for the moment, but I'll either merge it unchanged or I'll take the unit tests and make them work with the current approach. Either way these unit tests (and others that I will come up with) will all pass in v0.9.3.

Thanks for your contribution! Much appreciated.

FlorianRappl:
Alright - so I picked the tests from the PR, added more tests, and improved / fixed the existing `IsPreceding` implementation. Yes, it is true that an implementation as yours would have been better (as all cases are handled in a single implementation instead of having areas that overlap) - but I wanted to preserve the algorithm and I wanted to provide a working `IsPreceding` method.

All tests are good and I also tested with JavaScript against the [test page](http://w3c-test.org/dom/nodes/Node-compareDocumentPosition.html). Seems fine.

PR #219, opened 2015-12-29
Form.SetFieldValues & Form.Submit

YehudahA:
I created a new method for set field values, including select fields.
I changed the old Form.Submit method to use the new method for set values.


FlorianRappl:
PRs will only be accepted to `devel`. Please resubmit!

PR #220, opened 2015-12-29
Form.SerFieldValues method

YehudahA:
Add Form.SerFieldValues method, and change Form.Submit to use this method.
This method set input and select elements.

### DIFF
--- a/AngleSharp/Extensions/ApiExtensions.cs
+++ b/AngleSharp/Extensions/ApiExtensions.cs
@@ -3,6 +3,7 @@
     using AngleSharp.Dom;
     using AngleSharp.Dom.Events;
     using AngleSharp.Dom.Html;
+    using AngleSharp.Html;
     using System;
     using System.Collections.Generic;
     using System.Linq;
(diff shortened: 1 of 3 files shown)


AngleSharp/Extensions/ApiExtensions.cs
```diff
@@ -27,45 +28,35 @@ public static TElement CreateElement<TElement>(this IDocument document)
             where TElement : IElement
         {
             if (document == null)
-            {
```
FlorianRappl, line 30: Do not delete these braces. Single statements won't be allowed in the whole codebase and every part that still uses single-statements is currently changed.

AngleSharp/Extensions/ApiExtensions.cs
```diff
@@ -27,45 +28,35 @@ public static TElement CreateElement<TElement>(this IDocument document)
...
-                    {
-                        var element = (TElement)obj;
-                        var baseElement = element as Element;
+                for (var i = 0; i < parameters.Length; i++)
+                {
+                    var isDocument = parameters[i].ParameterType == typeof(Document);
+                    arguments[i] = isDocument ? document : parameters[i].DefaultValue;
+                }
 
-                        if (baseElement != null)
```
FlorianRappl, line 61: From which version did you fork? Deleting these lines will result in failing tests. I recommend you to fork from the latest version and the prefer the changeset from the upstream when merging.

FlorianRappl:
Couple of things that need to be fixed before this can be accepted:
- Use the latest version from the upstream. Your changes would remove a lot of necessary changes that should stay in the codebase.
- Please provide unit tests. These tests need to cover the case(s) that can't be handled with the current codebase, however, can be handled with your changes.

YehudahA:
@FlorianRappl I'm sorry, you're absolutely right.
I re-commited my changes.

FlorianRappl:
Looks much better - thanks. Could you also supply a few unit tests for that? Otherwise, the issue you mentioned won't be tested and there could be regression on this feature in the future.

Much appreciated! (sorry to be so picky :stuck_out_tongue_winking_eye:)

YehudahA:
@FlorianRappl, I will do it.
But I don't find any unit test for the original form.submit(fields) method :(

FlorianRappl:
One more reason to supply unit tests now :wink:.

YehudahA:
- I changed the code to support radio inputs [set input.IsChecked instead of set Input.Value]
- I wrote some unit tests.

FlorianRappl:
Very nice. Will be merged after v0.9.4 has been released.

PR #221, opened 2015-12-31
Fixed invalid range

temporaryunknown:


FlorianRappl:
Two things:
- Justification (e.g., specification why the current one is wrong)
- Please submit PR into `devel`

I think its not possible to modify the target. Hence I'll have to close this PR. I would appreciate if you resubmit a PR (into `devel`) and write a short description. Thanks a lot!

temporaryunknown:
Excuse me, please, I forgot to change branch. Fixed.

PR #222, opened 2015-12-31
Fixed invalid range

temporaryunknown:
Two values were missed: 0x8FFFF, 0x9FFFE (but 0x8FFFE and 0x9FFFF were doubled).


FlorianRappl:
Yeah, you are right! Thanks a lot for catching this. Very much appreciated :+1:!

temporaryunknown:
You are always welcome, I like AngleSharp!

PR #254, opened 2016-01-29
Support Bottom CSS Property

jogibear9988:


FlorianRappl:
Unfortunately, GitHub does not support changing the target for a PR. As only PRs to `devel` are accepted I'll have to close that. Please re-do against `devel` and feel free to also include the property for CSS grid. Thanks!

PR #264, opened 2016-02-03
Fix HttpRequester threading issues

Suchiman:
_Locking on this is futile_ if the object to be protected has static scope.
Also increases efficiency through double check locking


FlorianRappl:
Seems fine, but I would appreciate if you could change `var property;` to `var property = default(PropertyInfo);` (just for unified styling reasons). Much appreciated!

Suchiman:
@FlorianRappl adjusted

FlorianRappl:
Perfect! Much appreciated. :+1:

PR #283, opened 2016-03-14
Add .NET Core support.

### LINKED ISSUE #271: Include dotnet target in NuGet
As discussed in #270 we should support CoreCLR via NuGet.

This means duplicating the PCL target within a new folder called _\lib\dotnet_.

jeremymeng:
Fixes #271. 
- Add .NET Core support by duplicating PCL profile 259 output to dotnet folder.

~~\- Use NuSpec.ReferenceGenerator to manage the dependencies for dotnet group.~~

### DIFF
--- a/Nuget/AngleSharp.nuspec
+++ b/Nuget/AngleSharp.nuspec
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="utf-8"?>
 <package>
   <metadata>
     <id>AngleSharp</id>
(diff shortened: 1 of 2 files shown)


jeremymeng:
@FlorianRappl Please take a look.  Let me know if you don't want to add the `NuSpec.ReferenceGenerator` to the project.  I can remove it, and only include changes to the dependency group, the version number bump, and the `Package.ps1`.

I've tested it using a dnx console application project. It is working as expected.

FlorianRappl:
All in all this looks fine. Two things though:
- Is this reference really required? If it is not, I would not add it. I consider this a tool, hence it should be downloaded / installed via, e.g., chocolatey. In Node.js there are two different kinds of dependencies: development and runtime dependencies. In .NET no such difference has been made, which makes it hard to mark this dependency as a development dependency.
- Any reason for the version change? Usually, all project specific commits go into `devel` until a release point is used. If only the package script and the NuGet spec had been changed a PR into `master` would be fine, otherwise a PR into `devel` is needed for consistency.

Thanks for your efforts - much appreciated! :+1:

jeremymeng:
> Is this reference really required?

Yes I agree that it is a tool and not required in the project.  It is probably designed in this way so that people can have the nuspec updated automatically when they develop and new dependencies are introduced.  Not that much useful for stable project.  I remove it.

> Any reason for the version change?

I had thought that you might want a version bump with the newly added .NET Core support.  Anyway I revert the version change. 

Thanks for reviewing!

FlorianRappl:
Perfect, thanks a lot for your contribution! Awesome work :ok_hand:

jeremymeng:
Please let us know when a new NuGet Package is available.  Thanks!

FlorianRappl:
Hi @jeremymeng - thanks again for your support. I just wanted to notify you that AngleSharp v0.9.5 has been released yesterday; and of course your changes have been included!

jeremymeng:
Thank you @FlorianRappl !
/cc @jchannon @thecodejunkie

jchannon:
Awesome!!

Issue #300, opened 2016-03-24
Pooled instances have the possibility of conflicting with other instances


TheCloudlessSky:
There are a few places there instances created from `Pool` live longer than they should. They can be used after calling `ToPool()`. This causes the instances to be released to the pool, but it could still be used when someone else has allocated it from the pool.

For example, `TextSource.Disposed()` could be called more than once which causes _other_ `TextSource` instances to conflict. Here's a sample test (obviously calling `Dispose` twice shouldn't be done, but it illustrates the problem):

``` csharp
[Test]
public void failing_when_disposed_more_than_once()
{
    var parser = new HtmlParser();
    var document1 = parser.Parse("<!doctype><html><head></head><body><h1>Hello World</h1></body></html>");
    Assert.Equal("<!DOCTYPE ><html><head></head><body><h1>Hello World</h1></body></html>", document1.ToHtml());
    document1.Dispose();
    document1.Dispose();

    var document2 = parser.Parse("<!doctype><html><head></head><body><h1>Hello World</h1></body></html>");
    Assert.Equal("<!DOCTYPE ><html><head></head><body><h1>Hello World</h1></body></html>", document2.ToHtml());
    document2.Dispose();
    document2.Dispose();
}
```

Therefore, I suggest anything created from the `Pool` class be wrapped in some sort of pool wrapper. Accessing the instance will throw/be a no-op when accessing the instance after it's been disposed/released to the pool.

The alternative is manually managing the instances returned from the pool (setting them to null) which is similar to the dangers of manual memory management :cry:.

I have a couple of implementation options that I'd be glad to submit as a PR:
1. Never allow accessing the _real_ instance (e.g. `PooledStringBuilder`). Create proxy methods on the pooled class. This is the safest guarantee that instances won't be accidentally shared and there is significantly less possibility of dangling references later.
2. Create some sort of `Pooled<T>` wrapper. This has the disadvantage that consumer's could call `GetInstance()` and re-use that instance:
   
   ``` csharp
   class Pooled<T> : IDisposable
   {
       private bool disposed;
       private T instance;
       private readonly Action<T> release;
   
       public T GetInstance()
       {
           return instance;
       }
   
       public Pooled(T instance, Action<T> release)
       {
           this.disposed = false;
           this.instance = instance;
           this.release = release;
       }
   
       public void Dispose()
       {
           if (disposed) return;
   
           disposed = true;
           var localInstance = instance;
           instance = null;
           release(localInstance);
       }
[...]

PR #309, opened 2016-03-28
Pool fixes for #300

TheCloudlessSky:
As per #300:
- [x] Change `BaseTokenizer` and `TextSource` to have a mutable `StringBuilder` and use it to check if the class is disposed.
- [x] Change `CssBuilder` to not use pooled resources after they've been released to the pool.

In order to facilitate making the `StringBuilder` mutable, I figured it'd make more sense to make it a private settable property since it was `protected`. This required renaming the variable to `StringBuffer`). Site note: I suggest changing the `_events` variable to a property.

Let me know what you'd like to do with the `...State` classes and I'll be glad to either add the changes to this PR or submit a new one.

### DIFF
--- a/AngleSharp/Parser/BaseTokenizer.cs
+++ b/AngleSharp/Parser/BaseTokenizer.cs
@@ -15,7 +15,6 @@
     {
         #region Fields
 
-        protected readonly StringBuilder _stringBuffer;
         protected readonly IEventAggregator _events;
         readonly Stack<UInt16> _columns;
         readonly TextSource _source;
@@ -30,7 +29,7 @@
 
         public BaseTokenizer(TextSource source, IEventAggregator events)
         {
-            _stringBuffer = Pool.NewStringBuilder();
+            StringBuffer = Pool.NewStringBuilder();
             _events = events;
             _columns = new Stack<UInt16>();
             _source = source;
@@ -42,6 +41,8 @@
         #endregion
 
         #region Properties
+
+        protected StringBuilder StringBuffer { get; private set; }
 
         public TextSource Source
         {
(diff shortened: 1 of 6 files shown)


FlorianRappl:
Did you measure the perf? Is there any loss that the pooled `StringBuilder` is now accessed via a property? Usually, the compiler should optimize the trivial function call away, but I always like to confirm this myself.

TheCloudlessSky:
@FlorianRappl Automatic properties should have negligible performance because [they should be inlined by the JIT compiler](http://stackoverflow.com/questions/646779/does-c-sharp-inline-properties).

But, just in case...

``` csharp
[Test]
public void Performance()
{
    int times = 100000;

    var sw = new System.Diagnostics.Stopwatch();
    sw.Start();
    for (int i = 0; i < times; i++)
    {
        using (var source = new TextSource("<!doctype><html><head><title>Hello World</title></head><body><h1>Goodbye World</h1></body></html>"))
        using (var tokenizer = new HtmlTokenizer(source, null, HtmlEntityService.Resolver))
        {
            var token = tokenizer.Get();
            Assert.NotNull(token);
        }
    }
    sw.Stop();
    Console.WriteLine(sw.ElapsedMilliseconds + "ms");
}
```

Running in `Release` 10 times each for property/field:

image

FlorianRappl:
Yep, as I said I know that automatic properties **should** inline, but again: I always like to confirm this to see that no odd behavior has been introduced by anything.

BTW:

``` cs
var sw = new System.Diagnostics.Stopwatch();
sw.Start();
```

is a little bit more elegant to be used like

``` cs
var sw = System.Diagnostics.Stopwatch.StartNew();
```

Not that it matters though :dancers:.

Issue #313, opened 2016-03-30
Error is given for self-closing img tag ending with "/>" (unlike ending with just ">")


savehansson:
According to W3C `img` is a "void element" and as such the start tag _may_ end with `/>`. From [the W3C specification](https://www.w3.org/TR/html5/syntax.html#syntax-start-tag):

> 6_._ Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

AngleSharp gives error code `34 The given tag cannot be self-closed`. If the `/` character is omitted no error is given.

FlorianRappl:
Maybe I understood the specification wrong. But as far as I understood it HTML5 just tolerates trailing slashes, it does not advocate them. The advocated way is to omit the slash actually.

Now going to the [HTML 5.1 specification](https://www.w3.org/TR/2016/WD-html51-20160310/syntax.html#the-in-body-insertion-mode) I see _A start tag whose tag name is one of: "area", "br", "embed", "img", "keygen", "wbr"_. This leads to [the definition for the following quote](https://www.w3.org/TR/2016/WD-html51-20160310/syntax.html#acknowledge-the-tokens-self-closing-flag).

> Acknowledge the token’s self-closing flag, if it is set.

The interpretation that has been used to far lead to the current implementation.

Using validator.nu also does not show any errors with the trailing slash. I will file this as a bug / enhancement and solve it for the next release.

FlorianRappl:
See PR #315 for a fix.

PR #315, opened 2016-03-30
Issue #313

FlorianRappl:
I also looked over other uses of the ACK-flag. Seems to be alright so far.

### DIFF
--- a/AngleSharp.Core.Tests/Library/Error.cs
+++ b/AngleSharp.Core.Tests/Library/Error.cs
@@ -42,22 +42,10 @@
             var config = new Configuration(events: parseErrors);
             var document = source.ToHtmlDocument(config);
 
-            Assert.AreEqual(4, parseErrors.Received.Count);
+            Assert.AreEqual(1, parseErrors.Received.Count);
             Assert.AreEqual((int)HtmlParseError.DoctypeMissing, parseErrors.Received[0].Code);
             Assert.AreEqual(1, parseErrors.Received[0].Position.Column);
             Assert.AreEqual(1, parseErrors.Received[0].Position.Line);
-
-            Assert.AreEqual((int)HtmlParseError.TagCannotBeSelfClosed, parseErrors.Received[1].Code);
-            Assert.AreEqual(69, parseErrors.Received[1].Position.Column);
-            Assert.AreEqual(2, parseErrors.Received[1].Position.Line);
-
-            Assert.AreEqual((int)HtmlParseError.TagCannotBeSelfClosed, parseErrors.Received[2].Code);
-            Assert.AreEqual(82, parseErrors.Received[2].Position.Column);
-            Assert.AreEqual(10, parseErrors.Received[2].Position.Line);
-
-            Assert.AreEqual((int)HtmlParseError.TagCannotBeSelfClosed, parseErrors.Received[3].Code);
-            Assert.AreEqual(70, parseErrors.Received[3].Position.Column);
-            Assert.AreEqual(18, parseErrors.Received[3].Position.Line);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


driekus77:
Hi all,

Is there any sign of releasing this fix?
Are there workarrounds where I can work with?

Thanks for any help.

Henry

FlorianRappl:
Unfortunately, I currently have too much work to do besides AngleSharp, which is why there are still some other (unrelated) issues waiting to be completed before v0.9.6 can be released.

If you can't wait I recommend to get the repo (branch `devel`), build it, and then reference it. I am sorry that it is that inconvenient, but I'd like v0.9.6 to be complete before its being published. I hope to find the time to release v0.9.6 either this week or the next.

driekus77:
@FlorianRappl Thanks for your quick reply.

Unfortunately I've tried tried rebuilding AngleSharp develop branch but the problem still remains. Am I missing a setting somewhere? I am using AngleSharp under the PreMailer. Tested both by compiling the Release target en referencing AngleSharp from PreMailer.

Thanks for the help.
Henry

FlorianRappl:
What exactly is your problem? Can you provide a MWE? Potentially, there is not a problem but rather a mismatch of expectation.

driekus77:
Hi,

My valid XHTML input:
`
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
    <meta name="viewport" content="width=device-width"/>
`

Is parsed to invalid XHTML output:
`
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
    <meta name="viewport" content="width=device-width">
`

Because of the self closing tags missing. I've attached a zip file with my input html example: hero.html and a PreMailer.NET output example: hero.out.html.

Again Thanks for the help.

Kind regards,
Henry

[hero.zip](https://github.com/AngleSharp/AngleSharp/files/236203/hero.zip)

FlorianRappl:
This is (as already thought) not an issue with AngleSharp, but with the expectations you have. AngleSharp behaves like a browser (it has an HTML5 engine at its core), i.e., the input you show is parsed as usual.

The output you present is _one_ possible representation (not _the_ representation). You could implement your own serializer (given in form of an implementation of the `IMarkupFormatter` interface) to produce exactly the output you'd like to see.

So the misconception is that AngleSharp parses a string to a string. In fact it parses a string to an object model. The string you see is just _one representation_ (which happens to be the standard HTML5 way) and may be different to what you want / need.

driekus77:
Hi @FlorianRappl,

Aha now I'm getting it. Thanks again.

I also looked at other issues concerning XHTML formatter and AngleSharp. I'm a C# developer and willing to write a XhtmlMarkupFormatter. But I have 2 questions before I dive into that:
- Would it be possible to select the new to be XHtmlMarkupFormatter depending on original DocType?
- How would you implement this XHtml formatter? As extension (inheritance) of XmlMarkupFormatter or HtmlMarkupFormatter or Standalone? Where the Standalone could cause code dupplication.

Kind regards,
Henry

FlorianRappl:
It is possible to supply a serializer that works according to the doctype. I would go as follows:
- Start with your own implementation of `IMarkupFormatter`
- Provide a field with a child `IMarkupFormatter` (by default set to the default serializer)
- On serializing a doctype try to find a better match (otherwise fall back to the default serializer) and set the field to the new selection
- Always use the selected serializer for all the implemented methods

I would call this one `AutoSelectedMarkupFormatter` or similar. But for sure such a component would requrie the `XHtmlMarkupFormatter`.

Inheritance of the formatters is usually not encouraged. Instead, you should prefer composition. Independent of that decision - yes, the `XmlMarkupFormatter` is possibly the closest match!

Hope this helps you!

Issue #128, opened 2015-08-15
Escaping in script elements in XHTML


thorn0:
``` c#
var document = new HtmlParser().Parse("<script>console.log(a<a>a)</script>");
var formatter = new AngleSharp.Xml.XmlMarkupFormatter();
var result = document.ToHtml(formatter);
```

result:

``` html
<html><head><script>console.log(a<a>a)</script></head><body></body></html>
```

I thought it should've been

``` xml
<html><head><script>console.log(a&lt;a&gt;a)</script></head><body></body></html>
```

but it turns out the browsers don't understand the entities inside scripts, so CDATA and JS comments must be used for JS:

``` xml
<html><head><script>//<![CDATA[
console.log(a<a>a)//]]></script></head><body></body></html>
```

However, this is too JS-specific whereas script elements can contain other languages that might have a different syntax for comments. So I don't see how this problem can be solved in the general case.

Actually, I'm not trying to achieve compatibility with specifically XHTML. I merely want my markup to be valid XML. It's a reasonable expectation from a class named `XmlMarkupFormatter`, isn't it?

FlorianRappl:
Well, I don't know. The `XmlMarkupFormatter` can be used in conjunction with HTML, but do not expect it to output valid XML. Not everything that is valid in HTML can be translated to XML - HTML has a 5 tokenization states unlike XML. Things like script state (for script tags) or rawtext (e.g. style) or even raw characters (title, textarea) are not present in XML. The `XmlMarkupFormatter` is not prepared for these edge cases.

Instead the `XmlMarkupFormatter` is only prepared for output generated in an `XmlDocument` or `SvgDocument`. Here the parser already generated the tree using XML rules.

I would not exclude the option of making something like the `XmlMarkupFormatter` ready to output valid XML even from an HTML tree, but a lot more work would be required. Right now there is no priority on that from my side - but you are free to help out here.

thorn0:
As I mentioned, I myself don't see a generic solution to this problem, but at least it looks strange that `IMarkupFormatter` implementations can't control output for `script` elements at all.

FlorianRappl:
Well, how should they? The `TextContent` is already set, but it has been generated by the parser. The element itself does not know how it has been generated. It only knows about its current state.

thorn0:
[...]

PR #323, opened 2016-04-26
First draft implementing XhtmlMarkupFormatter and using it.

driekus77:
Hi @FlorianRappl,

I've manged to get a first draft working using your idea's. Copied the XmlMarkupFormatter to XhtmlMarkupFormatter and merged everything I thought was usefull from the HtmlMarkupFormatter.
Next to that I've added 2 x 2 simple unittests to check if the self closing tag is working. 

Review comments are welcome.

Kind regards,

Henry


AngleSharp.Silverlight/AngleSharp.Core.Silverlight.csproj
```diff
@@ -3131,8 +3131,10 @@
     <Compile Include="..\AngleSharp\Xml\XmlMarkupFormatter.cs">
       <Link>Xml\XmlMarkupFormatter.cs</Link>
     </Compile>
+    <Compile Include="AutoSelectedMarkupFormatter.cs" />
```
FlorianRappl: They are not linked - please delete the duplicates and link the files. Only the `Core` should contain the files.

AngleSharp.Legacy/AutoSelectedMarkupFormatter.cs
```diff
@@ -0,0 +1,170 @@
+﻿using System;
```
FlorianRappl: Just for consistency - could we move the usings inside the namespace?

AngleSharp.Legacy/AngleSharp.Core.Legacy.csproj
```diff
@@ -3116,8 +3116,10 @@
     <Compile Include="..\AngleSharp\Xml\XmlMarkupFormatter.cs">
       <Link>Xml\XmlMarkupFormatter.cs</Link>
     </Compile>
+    <Compile Include="AutoSelectedMarkupFormatter.cs" />
```
FlorianRappl: See below with Silverlight: Please remove the duplicates and just link against the originals in `Core`.

AngleSharp.Core.Tests/Xhtml/PreserveValidXhtml.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System.IO;
```
FlorianRappl: This is really just a minor one: One using is above the namespace, the other two are within the namespace. All within would be appreciated, thanks!

AngleSharp/Dom/ShadowRoot.cs
```diff
@@ -50,7 +50,7 @@ public IElement Host
 
         public String InnerHtml
         {
-            get { return ChildNodes.ToHtml(HtmlMarkupFormatter.Instance); }
+            get { return ChildNodes.ToHtml(new AutoSelectedMarkupFormatter(this.Owner.Doctype)); }
```
FlorianRappl: Actually, as `InnerHTML` is an official W3C method the official formatter needs to be taken (expected behavior).

AngleSharp/Extensions/FormatExtensions.cs
```diff
@@ -55,7 +55,7 @@ public static void ToCss(this IStyleFormattable style, TextWriter writer)
         /// <returns>The source code snippet.</returns>
         public static String ToHtml(this IMarkupFormattable markup)
         {
-            return markup.ToHtml(HtmlMarkupFormatter.Instance);
+            return markup.ToHtml(new AutoSelectedMarkupFormatter());
```
FlorianRappl: This really boils down to a matter of taste: What do you like as a standard value? I would like to match the behavior of standard browsers, as this is the goal of AngleSharp. Therefore I would leave the current implementation as it is (see `InnerHtml`).

AngleSharp/Extensions/FormatExtensions.cs
```diff
@@ -83,7 +83,7 @@ public static String ToHtml(this IMarkupFormattable markup, IMarkupFormatter for
         /// <param name="writer">The output target of the serialization.</param>
         public static void ToHtml(this IMarkupFormattable markup, TextWriter writer)
         {
-            markup.ToHtml(HtmlMarkupFormatter.Instance);
+            markup.ToHtml(new AutoSelectedMarkupFormatter());
```
FlorianRappl: Same as for line 58.

AngleSharp/Html/HtmlMarkupFormatter.cs
```diff
@@ -16,6 +16,15 @@ public sealed class HtmlMarkupFormatter : IMarkupFormatter
         /// </summary>
         public static readonly IMarkupFormatter Instance = new HtmlMarkupFormatter();
 
+        /// <summary>
+        /// ChildFormatter Property
+        /// </summary>
+        public IMarkupFormatter ChildFormatter
```
FlorianRappl: Is there a reason for this property at `HtmlMarkupFormatter`? I guess only the auto selected one needs this.

AngleSharp/Html/PrettyMarkupFormatter.cs
```diff
@@ -34,6 +34,15 @@ public PrettyMarkupFormatter()
         #region Properties
 
         /// <summary>
+        /// Child / Implementation Formatter
+        /// </summary>
+        public IMarkupFormatter ChildFormatter
```
FlorianRappl: Same as for the `HtmlMarkupFormatter`.

AngleSharp/IMarkupFormatter.cs
```diff
@@ -9,6 +9,15 @@
     public interface IMarkupFormatter
     {
         /// <summary>
+        /// ChildFormatter property.
+        /// </summary>
+        IMarkupFormatter ChildFormatter
```
FlorianRappl: Why is the interface extended with information like this?

FlorianRappl:
Alright, I've added some comments. From the functional point of view this looks great and I do want to merge it. There are some subtle, but really minor, issues that are not worth discussing much.

What needs to be discussed, however, are two points:
1. Changing of default behavior (I think the short `ToHtml` extension methods should be left as-is, to be aligned with the `InnerHtml`, where I have reasons to leave them in their current state; as explained).
2. The extension of `IMarkupFormatter` with the `ChildFormatter` property. Why is this required? To me it is only there for a specific formatter and all the others have no use for the property, i.e., the property should not be on the interface. Am I missing something?

Also please remove the duplicates (they happen easily since Visual Studio doesn't raise awareness for linking against copying; one may easily copy instead of link as this is the preferred way of Visual Studio) and replace them with links to files being placed in `AngleSharp.Core`.

Thanks a lot for your efforts - great job and will be placed in v0.9.6 as a solution for #128.

driekus77:
I've Commited "Fixing review comments" which does what it says.

To answer on your final comments:
Point 1 Changing of default behavior: I was hoping to fix AngleSharp only so that I didn't have to propose a change in the PreMailer.Net project. I've reverted related changes back to what it was.

Point 2 ChildFormatter in interface: That was just me not thinking straight :-(. Sorry.

Removing the duplicates: I did it by editing the project files. Could not find another way.

Feel free to add more comments if you prefer.

Kind regards,

Henry

FlorianRappl:
Regarding the linking, in VS you have two options:
1. Per drag and drop of files from other projects (additional key press: ALT (?))
2. "Add existing item" and instead of pressing "Add" use the dropdown menu and select "Add as link".

Otherwise looks great now - thanks a lot :+1:. Let's merge it!

AngleSharp/Xhtml/XhtmlMarkupFormatter.cs
```diff
@@ -0,0 +1,216 @@
...
+        String IMarkupFormatter.Text(String text)
+        {
+            var temp = Pool.NewStringBuilder();
+
+            for (int i = 0; i < text.Length; i++)
+            {
+                switch (text[i])
+                {
+                    case Symbols.Ampersand: temp.Append("&amp;"); break;
+                    case Symbols.NoBreakSpace: temp.Append("&nbsp;"); break;
```
thorn0, line 116: It's probably safer to use `&#160;` instead of `&nbsp;` here. Support for `&nbsp;` in XHTML depends on the doctype. See http://stackoverflow.com/questions/17095154/xhtml-and-nbsp-doesnt-work

FlorianRappl:
Agreed, please file another PR.

driekus77:
Oke I will change it to &#160; Sorry didn't know about this issue with XHTML and nbsp.

Kind regards,
Henry

FlorianRappl:
I meant that @thorn0 could file another PR, but if you do it then this is fine with me as well. Thanks again for your commitment, much appreciated!

Issue #160, opened 2015-09-28
Invalid character detected.


useruser1:
at AngleSharp.Dom.Element.SetAttribute(String name, String value)
   at AngleSharp.Parser.Html.HtmlForeignExtensions.AdjustAttribute(Element element, String name, String value)
   at AngleSharp.Parser.Html.HtmlDomBuilder.InBodyStartTag(HtmlTagToken tag)
   at AngleSharp.Parser.Html.HtmlDomBuilder.InBody(HtmlToken token)
   at AngleSharp.Parser.Html.HtmlDomBuilder.Home(HtmlToken token)
   at AngleSharp.Parser.Html.HtmlDomBuilder.Consume(HtmlToken token)
   at AngleSharp.Parser.Html.HtmlDomBuilder.Parse(HtmlParserOptions options)
   at AngleSharp.Parser.Html.HtmlParser.Parse(String source)

example source - http://mems.duke.edu/faculty

FlorianRappl:
Thanks for the report. I was able to identify the problem (invalid character in foreign element attribute) and fix the behavior. The problem will be fixed in the upcoming release (v0.9.3).

PR #327, opened 2016-04-28
Fix XHTML issue with nbsp by using #160 in stead.

driekus77:
Small fix for XHTML issue with `&nbsp;` by replacing it by `&#160;`.

As See http://stackoverflow.com/questions/17095154/xhtml-and-nbsp-doesnt-work 
Thanks to @thorn0 for pointing this out!

Kind regards,
Henry


FlorianRappl:
Thanks for your contribution and the instant improvement!

PR #343, opened 2016-05-16
ToHtml(TextWriter) does not write to the TextWriter

TheCloudlessSky:
The new `ToHtml(TextWriter)` extension method in `0.9.6` doesn't use the `TextWriter` resulting in no output being produced. I've added test cases for the `ToHtml()` extensions as well as fixing the bug.

The current workaround is to use the non-extension:

``` csharp
document.ToHtml(writer, HtmlMarkupFormatter.Instance);
```


FlorianRappl:
Yes, you are right - my bad! Thanks for providing the test cases and the fix. I'll merge the PR once its checks are done.

Issue #334, opened 2016-05-09
Change code to C# 6


FlorianRappl:
Currently, the code uses C# 5 to work on any platform with most compilers. However, I think it is now time to move on and continue with C# 6. First, the move towards CoreCLR focuses on a more modern stack anyway (also requiring tools that know C# 6 anyway), second C# 6 compilers should now be available for every platform.

The features that will be used the most are certainly the conditional access operator ("Elvis" `?.`) and the `nameof` operator. Also `using static` may be handy in one or the other class.

This is merely an information such that developing AngleSharp could potentially require a toolset / compiler update soon. Anticipated for v0.9.8.

aloisdg:
With C# 6, you can also replace all your `String.Format()` thanks to interpolation string. For example:

``` C#
public override String ToString()
{
        return String.Format("({0}) -- ({1})", _start, _end);
}
```

Will become:

``` C#
public override String ToString()
{
        return $"({_start}) -- ({_end})";
}
```

It is very handy!

[snippet](https://github.com/AngleSharp/AngleSharp/blob/dc85fd5f6329647b2a3577c57de90f5ff68f23a4/src/AngleSharp/TextRange.cs#L62)

FlorianRappl:
Yep!

qaqz111:
With C# 6, you can also replace all your `1-line methods` thanks to lambda expression body syntax. For example:

``` C#
public override String ToString()
{
    return $"({_start}) -- ({_end})";
}
```

Will become:

``` C#
public override String ToString() => $"({_start}) -- ({_end})";
```

It is very handy!

:laughing: :laughing: :laughing:

FlorianRappl:
The expression body syntax is not really useful for methods (its adding a layer of inconsistency that brings more noise to the code than order).

However, the expression body syntax is most useful for computed properties, which should be one-liners by nature.

qaqz111:
What does `a layer of inconsistency that brings more noise to the code than order` exactly mean?

I've peeked into IL disassemble of the 2 different syntax, there's not significant differences between them, and usually the expression body make a shorter IL code segment.

qaqz111:
Oh I see, you just mean what they looks are, relative to those multiple lines methods that included in curly braces. 

Hmmmmmm, I think, this is just a personal habits?
I'd like to write the short 1-line methods in expression form, so that I can see what the method does without expand the `[+]` in VS editor.

So, whether use it or not is not very important! :smile:

[...]

PR #346, opened 2016-05-22
Adding .NET Core Build

### LINKED ISSUE #270: CoreCLR support
Are there any plans to support CoreCLR?

Suchiman:
Fixes #270 
This PR replaces AngleSharp.Legacy and AngleSharp.Silverlight with a .NET Core / project.json build that multitargets for net40, silverlight50, pcl259 and netstandard1.0

Replaces the defines LEGACY and SILVERLIGHT with the predefined NET40 and SL50 symbols.

Also fixes a race with the GC in WeakReference<T>.TryGetTarget which i've noticed incidentally while making multitargeting work.


FlorianRappl:
Good one - with the change to xproj and project.json I see this also more aligned to #334 (the solution also changed to Visual Studio 2015).

However, one thing that confuses me are the changes in the _gitignore_ file. Any reason for deleting so much stuff? For instance, the rule to ignore the tools folder is important.

Also the CI is failing - potentially, AppVeyor needs to be re-configured for a project.json build. I'll have to look into this.

Suchiman:
Mhh, might have to revise the gitignore update, I took the most recent version from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

FlorianRappl:
Yes, please revert it and only consider adding parts as necessary (I actually doubt that adding further stuff is required for xproj / project.json support, but I don't know for sure).

Suchiman:
Minimalized gitignore changeset

FlorianRappl:
Great thanks. Does cake build for you locally? As long as AppVeyor check fails I hesitate to merge the PR.

Maybe the issue is more general and can be solved locally. A short search did not yield any results pointing in direction of AppVeyor.

Suchiman:
I will have to take a look later, but i assume you have to prepare AppVeyor, for example, https://github.com/serilog/serilog/blob/dev/appveyor.yml#L5-L9

FlorianRappl:
Yes was definitely a good hint and worth a try, however, even adding

``` PowerShell
mkdir -Force ".\build\" | Out-Null
Invoke-WebRequest "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1" -OutFile ".\build\installcli.ps1"
$env:DOTNET_INSTALL_DIR = "$pwd\.dotnetcli"
'& .\build\installcli.ps1 -InstallDir "$env:DOTNET_INSTALL_DIR" -NoPath'
$env:Path = "$env:DOTNET_INSTALL_DIR;$env:Path"
```

to the initital setup does not make it work. I think the problem lies in the cake build script. Somehow MSBuild needs to be configured correctly. Hence, I think you should first make it work locally before AppVeyor can handle it.

If it works locally I am sure we can find a way to automate it!

Suchiman:
The cake script runs locally now, but i'm not sure why AppVeyor pukes on the csproj while trying to restore packages...

Judging from the message, it almost looks like trying to build the csproj in the restore step.

```
Executing task: Restore-Packages
MSBuild auto-detection: using msbuild version '14.0' from 'C:\Program Files (x86)\MSBuild\14.0\bin'.
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(724,5):
    error : The OutputPath property is not set for project 'AngleSharp.Core.csproj'.
            Please check to make sure that you have specified a valid combination of Configuration and Platform for this project.
            Configuration='Release' Platform='Any CPU'.
            You may be seeing this message because you are trying to build a project without a solution file,
            and have specified a non-default Configuration or Platform that doesn't exist for this project.
            [C:\projects\anglesharp\src\AngleSharp\AngleSharp.Core.csproj]
```

FlorianRappl:
I think the problem could be caused by the following

> Looks like a bug in nuget or something. The problem is that AngleSharp.Core.csproj uses AnyCPU but the error is that there is no OutputPath defined for Any CPU. From what I understand, the sln is supposed to remap Any CPU to AnyCPU, but for whatever reason, when nuget encounters a project.json file, it doesn't do the remapping.

So the same author solved it via

> Worked around it by adding some logic to the .csproj to manually remap.

Essentially two platforms have been added to the csproj file:

``` xml
    <Platform Condition=" '$(Platform)' == 'Any CPU' ">AnyCPU</Platform>
    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
```

Maybe this does the trick!

Romanx:
Hey there,

I recently updated my project to dotnet core and use appveyor. Feel free to take a look at my [appveyor.yml](https://github.com/StubbleOrg/Stubble/blob/master/appveyor.yml)

Appveyor recently updated to include the CLI ad long as you're targeting VS2015 in the image. There's some other stuff i'm doing with coveralls but that should be easy to ignore :+1:

FlorianRappl:
Hi @Romanx thanks for the input. Yes, looks good, but I am already using the VS 2015 image. Also the dotnet restore and all that should probably not be the origin of the problem (the build is invoked from Cake actually). From the error message it sounds more like a configuration problem (as the one described / solved above).

The quotes above came from the [MailKit](https://github.com/jstedfast/MailKit/issues/212) repo. There a similar thing happened.

Romanx:
@FlorianRappl Whoops, my bad!

Suchiman:
Looks better now @FlorianRappl !

FlorianRappl:
This is great!

FlorianRappl:
Hey @jeremymeng could you please review the changes for the nuspec file? Since you PR'ed the initial changes (#283) and since you are the expert I would love to get your feedback before I accept the PR. (Essentially, it is a change from `dotnet` to `netstandard` - which [should be fine](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md) according to my limited knowledge.)

In my opinion everything looks great and I would give it a go. Thanks already, much appreciated :rose:!

jeremymeng:
@FlorianRappl The changes look good to me.  Thank you @Suchiman!

FlorianRappl:
Wonderful - much appreciated! Thanks for the review @jeremymeng and thanks for the wonderful PR @Suchiman. :+1: 

:rocket: for OSS!

PR #350, opened 2016-05-26
Add a monoandroid and monotouch target to the NuGet package

### LINKED ISSUE #349: Add "monoandroid" and "monotouch" to the nuspec, otherwise AngleSharp can't be installed on Xamarin
NuGet will try to install the "dotnet" or "netstandard" target into Xamarin Android and iOS projects, but since the dependencies listed there aren't compatible, the install fails

flagbug:
Fixes #349


FlorianRappl:
Looks great!

PR #394, opened 2016-09-06
Fix BaseLoader crash when Set-Cookie not present in headers

osolliec:
The following crashes :

```
            var config = Configuration.Default.WithDefaultLoader();

            var address = "http://google.fr";
            var client = BrowsingContext.New(config);
            var document = await client.OpenAsync(address);
```

This pr aims to fix the crash when there is no Set-Cookie header.

### DIFF
--- a/src/AngleSharp/Network/Default/BaseLoader.cs
+++ b/src/AngleSharp/Network/Default/BaseLoader.cs
@@ -152,7 +152,10 @@
                 if (response != null)
                 {
                     redirectCount++;
-                    SetCookie(request.Address, response.Headers[HeaderNames.SetCookie]);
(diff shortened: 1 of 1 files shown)


FlorianRappl:
Thanks for the fix, however:
- never PR against `master`
- please write about the issue first (and check in `devel`): you would have noticed that this one is already fixed in `devel`

Anyhow, thanks for your commitment - much appreciated!

osolliec:
Ok I didn't notice.

When do you think the fix will make it in a new nuget package?

Thanks !

FlorianRappl:
I would release a hotfix (v0.9.8.1) on Friday, but I am currently waiting if any other issues pop up. So far does not seem like it (if you find another one - please submit it)!

👍 on your efforts!

Issue #285, opened 2016-03-15
Improve and finalize CSSOM


FlorianRappl:
This is a continuation of #230. Basically, the CSSOM representation has been started to change, however, it is not in a final stage. Ideally, it will be more fine-grained than it is right now.

Besides internal changes to how the CSSOM is structured (which will be just a continuation of the work from #230, as mentioned previously), the CSSOM API will be defined, which should provide many more methods / properties than the W3C specification. All in all this should enable to gather as much information as desired.

Requirements are currently collected. If you have some use-case, feel free to drop it in.

FlorianRappl:
This one should also already consider [CSS Houdini](https://github.com/w3c/css-houdini-drafts), e.g.:
- https://drafts.css-houdini.org
- https://wiki.css-houdini.org

If the core ideas are followed, then the CSSOM should be already quite future proof (of course adjustments will be necessary for the final spec., but these should then be as minimalistic as possible).

FlorianRappl:
Continued in AngleSharp.Css!

PR #409, opened 2016-09-23
#285 stroke property

jswietek:
Implemented stroke-dasharray property


FlorianRappl:
Another good PR, thanks!

PR #412, opened 2016-09-24
Issue #285 text props

jswietek:
Implemented missing text-\* css properties:
- text-align-last
- text-anchor
- text-justify


FlorianRappl:
This one also looks good, but it does go to `master`. Potentially, this is just by mistake. I have to close it, as GitHub does not allow to change the target.

Please re-PR to `devel` (as with the others). Thanks 👍 !

PR #413, opened 2016-09-25
Issue #285 text props

jswietek:
Implemented missing text-\* css properties:
- text-align-last
- text-anchor
- text-justify

Re-targeted to devel


FlorianRappl:
Very good - thanks a lot! Much appreciated 👍

Issue #439, opened 2016-10-20
Pseudo-class :has() not properly selecting children


daveaglick:
Using `:has(> p)` will return any element that contains a `p` descendant instead of just those elements that contain `p` as a direct child.

Here's a failing test:

```
[Test]
public void HasSelectorChild()
{
        var source = @"<div><div><p>Hello in a paragraph</p></div></div>";

        var document = source.ToHtmlDocument();
        var selector = "div:has(> p)";
        var result = document.QuerySelectorAll(selector);
        Assert.AreEqual(1, result.Length);
        Assert.AreEqual(document.Body.ChildNodes[0].ChildNodes[0], result[0]);
}
```

I'm taking a look at this now and am going to attempt a PR. Might need some help depending on how far I get into the parser code.

daveaglick:
Now that I'm digging into it, I think the problem is deeper than just getting children nodes correctly. For example, the following test also fails:

```
[Test]
public void HasSelectorFollowing()
{
    var source = @"<div><div><p>Hello in a paragraph</p><p>Another paragraph</p></div></div>";

    var document = source.ToHtmlDocument();
    var selector = "p:has(+ p)";
    var result = document.QuerySelectorAll(selector);
    Assert.AreEqual(1, result.Length);
    Assert.AreEqual(document.Body.ChildNodes[0].ChildNodes[0].ChildNodes[0], result[0]);
}
```

In both cases I _think_ it's because the `:has()` pseudo-class is missing some context about its parent node. Still digging...

daveaglick:
As I go down the rabbit hole, I think this might be blocked by #352, specifically support for setting a real `:scope`.

What I think it missing is the ability of the inner selector to reference a particular scope element as the starting point. For example, in the selector `div:has(> p)`, the nested selector `> p` should be evaluated for every `div` using it as the scope. Just as `:scope` only points to `:root` right now because there's no way to set a scope for a given selector, `:has()` is broken due to the same limitation.

I'm sure you already know this, but I'll throw it in here for my own reference. I think the [relevant specification text](https://drafts.csswg.org/selectors-4/#relative) is:

[...]

Issue #352, opened 2016-05-26
Renew CSS selectors


FlorianRappl:
The CSS selectors - as they are currently - are limited.

For one they can be only constructed internally, thus also limiting the extensibility. This has been addressed somehow, however, now we have some factories and no uniform standard. Also the factories as they are cannot cover all cases.

One example is #347. The `:scope` requires an additional argument being used (what is current scope?) in construction. This, however, is not given. Only a `string` can be passed to the selector "constructor", which - in fact - does not do any constructing but rather just "selects" the right selector (pun intended).

It was decided that this topic goes hand in hand with #146 and will be addressed in v0.9.8.

FlorianRappl:
With the renewed way of constructing selectors `:scope` from #347 will be finished (currently set to be equivalent to `:root`).

For specification, see [MDN](https://developer.mozilla.org/de/docs/Web/CSS/:scope) and [W3C Selector L4](https://drafts.csswg.org/selectors-4/#the-scope-pseudo).

atifaziz:
> The :scope requires an additional argument being used (what is current scope?) in construction. This, however, is not given. Only a string can be passed to the selector "constructor"

I believe the `:scope` is defined by the element on which, for example, [`querySelectorAll`](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll) is being called.

FlorianRappl:
Yes - that is the mentioned argument. It is called `context`. But the `context` could be also given implicitly, e.g., if the selector is part of a sheet that comes from a `style` element with the `scoped` attribute being set. In this case the `style`s parent is the `context`.

atifaziz:
Understood. I was only coming from the angle of using CSS selectors as a method for querying and selecting DOM elements.

FlorianRappl:
Essentially done (see PR #487). The rest will coupled to / handled in the `css-extension` branch.

PR #440, opened 2016-10-20
Introduces the concept of scope and improves the :has() pseudo-class

daveaglick:
Trying to get `:scope` and `:has()` working correctly (#439). I've introduced the concept of scopes to queries. Because an `ISelector` doesn't maintain a query state (and thus shouldn't know the scope before matching), the scope gets passed down from the outer-most `.QuerySelector()` call.

I think I'm pretty close, though I still can't get some of my new tests for `:has()` to pass. I suspect I need to spend a little more time tweaking `HasFunctionState` to properly use the new scoping stuff.

Mainly I just wanted your feedback at this point. I know this is something you were looking at in #352 so I figured I'd surface this in case it's close to the direction you were thinking of going.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -552,6 +552,18 @@
             Assert.IsNull(value);
             Assert.AreEqual("eeeee", results[1].TextContent);
         }
+        
+        [Test]
+        public void ScopeSelectorChild()
+        {
+            var source = @"<p>First paragraph</p><div><p>Hello in a paragraph</p></div>";
+
+            var document = source.ToHtmlDocument();
+            var selector = ":scope > p";
+            var result = document.Body.Children[1].QuerySelectorAll(selector);
+            Assert.AreEqual(1, result.Length);
+            Assert.AreEqual(document.Body.ChildNodes[1].ChildNodes[0], result[0]);
+        }
 
         [Test]
         public void HasSelectorSimple()
@@ -564,6 +576,42 @@
             var result = document.QuerySelectorAll(selector);
             Assert.AreEqual(1, result.Length);
             Assert.AreEqual(document.Body.ChildNodes[0], result[0]);
+        }
+
+        [Test]
+        public void HasSelectorChild()
+        {
+            var source = @"<div><div><p>Hello in a paragraph</p></div></div>";
+
+            var document = source.ToHtmlDocument();
+            var selector = "div:has(> p)";
+            var result = document.QuerySelectorAll(selector);
+            Assert.AreEqual(1, result.Length);
+            Assert.AreEqual(document.Body.ChildNodes[0].ChildNodes[0], result[0]);
+        }
+
+        [Test]
+        public void HasSelectorFollowing()
+        {
+            var source = @"<div><div><p>Hello in a paragraph</p><p>Another paragraph</p></div></div>";
+
+            var document = source.ToHtmlDocument();
+            var selector = "p:has(+ p)";
+            var result = document.QuerySelectorAll(selector);
+            Assert.AreEqual(1, result.Length);
+            Assert.AreEqual(document.Body.ChildNodes[0].ChildNodes[0].ChildNodes[0], result[0]);
+        }
+
+        [Test]
+        public void HasSelectorFollowingWithoutRelative()
+        {
+            var source = @"<div></div><div><p>Hello in a paragraph</p><p>Another paragraph</p></div>";
+
+            var document = source.ToHtmlDocument();
+            var selector = "div:has(p + p)";
+            var result = document.QuerySelectorAll(selector);
+            Assert.AreEqual(1, result.Length);
+            Assert.AreEqual(document.Body.ChildNodes[1], result[0]);
         }
 
         [Test]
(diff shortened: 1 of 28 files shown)


daveaglick:
I think I'm getting close. It can pass the simple `:scope` test I added and it also passes all the additional `:has()` tests except one.

The failing test is from the spec and uses an adjacent sibling selector `p:has(+ p)`. It's tricky because the `HasFunctionState` implementation typically looks at children of the current node for matches. In this case however, we'd need to look at the current node itself to find the match.

My thought at the moment is to change the behavior under certain conditions to go up a level and match against self. I'll play a little more with this tomorrow to see if I can get a clear test board.

FlorianRappl:
Alright, very cool! Keep this one updated - it would be great to have it in the next release!

daveaglick:
All tests now pass. I resolved the issue of supporting sibling matches relative to the scope of `:has()` (such as the `+` adjacent sibling example) by checking if the scope is being using either implicitly (as is the case when the pseudo-class selector starts with a delimiter) or explicitly (I.e., `:scope` in the pseudo-class selector). In both cases, we search for matches a level up so we check for matches not only in the target element's child nodes, but it's siblings too. This should be safe since we've ensured the query scope is in use, and that refers to the target element.

At this point I'll consider this PR ready for review and merge. It's a big change, so it'll obviously need an extra set of eyes. Hopefully I've covered all the bases.

FlorianRappl:
Alright, I looked over it - looks very good. Besides the cosmetic points of fitting to the overall source code style of AngleSharp I guess the point to discuss is if we really want to expose the scope parameter in _some_ `QuerySelector/All` signatures. I would not include it. I would only have the `Match` signature changed - if one really wants to use the scope parameter one could always construct an `ISelector` and use it directly.

Apart from that discussion it all seems very good - thanks for the PR! Looks fantastic 👍

daveaglick:
Thanks :) The scope needs to be set when we make the outer call to `QuerySelector()` since it tracks the document/element that we're executing the query against. To make sure I understand, you're thinking that instead of passing the scope through the internal calls to the `QuerySelector()` helpers that we should wrap the whole selector in something like a `ScopedSelector` and use that to pass the scope instead of the method signatures? Let me take a stab at that and see what I come up with...

daveaglick:
So I didn't get too far. The problem is with the methods in the internal `QueryExtensions` class. For example, take the following implementation (before I modified it):

```
public static IElement QuerySelector(this INodeList elements, String selectors)
{
    var sg = CreateSelector(selectors);

    if (sg == null || sg is UnknownSelector)
        throw new DomException(DomError.Syntax);

    return elements.QuerySelector(sg);
}
```

It gets the selector string and is responsible for creating the representative `ISelector`. So if I were going to use a new `ScopedSelector` to wrap the underlying selector, this is where I'd instantiate it and set the element scope. However, notice there's no element being passed in the method signature, only the set of nodes I'm evaluating the query against. So in order to create the scoped selector and give it a scope I still need to pass the scope `IElement` to this method.

So it's possible to use a wrapping `ISelector` that maintains the scope, but I'd still need to add the extra `IElement` to all the `QuerySelector()` helper methods that are responsible for building the selector from the selector string. AFAIK, these are all internal so it shouldn't change the API (for that matter, I don't think any of the `QuerySelector()` methods I originally added scope to were public).

Given that I'd need to add the scope element to at least a few `QuerySelector()` methods, is it worth removing it from the others and using an `ISelector` to pass scope or should we leave it as-is? Or equally likely, maybe I've overlooked something?

FlorianRappl:
The `QuerySelector/All` methods (and extension methods, which exist for convenience) represent the DOM methods with the same names. I would rather create a _public_ extension method for the `ISelector` named `MatchAll`, which is performing a selection with scope against a set of nodes.

daveaglick:
Makes sense to keep the methods that match the DOM consistent. How would that work for calls like `someElement.QuerySelectorAll(":scope > p")`? Ideally, the scope for this query should be set to `someElement` even though we never explicitly created an `ISelector` or called any methods on one?

What if we added new alternative internal helpers called something like `QuerySelectorWithScope()` and then called those from inside the public `QuerySelector()` methods? We could also use the idea of a `ScopedSelector` to avoid modifying the other internal `QuerySelector` methods. That would keep the edits to both the public and internal APIs to a minimum.

I'll try and work up and iteration of this idea today or tomorrow to help continue the discussion.

daveaglick:
Think I just realized what you were going after. Essentially the same as above, except instead of a `QuerySelectorWithScope()` helper extension, we'll have an extension on `ISelector` that does the wrapping and introduction of scope. Is that about right?

FlorianRappl:
Regarding the first point: In the browser the executing element seems to be the scope. So I can confirm this behavior as you anticipated it.

I compared

``` js
document.body.querySelectorAll('body > a') //a
document.body.querySelectorAll(':scope > a') //b
```

both yield the same results. Other experiments resulted in the same output. So I think yes - the scope should be set to the element where the query is run against. If the element is run against the `document` then the `IDocument` node needs to be the scope.

Regarding the second point: Yes, exactly! Instead of the `QuerySelectorWithScope` we introduce an extension to `ISelector`, which is more natural and should also be exposed to the outside (i.e., `public`).

daveaglick:
Okay, pushed a new commit that I hope is a bit closer to what you were thinking. I added the change as a second commit instead of amending so there'd be a record of the first version for now. I can squash if/once you're ready to merge.

FlorianRappl:
Looks very good! Thanks for your patience with me 👍 !

daveaglick:
Hey, cool! Wasn't expecting a merge so quickly :smile: And _thank you_ - I use this a ton, so glad I finally found a way to contribute to such a great project.

Issue #442, opened 2016-10-22
Xhtml entities


corliss:
Hi - long time! Great work.

I just synced up to the latest revision after almost a year and am having a bit of difficulty with Xhtml entities, in particular `&apos;`

According to [wikipedia](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Entities_representing_special_characters_in_XHTML) `&apos;` is the only entity that is handled differently between XHTML and HTML.

What is the right way to parse XHTML that contains the `&apos;` entity as well as other XHTML entities? The `XmlEntityService` fails because it only has the five basic entities, and `HtmlEntityService` fails on `&apos;`. I did not find an `XhtmlEntityService`. 

Thanks!

corliss:
I was able to hack around my issues by modifying `HtmlEntityService`. I changed the line containing `apos` to `AddBoth` instead of `AddSingle`, and it worked. However, I don't understand the difference between `AddSingle` and `AddBoth`. Another failing unit test involving `mdash` was worked around in the same way.

FlorianRappl:
The easiest way would be to implement your own entity service and use the one that is the closest (i.e., `HtmlEntityService` in your case) for almost all entities - except the ones that are different.

At this stage I am not sure if I should include such a service in the core. Already the XHTML formatter is probably misplaced. I leave that one open until I made up my mind. Also I introduce an XHTML tag to find issues related to XHTML for later usage.

corliss:
Thanks - I don't mind writing a custom entity service, but I'd like to do it correctly. Could you please explain what should be done? For example, I don't understand `AddBoth`  vs `AddSingle`. 

Which entities would be affected? I don't want to miss out on any. I've already made the hack for `apos`, `mdash` and `theta` - just based on random testing.

If I want to do exhaustive unit tests for this service, how should I do that?

Having said that, can I ask why you don't want such a service in the core? (Again, I don't mind writing it and sending a PR.) Isn't XHTML common enough for Anglesharp to support?

corliss:
I also notice that some entities are listed multiple times with different cases, e.g. `gt`, `GT`, and `Gt` - are these all required? The wikipedia page lists only gt. (As you can tell, I am not super knowledgeable about entities and quite wary of the bugs that may arise.)

FlorianRappl:
The `IEntityProvider` interface _is_ really simple: You only need one method called `GetSymbol`, which takes a string and returns a string. That's all!

[...]

PR #444, opened 2016-10-23
Issue #442

FlorianRappl:
Renamed the entity services to entity provider (to match the interface name) and unified calls (all end now with semicolons - if available).


corliss:
Thanks. I am currently on the `master` branch and am going to sync to the `devel` branch. Is it stable, or should I hold off until `devel`is merged into `master`?

FlorianRappl:
It is stable in the sense that `devel` is always building, but API-wise it is not stable. Also since in the upcoming version (v.10.0) CSS is dropped (and I think you care about CSS a lot) you should wait until it is fully released (with the next release also AngleSharp.Css will be published - which gives _at least_ the functionality as it is right now).

FlorianRappl:
(But you could theoretically do a git-cherry-pick to get the specific changes.)

corliss:
Thanks - I will hold off in that case. Are you separating out the CSS assembly? Will it be a separate nuget?

FlorianRappl:
Yes!

Issue #460, opened 2016-11-22
Build on Linux


FlorianRappl:
From the discussion in #448 it seems that AngleSharp currently has problems to build on Linux.

This should be fixed and instructions on how to build on Linux (and Windows, Mac OS) should be provided for completeness (e.g., install .NET Core using ..., clone the code, ... run x).

The instructions should be part of the upcoming documentation.

FlorianRappl:
⚠️ ⚠️ ⚠️ 

**The following content is a WIP**

Follow the instructions on the [.NET core homepage](https://www.microsoft.com/net/core#linuxubuntu), e.g., for Ubuntu:

```bash
sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ xenial main" > /etc/apt/sources.list.d/dotnetdev.list'
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893
sudo apt-get update
sudo apt-get install dotnet-dev-1.0.0-preview2-003131
```

The version of .NET core to be installed can also be [obtained from the list of downloads](https://www.microsoft.com/net/download/linux).

Furthermore, due to current limitations in [Cake](http://cakebuild.net) we also require Mono for the Bootstrapper.

```bash
sudo sh -c 'echo "deb http://download.mono-project.com/repo/debian wheezy main" > /etc/apt/sources.list.d/mono-xamarin.list'
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
sudo apt-get update
sudo apt-get install mono-devel
sudo apt-get install referenceassemblies-pcl 
```

*Apparently, NuGet v3 with Mono has a severe bug that does not work when the UTC offset is positive (e.g., Germany has UTC+1). The solution is to (at least temporarily) set the time to some other country, e.g., UK where we have UTC+0.*

Potentially, Mono needs to be updated to the latest version. To identify this the command `mono --version` may be used. The minimum tested version is `4.6.2`. We recommend running

```bash
sudo apt-get upgrade mono-devel
```

to ensure that the latest version is definitely installed.

To prevent NuGet bugs spoiling the build the following download command is recommended:

```bash
curl https://dist.nuget.org/win-x86-commandline/v2.8.6/nuget.exe -o tools/nuget.exe
``` 

Clone AngleSharp:

```bash
[...]

PR #469, opened 2016-11-24
Build on Linux

fogzot:
Various changes to build on Linux with Mono:

* Renamed a variable in build.cake from "target" to "targetDir" to avoid
  a conflict with a variable with the same name in an outer scope.

* Added --skip-dotnet-core option to build.sh: "dotnet restore" is still
  flacky and segfault on some Ubuntu and Debian installations with updated
  kernel or libssl. This allows to build the 4.5 profile and run the tests
  event if DOTNET Core is unavailable.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -16,6 +16,7 @@ using Octokit;
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Release");
+var skipDotNetCore = Argument("skip-dotnet-core", "no") == "yes";
 var isLocal = BuildSystem.IsLocalBuild;
 var isRunningOnUnix = IsRunningOnUnix();
 var isRunningOnWindows = IsRunningOnWindows();
@@ -52,7 +53,9 @@ Task("Restore-Packages")
     .Does(() =>
     {
         NuGetRestore("./src/AngleSharp.Core.sln");
-        DotNetCoreRestore("./src/AngleSharp/project.json");
+        if (!skipDotNetCore) {
+            DotNetCoreRestore("./src/AngleSharp/project.json");
+        }
     });
 
 Task("Build")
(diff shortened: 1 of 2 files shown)


FlorianRappl:
I made similar changes, see [my branch](https://github.com/FlorianRappl/AngleSharp/blob/issue-%23460/build.cake).

However, the reason I did not PR yet is that I also want the unit tests to work.

Also I would not skip dotnet core completely - in my case I decided only to skip all frameworks except the .NET Standard framework.

fogzot:
Sorry, I didn't understand you were actively working on this.
You need to skip dotnet core completely because even the restore step fails without the dotnet core tools installed and that breaks the build.

FlorianRappl:
Please see the instructions that are a WIP in the original issue #460. Dotnet core needs to be installed to build AngleSharp. I have no interest in providing a non-dotnet core build as AngleSharp will definitely move to support *only* .NET Standard in a future version (potentially already for v1, otherwise for v2).

fogzot:
I don't see anything wrong in supporting only .NET Standard in the future but right now requiring it makes impossible to build AngleSharp on machines where .NET Core and its tools are unsupported. Anyway, it is not a big problem: I can just apply my patch after a pull, before the local build.

Right now I am running the tests and trying to understand why they fail on Mono.

FlorianRappl:
For me on a fresh Ubuntu installation it worked without any problems. Just adding the .NET feed and installing .NET core made it work. See my WIP instructions.

Regarding the failing tests I am quite sure that this is (partially) from an encoding issue, which could have been solved by not taking the sources from the fs (as a string), but rather from a packed / isolated format. I will investigate / improve this soon. The other failing tests are probably from a bug in the Mono implementation (`CookieContainer`).

fogzot:
It segfault on Debian and also on Ubuntu, depending on the software/libs you install. There are multiple bugs open on dotnet/coreclr and dotnet/cli. But, as I said, it is not a problem. They'll fix it eventually. In the meantime I'll try to help as I can on the failing tests and XML support.

I did some quick test on the Mono `CookieContainer` and just setting/getting cookies works as expected - in fact the tests that simply set/get a cookie seem to work. AFAIK, the failing ones are (all?) related to requests to `httpbin.org`.

FlorianRappl:
Interesting - but I think most (all?) integration tests regarding cookies are running against httpbin.org. So the problem may lie in the requester.

FlorianRappl:
Cherry picked all important stuff into `master`. Thanks!

PR #476, opened 2016-12-01
Optimized away INode.GetAllElements() allocations

HellBrick:
The PR greatly reduces the allocations made by `INode.GetAllElements()` extension that's called under the hoods every time the document tree is traversed to fill `HtmlCollection`. This method was surprisingly high on the list of top allocators when doing things like enumerating all links/scripts found in the HTML document. To illustrate the point, let's consider a simple benchmark:

```C#
HtmlParser parser = new HtmlParser();

for ( int i = 0; i < 1000; i++ )
{
	// I used some random real-world ~110KB HTML file here
	IHtmlDocument document = parser.ParseDocument( TestData.TorrentSite );
	foreach ( IElement link in document.Links )
	{
	}

	foreach ( IHtmlScriptElement script in document.Scripts )
	{
	}
}
```

Before this PR the allocation profile looked like this:

Profile: before

As you can see, the memory pressure caused by iterating through a parsed document was basically on par with creating all the strings for attribute values, tag names, etc. This was caused by all the `IEnumerable<T>` and `IEnumerator<T>` instances that were allocated recursively when traversing the tree. This PR addresses this by eliminating recursion and allocating fixed number of objects per `GetAllElements()` call ( 1 `Enumerable`, 1 `Enumerator` and 1 `Stack` for inner enumerable + 2 `Enumerable` / `Enumerator` pairs for `OfType()` and `Where()`. `Stack` can allocate multiple additional arrays under the hood, but they are not enough to be noticed in the profile). Here's the profile after this PR:

Profile: after

`GetAllElements()` memory pressure was basically eliminated (it's somewhere around 0.3 MB now).

### DIFF
--- a/src/AngleSharp/Dom/Internal/CollectionExtensions.cs
+++ b/src/AngleSharp/Dom/Internal/CollectionExtensions.cs
@@ -3,6 +3,8 @@
     using AngleSharp.Text;
     using System;
     using System.Collections.Generic;
+    using System.Collections;
+    using System.Linq;
 
     /// <summary>
     /// A bunch of methods for getting DOM elements.
@@ -158,22 +160,9 @@
 
         private static IEnumerable<T> GetAllElements<T>(this INode parent, Predicate<T> predicate)
             where T : class, INode
-        {
-            for (var i = 0; i < parent.ChildNodes.Length; i++)
-            {
-                var child = parent.ChildNodes[i] as T;
-
-                if (child != null && predicate(child))
-                {
-                    yield return child;
-                }
-
-                foreach (var element in parent.ChildNodes[i].GetAllElements<T>(predicate))
-                {
-                    yield return element;
-                }
-            }
-        }
+            => new NodeEnumerable(parent)
+            .OfType<T>()
+            .Where(node => predicate(node));
 
         private static IEnumerable<T> GetDescendendElements<T>(this INode parent, Predicate<T> predicate)
             where T : class, INode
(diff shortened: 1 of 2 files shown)


src/AngleSharp/Dom/Internal/CollectionExtensions.cs
```diff
@@ -158,6 +160,12 @@ public static T GetElementById<T>(this IEnumerable<T> elements, String id)
 
         private static IEnumerable<T> GetAllElements<T>(this INode parent, Predicate<T> predicate)
             where T : class, INode
+            => new NodeEnumerable(parent)
+            .OfType<T>()
+            .Where(node => predicate(node));
```
FlorianRappl: I think we can optimize here by changing `Predicate<T>` to `Func<T, Boolean>`? Then we have `Where(predicate)`, i.e., 1 delegate instead of 2 delegates. I am not sure if the CLR can optimize this away otherwise.

FlorianRappl:
Great change 👍 ! Memory is very important and this one is definitely a step in the right direction. Did the performance of large queries also improve?

src/AngleSharp/Dom/Internal/CollectionExtensions.cs
HellBrick: Nice catch, I've switched the code to `Func<T, bool>`.

FlorianRappl:
Very cool - thanks. After PR I will adjust the style to adhere to the project (here single statements are not blocked + keywords instead of types).

HellBrick:
>Did the performance of large queries also improve?

I haven't really tested large queries. This PR affects somewhat specific things (`IDocument`'s `Forms`, `Images`, `Scripts`, etc.) and no matter what you do from there, as long as you enumerate through all the results (without bailing early by `FirstOrDefault` or something), you had to pay the allocation price. The enumeration itself became a bit faster, but it was fast enough anyway, the whole point of the change is to avoid death by GC under load =)

Since I happen to have the numbers, here's the small speed benchmark:

```C#
public class TreeTraversingBenchmark
{
	private static readonly IDocument _document = new HtmlParser().ParseDocument( TestData.FastTorrent );

	[Benchmark]
	public void EnumerateAllNodes()
	{
		foreach ( var node in _document.All )
		{
		}
	}

	[Benchmark]
	public void EnumerateAllLinks()
	{
		foreach ( var node in _document.Links )
		{
		}
	}

	[Benchmark]
	public void EnumerateLinksWithImagesInside()
	{
		foreach ( var node in _document.Links.Where( l => l.FirstChild?.NodeName == "img" ) )
		{
		}
	}
}
```

```ini
BenchmarkDotNet=v0.10.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7 CPU 960 3.20GHz, ProcessorCount=8
Frequency=3162609 Hz, Resolution=316.1946 ns, Timer=TSC
Host Runtime=Clr 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
GC=Concurrent Workstation
JitModules=clrjit-v4.6.1586.0
Job Runtime(s):
        Clr 4.0.30319.42000, Arch=64-bit RELEASE [RyuJIT]
```

Before:
```
                         Method |        Mean |     StdDev |      Median |
------------------------------- |------------ |----------- |------------ |
              EnumerateAllNodes | 563.1366 us |  9.6620 us | 562.6725 us |
              EnumerateAllLinks | 499.8031 us | 26.8478 us | 493.6902 us |
 EnumerateLinksWithImagesInside | 509.4505 us |  8.0875 us | 506.4249 us |
```

After:
```
                         Method |        Mean |     StdDev |      Median |
------------------------------- |------------ |----------- |------------ |
              EnumerateAllNodes | 322.8301 us |  3.9090 us | 322.2199 us |
              EnumerateAllLinks | 405.7342 us | 12.3445 us | 401.5507 us |
 EnumerateLinksWithImagesInside | 396.6019 us |  1.9667 us | 396.3261 us |
```

>After PR I will adjust the style to adhere to the project (here single statements are not blocked + keywords instead of types)

Great, I'll look at the commit where you fix the code style and will try to follow the style better. Hopefully, this is not my last PR here, I have a few more optimizations in mind ;)

FlorianRappl:
This is great news! Much appreciated 👍 

Thanks for the benchmark. I think the improvement is great - besides the max. 40% gain we also see that the standard deviation decreased a lot; so less allocations means less dependent on the system and much more stable performance. 🎉 

Tell me where to ship the beer to 🍻 🍻 🍻  ... : )

PR #481, opened 2016-12-02
Optimized CharacterData.Append() a bit

HellBrick:
Just a small thing I've noticed while looking around. This method is called *a lot* when appending tokens to `CharacterData`, and every time `String.Remove()` was called with `count = 0`. For the reasons unknown, this is not a noop, but actually copies the string. The difference on my current benchmark:

Before:
Before

After:
After

I'm not a huge fan of those `Insert()`s either: we create a token, allocate a string for it, then pass it around a bit... just to append it to another string. And then do it again a few more dozen times =) This scenario can be optimised further, but some preparation is needed to handle it properly, so this will do for now.

### DIFF
--- a/src/AngleSharp/Dom/Internal/CharacterData.cs
+++ b/src/AngleSharp/Dom/Internal/CharacterData.cs
@@ -180,7 +180,12 @@
             owner.QueueMutation(MutationRecord.CharacterData(target: this, previousValue: _content));
 
             var deleteOffset = offset + data.Length;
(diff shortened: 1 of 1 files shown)


FlorianRappl:
I am not a huge fan either. But the code as given is already a huge optimization regarding the specification, which does it for every character. A couple of tricks have been needed to reduce it too much larger strings. The detection of the copy is a good thing - should definitely be avoided.

Great job 🍺

Issue #483, opened 2016-12-04
Two ContextLoadingTests fail with DirectoryNotFoundException when running tests by dotCover


HellBrick:
The tests are `LoadContextFromStreamLoadedShouldNotBeStuckForever` and `    LoadContextFromStreamLoadedShouldNotFaceBufferTooSmall` and they both fall with the similar exception, only the file name differs:
```
System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Users\HellBrick\AppData\Local\Temp\alby2xhq.a1k\aqkizyvj.bqc\AngleSharp.Core.Tests\assembly\dl3\Resources\kommersant.ru-1.res'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.File.InternalWriteAllBytes(String path, Byte[] bytes, Boolean checkHost)
   at AngleSharp.Core.Tests.External.PageRequester.<AddResourceAsync>d__6.MoveNext() in F:\Software\AngleSharp\src\AngleSharp.Core.Tests\External\PageRequester.cs:line 59
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at AngleSharp.Core.Tests.External.PageRequester.<PerformRequestAsync>d__4.MoveNext() in F:\Software\AngleSharp\src\AngleSharp.Core.Tests\External\PageRequester.cs:line 29
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #485, opened 2016-12-04
Made resource directory lookup relative to the source file

HellBrick:
Here's what I can suggest to deal with #483. Instead of assuming the dll location, this assumes the source file location, which in my opinion makes more sense: both resource folder and the source file locations are explicitly controlled by the repository, unlike the dll location, which is controlled by the test runner.

### DIFF
--- a/src/AngleSharp.Core.Tests/External/PageRequester.cs
+++ b/src/AngleSharp.Core.Tests/External/PageRequester.cs
@@ -17,7 +17,13 @@
 
         private static String GetResourceDirectory( [CallerFilePath] String fileName = null )
         {
-            return Path.Combine(Path.GetDirectoryName(fileName), "..", "Resources");
+            var directoryPath = Path.Combine(Path.GetDirectoryName(fileName), "..", "Resources");
+            if (!Directory.Exists(directoryPath))
+            {
+                Directory.CreateDirectory(directoryPath);
+            }
+
+            return directoryPath;
         }
 
         public override Boolean SupportsProtocol(String protocol)
--- a/src/AngleSharp.Core.Tests/External/PageRequester.cs
+++ b/src/AngleSharp.Core.Tests/External/PageRequester.cs
@@ -7,12 +7,18 @@
     using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
+    using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


FlorianRappl:
This one, however, assumes that the code will always be bundled with the resources and all - what if I change the path? I think the origin of this issue is already a what-if. So if you want to answer the one "what-if" then all the others have to be answered as well.

tl;dr: Good fix for your use-case but not as general as intended.

HellBrick:
That's already an assumption that's used today. And if the resources aren't bundled with the code, then such tests can never work anyway, because whoever clones the repo won't get the resources at all. I'm not suggesting anything drastic here, just trying to adjust implementation of the assumption a bit so dotTrace/Resharper users would be able to run tests as well.

FlorianRappl:
The resources are bundled with the code - they are checked in. The download is only made once (or if the local resources are deleted). So if one does not receive the resources then they are created. The test may still fail in such cases as the referenced website may have changed.

The current assumption is imho a natural one: Just leave the binary where it is. But the argument / fix is also a very small one, so again: yes this works for the particular scenario.

src/AngleSharp.Core.Tests/External/PageRequester.cs
```diff
@@ -7,13 +7,19 @@
...
     sealed class PageRequester : BaseRequester
     {
         private readonly static DefaultHttpRequester _default = new DefaultHttpRequester();
-        private readonly static String _directory = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "..", "..", "Resources");
+        private readonly static String _directory = GetResourceDirectory();
         private readonly static SiteMapping _mapping = new SiteMapping(Path.Combine(_directory, "content.xml"));
 
+        private static String GetResourceDirectory( [CallerFilePath] String fileName = null )
+        {
+            return Path.Combine(Path.GetDirectoryName(fileName), "..", "Resources");
```
FlorianRappl: Maybe let's use the chance to create the directory if it does not exist already.

HellBrick: Done

FlorianRappl:
Great - thanks; appreciated! 🏖

PR #493, opened 2016-12-13
Fix a few possible null references

mganss:
I've found these while playing with the static analysis tool [PVS-Studio](http://www.viva64.com/en/b/0457/). It reports a large number of additional issues. Most of them are benevolent or false alarms but some could be fixed, plus there are some where I'm not sure. Here are a few examples:

- `V3072 The 'CreateDocumentOptions' class containing IDisposable members does not itself implement IDisposable. Inspect: _source. CreateDocumentOptions.cs 14` (quite a few of those)
- `V3114 IDisposable object 'ms' is not disposed before method returns. StringExtensions.cs 764` (MemoryStream)
- `V3090 Unsafe locking on 'this' instance in class 'BaseLoader'. BaseLoader.cs 19`
- `V3013 It is odd that the body of 'NegatedAttributeExistenceSelectorWithDeclaredNamespaceA' function is fully equivalent to the body of 'NegatedAttributeExistenceSelectorWithDeclaredNamespaceB' function. CssW3CSelector.cs 2500` (large number of these, at first glance there seems to be some css which was not copied from the test cases at w3.org)

If you want to check it yourself, here's a shell one-liner to add the comment which is required by the free version of PVS-Studio to all *.cs files:

```shell
 find . -not -path "*/obj/*" -name "*.cs" -exec sed -i -e '1s/^\xef\xbb\xbf//' -e '1i// This is an open source non-commercial project. Dear PVS-Studio, please check it.\r\n// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com\r\n' {} \;
```

### DIFF
--- a/src/AngleSharp/Browser/TaskEventLoop.cs
+++ b/src/AngleSharp/Browser/TaskEventLoop.cs
@@ -125,13 +125,13 @@
 
             public Boolean IsRunning
             {
-                get 
-                { 
+                get
+                {
                     return _task != null &&
-                           _task.Status == TaskStatus.Running || 
-                           _task.Status == TaskStatus.WaitingForActivation || 
+                           (_task.Status == TaskStatus.Running ||
+                           _task.Status == TaskStatus.WaitingForActivation ||
                            _task.Status == TaskStatus.WaitingToRun ||
-                           _task.Status == TaskStatus.WaitingForChildrenToComplete; 
+                           _task.Status == TaskStatus.WaitingForChildrenToComplete);
                 }
             }
 
--- a/src/AngleSharp/BrowsingContextExtensions.cs
+++ b/src/AngleSharp/BrowsingContextExtensions.cs
@@ -347,10 +347,11 @@
             foreach (var service in services)
             {
                 var otherCulture = service.Culture;
-                var otherTwoLetters = otherCulture.TwoLetterISOLanguageName;
 
                 if (otherCulture != null)
                 {
+                    var otherTwoLetters = otherCulture.TwoLetterISOLanguageName;
+
                     if (otherCulture.Equals(culture))
                     {
                         return service;
--- a/src/AngleSharp/Html/Dom/Internal/HtmlElement.cs
+++ b/src/AngleSharp/Html/Dom/Internal/HtmlElement.cs
@@ -583,13 +583,13 @@
             {
                 parent = parent.ParentElement;
             }
-            
+
             if (parent == null)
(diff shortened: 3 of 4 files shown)


FlorianRappl:
It's good that you looked into PVS Studio (I have a license) already. I had this on my list for the v0.10 release as well.

Looking into the issues right now.

src/AngleSharp/Browser/TaskEventLoop.cs
```diff
@@ -125,13 +125,13 @@ public Boolean IsCompleted
...
             public Boolean IsRunning
             {
-                get 
-                { 
+                get
+                {
                     return _task != null &&
-                           _task.Status == TaskStatus.Running || 
-                           _task.Status == TaskStatus.WaitingForActivation || 
+                           (_task.Status == TaskStatus.Running ||
```
FlorianRappl, line 131: The one before should have worked, still grouping is a best practice. I would, however, create an extension method _task.IsNotYetCompleted(). One could then combine it do have `_task?.IsNotYetCompleted() ?? false`.

FlorianRappl:
So the possible null-references are all good here - this is a great PR!

Regarding the four points:

* V3072 The 'CreateDocumentOptions': I think false alarm as this one is used as intermediate handover. The `IResponse` is disposable (the stream is contained), but it will be consumed / released somewhere else. (**false**)
* V3114 IDisposable object 'ms' is not disposed before method returns. Yes, I think this should be introduced. (**true**)
* V3090 Unsafe locking on 'this' instance in class 'BaseLoader'. I need to have a closer look here. (**unsure**)
* V3013 It is odd that the body of 'NegatedAttributeExistenceSelectorWithDeclaredNamespaceA' function is fully equivalent to the body of 'NegatedAttributeExistenceSelectorWithDeclaredNamespaceB' function. There was some difference which never manifested (commented out for long I think); so long story short: some of these tests may be obsolete. (**true**)

src/AngleSharp/Browser/TaskEventLoop.cs
mganss: Do you mean the code without the parens? This would have thrown if `_task == null` because && has higher precedence than ||.

FlorianRappl: True! Ultimately, I was referring to expressing the intent via an extension method and using that one with the null-conditional operator. But the modified version is great!

Issue #501, opened 2017-01-02
AngleSharp does not correctly compute href when the base url has a port


meziantou:
AngleSharp does not correctly compute href when the base url has a port:

```
var config = Configuration.Default.WithDefaultLoader();
var browsingContext = BrowsingContext.New(config);
var url = "http://localhost:5000";
var content = "<a id='sample' href='http://www.sample.com'></a>";

var htmlDocument = await browsingContext.OpenAsync(action => action.Address(url).Content(content));
var element = (IHtmlAnchorElement)htmlDocument.GetElementById("sample");
```
`element.Href` is `http://www.sample.com:5000/` instead of `http://www.sample.com`

FlorianRappl:
Yes - you are right! Thanks for the report - will be fixed in the next version 👍

FlorianRappl:
Fixed, see PR #503.

PR #502, opened 2017-01-02
Issue #501

FlorianRappl:
Fixed the relative URL-port problem as described in #501.


FlorianRappl:
Wrong target branch! 💣

PR #524, opened 2017-01-28
Do not use recursion to query elements

### LINKED ISSUE #523: StackOverflowException in document.QuerySelectorAll("a")
AngleSharp.dll produces fatal crash (that kills the whole process) when we try to extract links from http://onestop.fiu.edu/financial-aid/loans/ page (that seems safe with only 108 links).
We were able to reproduce this crash on some developers machines, but not on all machines.
Please try to run code below on Windows 7 machine.
```
using AngleSharp.Parser.Html;

[TestMethod]
public void AngleSharpFailureTest()
{
	var pageHtml = LoadPageHtmlContentFromUrl("http://onestop.fiu.edu/financial-aid/loans/"); // This line is a pseudo-code. Please use your own url load method, or load pageHtml from attached AngleSharpFailureTest.html.zip
	var document = parser.Parse(pageHtml);
	// The QuerySelectorAll("a") call in the next line:
	// - Works well when running without debugger
	// - Works well on Windows 10 machine with 16 GB RAM under Visual Studio 2013 debugger
	// - Fatally fails on Windows 7 machine with 16 GB RAM under Visual Studio 2013 debugger
	// - Fatally fails on Windows 7 machine with 8 GB RAM under Visual Studio 2013 debugger
	// - Fatally fails on Windows 7 machine with 4 GB RAM under Visual Studio 2013 debugger
	// - Fatally fails on Windows Server 2012 R2 virtual machine with 9 GB RAM under Visual Studio 2013 debugger
	// Fatal failure: "An unhandled exception of type 'System.StackOverflowException' occurred in AngleSharp.dll".
	// That fatal failure kills the whole process, so we cannot catch that StackOverflowException.
	var nodes = document.QuerySelectorAll("a");
	Assert.AreEqual(108, nodes.Length);
}
```
[AngleSharpFailureTest.html.zip](https://github.com/AngleSharp/AngleSharp/files/736935/AngleSharpFailureTest.html.zip)

image

campersau:
because it could result in a StackOverflowException in very deep documents

fixes #523

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -812,5 +812,45 @@ nav h1, nav h2, nav h3, nav h4, nav h5, nav h6";
             Assert.AreEqual("this", result[0].ClassName);
             Assert.AreEqual("3", result[0].TextContent);
         }
+
+        [Test]
+        public void FindDeepElement()
+        {
+            var count = 10000;
+            var doc = String.Empty.ToHtmlDocument();
+
+            var node = (IElement)doc.Body;
+            for (var i = 0; i < count; i++)
+            {
+                var newNode = doc.CreateElement("div");
+                node.AppendChild(newNode);
+                node = newNode;
+            }        
+            node.AppendChild(doc.CreateElement("a")); 
+
+            var result = doc.QuerySelector("a");  
+            Assert.NotNull(result);
+        }
+
+        [Test]
+        public void FindDeepElements()
+        {
+            var count = 10000;
+            var doc = String.Empty.ToHtmlDocument();
+
+            var node = (IElement)doc.Body;
+            for (var i = 0; i < count; i++)
+            {
+                var newNode = doc.CreateElement("div");
+                node.AppendChild(newNode);
+                node = newNode;
+            }
+            node.AppendChild(doc.CreateElement("a"));
+            node.AppendChild(doc.CreateElement("a"));
+
+            var result = doc.QuerySelectorAll("a");
+            Assert.AreEqual(result.Length, 2);
+        }
+
     }
 }
--- a/src/AngleSharp/Css/Dom/SelectorExtensions.cs
+++ b/src/AngleSharp/Css/Dom/SelectorExtensions.cs
@@ -26,11 +26,9 @@
                     return element;
                 }
 
-                if (element.HasChildNodes)
+                foreach (var child in element.Descendents<IElement>())
                 {
-                    var child = selector.MatchAny(element.Children, scope);
-
-                    if (child != null)
+                    if (selector.Match(child, scope))
                     {
                         return child;
                     }
@@ -76,9 +74,12 @@
                     result.Add(element);
                 }
 
-                if (element.HasChildNodes)
+                foreach (var child in element.Descendents<IElement>())
                 {
-                    selector.MatchAll(element.Children, scope, result);
+                    if (selector.Match(child, scope))
+                    {
+                        result.Add(child);
+                    }
                 }
             }
         }
--- a/src/AngleSharp/Dom/NodeExtensions.cs
+++ b/src/AngleSharp/Dom/NodeExtensions.cs
@@ -103,11 +103,18 @@
         {
             foreach (var child in parent.ChildNodes)
             {
-                yield return child;
-
-                foreach (var subchild in child.GetDescendants())
+                var stack = new Stack<INode>();
+                stack.Push(child);
+                while (stack.Count > 0)
                 {
-                    yield return subchild;
+                    var next = stack.Pop();
+                    yield return next;
+
+                    var length = next.ChildNodes.Length;
+                    while (length > 0)
+                    {
+                        stack.Push(next.ChildNodes[--length]);
+                    }
                 }
             }
         }


src/AngleSharp.Core.Tests/Css/CssSelector.cs
```diff
@@ -812,5 +812,45 @@ public void NthChildSpanThisApplied()
             Assert.AreEqual("this", result[0].ClassName);
             Assert.AreEqual("3", result[0].TextContent);
         }
+
+        [Test]
+        public void FindDeepElement()
+        {
+            var count = 10000;
+            var doc = "<html></html>".ToHtmlDocument();
```
FlorianRappl: Just make `String.Empty.ToHtmlDocument()` - this code does not make much sense.

src/AngleSharp.Core.Tests/Css/CssSelector.cs
```diff
@@ -812,5 +812,45 @@ public void NthChildSpanThisApplied()
...
+
+            var node = doc.DocumentElement;
```
FlorianRappl: Append the `div` elements to `doc.Body`.

src/AngleSharp.Core.Tests/Css/CssSelector.cs
```diff
@@ -812,5 +812,45 @@ public void NthChildSpanThisApplied()
...
+
+            var result = doc.QuerySelector("a");  
+            Assert.NotNull(result);
+        }
+
+        [Test]
+        public void FindDeepElements()
+        {
+            var count = 10000;
+            var doc = "<html></html>".ToHtmlDocument();
```
FlorianRappl: Same as above.

src/AngleSharp/Css/Dom/SelectorExtensions.cs
```diff
@@ -21,16 +22,9 @@ public static IElement MatchAny(this ISelector selector, IEnumerable<IElement> e
         {
             foreach (var element in elements)
             {
-                if (selector.Match(element, scope))
+                foreach (var child in new[] { element }.Concat(element.Descendents<IElement>()))
```
FlorianRappl: For performance reasons Linq should be avoided here.

campersau:
OMG this was bad... Sorry, should have tested it more before submitting...
It passes now.

FlorianRappl:
Looks great so far 🥇 - I'll have to check performance for regression before I give it a go. But from the code's perspective it looks fine to! Will be in for v0.10.

dennisgorelik:
@FlorianRappl When do you plan to release v0.10?

src/AngleSharp/Css/Dom/SelectorExtensions.cs
dennisgorelik: @FlorianRappl Did you measure how much that change (avoiding Linq) improved performance here?

campersau: Not sure about performance but the best would be a `DescendantsAndSelf` method like in [HtmlAgilityPack](https://htmlagilitypack.codeplex.com/SourceControl/latest#Trunk/HtmlAgilityPack/HtmlNode.cs) to reduce code duplication.

dennisgorelik: @campersau Do you mean that using DescendantsAndSelf (from System.Xml.Linq) would make code more elegant and compact, but you do not know what would be the performance implications?

FlorianRappl: I think he is not referring to the method from `System.Xml.Linq`, but rather to a general method with such a name and behavior.

src/AngleSharp/Dom/NodeExtensions.cs
```diff
@@ -103,11 +103,18 @@ public static IEnumerable<INode> GetDescendants(this INode parent)
         {
             foreach (var child in parent.ChildNodes)
             {
-                yield return child;
-
-                foreach (var subchild in child.GetDescendants())
+                var stack = new Stack<INode>();
```
campersau, line 106: This can also be moved above the `foreach` to reduce allocations for parent nodes which have a lot of child nodes.
(Maybe with a check if the parent `HasChildNodes`.)

src/AngleSharp/Css/Dom/SelectorExtensions.cs
dennisgorelik: Would that "general" DescendantsAndSelf method help supporting CSS interface defined in w3c?

src/AngleSharp/Dom/NodeExtensions.cs
dennisgorelik: @campersau Does this code change in GetDescendants() change the order in which elements are getting processed?
In previous "Recursion" implementation, descendants of earlier siblings were processed before later siblings [of descendants' parent]. Right?
In this new "Loop" implementation, descendants are only processed after all siblings [of descendants' parent] are processed.
However even if that change of the processing order is actually happening - it is OK for me. Just want to be clear here.

campersau: @dennisgorelik No, this does not change the order in which elements are getting processed. There are tests which cover this. This loop behaves the same as the previouse recursiv version.

FlorianRappl:
A performance test indicates that it did not win significantly, but it also did not lose any performance. So I would give this one a go!

campersau:
Just want to add that `GetDescendants` could also be implemented like this:

```csharp
public static IEnumerable<INode> GetDescendants(this INode parent)
{
    var node = parent;
    var container = parent;

    while (true)
    {
        if (container != null && container.HasChildNodes)
        {
            node = container.FirstChild;
        }
        else
        {
            while (node != parent && node == node.Parent.LastChild)
            {
                node = node.Parent;
            }

            if (node == parent) break;

            node = node.NextSibling;
        }
        yield return node;
        container = node;
    }
}
```

Which is identical to [`System.Xml.Linq`](https://referencesource.microsoft.com/#System.Xml.Linq/System/Xml/Linq/XLinq.cs,4462cd5c87146151) and does not require a `Stack`. But since `NextSibling` is very slow it isn't very performant.

dennisgorelik:
MoreLINQ has traversal implementation:
https://github.com/morelinq/MoreLINQ/blob/d987f0ef30390a9b66d5cc608e5e39648d26097a/MoreLinq/Traverse.cs

It would be nice if AngleSharp supported something like TraverseDepthFirst().
DepthFirst is important in a case when we need to convert HTML to text.

It's not a priority. Replacing Recursion with a loop is far more important, because for different traversal I can use MoreLINQ directly.

campersau:
@dennisgorelik As you can see the implementation of TraverDepthFirst is identitcal to this PR. https://github.com/morelinq/MoreLINQ/blob/d987f0ef30390a9b66d5cc608e5e39648d26097a/MoreLinq/Traverse.cs#L77-L90

dennisgorelik:
@campersau Are you saying that MoreLINQ/TraverseDepthFirstImpl() 
https://github.com/morelinq/MoreLINQ/blob/d987f0ef30390a9b66d5cc608e5e39648d26097a/MoreLinq/Traverse.cs#L77-L90
is functionally identical to AngleSharp/GetDescendants()
https://github.com/campersau/AngleSharp/blob/778db1c29ae14e3a9382af15d79e98b52aba1850/src/AngleSharp/Dom/NodeExtensions.cs#L108-L118
?

campersau:
@dennisgorelik Yes, if you include the two lines above L108:  https://github.com/campersau/AngleSharp/blob/778db1c29ae14e3a9382af15d79e98b52aba1850/src/AngleSharp/Dom/NodeExtensions.cs#L106-L118

PR #525, opened 2017-01-28
Added "Development language" section

dennisgorelik:
Mentioning C# 6 in the README.md should ads important detail to AngleShapr project description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,10 @@ The project's timeline, upcoming implementations, (currently) missing features a
 
 More information is also available by following some of the hyper references mentioned in the Wiki. In-depth articles will be published on the CodeProject, with links being placed in the Wiki at GitHub.
 
+Development language
+--------------------
+AngleSharp is written in C# 6 and thus requires Roslyn as a compiler. You will need to use Visual Studio 2015+ for development here.
+
 Current status
 --------------
 
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ More information is also available by following some of the hyper references men
 
 Development language
 --------------------
-AngleSharp is written in C# 6 and thus requires Roslyn as a compiler. You will need to use Visual Studio 2015+ for development here.
+AngleSharp is written in C# 6 and thus requires Roslyn as a compiler. You will need to use IDE like Visual Studio 2015+ for development here.
 
 Current status
 --------------


FlorianRappl:
I don't think that this additional information is important for anyone consuming the library.

I would rather add this info in the Wiki or the contribution guidelines. But the readme does not contain any such information (otherwise development conventions, folder structure, ..., would all be interesting, too).

Also note that VS 2015+ is not required. You could, e.g., develop AngleSharp in macOS with the IDE of your choice (naturally, ST or VSCode are great options, but you could also use Vim or Emacs as OmniSharp will boost your productivity).

dennisgorelik:
1) README.md was the first place I checked to see code version to check code C# version.
Listing that .NET 4.5 is supported without specifying actual C# code version was implicitly misleading.

2) I agree that https://github.com/AngleSharp/AngleSharp/blob/master/.github/CONTRIBUTING.md should be updated too.
Currently it is out of date: "AngleSharp uses features from C# 5 and will soon switch to C# 6."

3) When I wanted to compile AngleSharp devel branch - I did not plan to contributing code and, therefore, did not even look at https://github.com/AngleSharp/AngleSharp/blob/master/.github/CONTRIBUTING.md

4) I think Visual Studio is a choice in 99% of cases for .NET projects, so even if it is not absolutely correct, mentioning Visual Studio 2015 would help to the reader who is considering using AngleSharp library.

5) I commited https://github.com/AngleSharp/AngleSharp/pull/525/commits/be39c37328b6783ced56cca3d5ad73676c639f6c

FlorianRappl:
.NET 4.5 (framework) and C# 6 (language) are two different things. One is for the consumer (user) of the library, the other thing is for the producer (developer).

Again, if this information is included, then it would rather be in a dedicated development section. Right now it looks out of context to me.

dennisgorelik:
> .NET 4.5 (framework) and C# 6 (language) are two different things. 

Are you implying that the typical reader of current version of README.md is very unlikely to assume that AngleSharp source code would be compilable by any system that has .NET 4.5 + any C# ?

> rather be in a dedicated development section

That's why I put it into new separate "Development language" section.

--
Obviously I now know already that AngleSharp is wrtitten in C# 6, so I am ok if you do not add it to README.md
But I do care about your rejection reasons, because I would like to improve my documentation writing skills.

FlorianRappl:
> Are you implying that the typical reader of current version of README.md is very unlikely to assume that AngleSharp source code would be compilable by any system that has .NET 4.5 + any C# ?

No, I am stating that the version of the language has nothing to do with the framework. C# (any version) can also be used independent of the .NET Framework - the same is true vice versa (think F#, VB, ...). A consumer of the library does not compile it and therefore only cares about the supported frameworks.

> That's why I put it into new separate "Development language" section.

Yeah, I am not talking about a dedicated "Development language" section, but rather a "Development" section, where the used language (incl. minimum version) is just one thing to be mentioned (along other points such as IDE, build system, conventions, ...).

Also note that this is not a rejection (the PR is not closed yet), but rather a discussion to reach a conclusion.

dennisgorelik:
I understand that in theory C# can be independent from .NET framework.
But .NET framework version hints what C# version was most likely used to write that project.

In any case independence of C# version from .NET framework version - is not the reason to exclude C# version from the README.md

Just "Development" - would be a good section title too.
You may add other items too.
But you do not have to add other items right now. And the most important among these items would be C# version number anyway.

FlorianRappl:
> In any case independence of C# version from .NET framework version - is not the reason to exclude C# version from the README.md

The README is geared towards consumers. That should give you the background why mentioning supported platforms is more important than the version of C#.

> Just "Development" - would be a good section title too.

And its not about a section title, its about the content. The content should be about development principles for the project. Right now it would just state the version of C#.

TL;DR: I am not against mentioning the C# version number, but as stated I would prefer to have the complete picture (how to get it, how to build it, where things are placed).

dennisgorelik:
1) Yes, README.md is (and should be) geared toward AngleSharp library consumers
But consumers still are interested in getting answer to the question: "If I have to fix a bug or add a feature myself - what C# version would I have to use?"
That is why adding C# version number to the README.md makes sense.

2) From the perspective of that consumer question (above) - usually it is not important to know development principles behind the project. But if you can deliver these principles in a very brief form - such information can still be on README.md.

3) In any case, the value of adding C# version in formation to README.md is higher, that the value of adding other development principles.
So whether to add them - should be decided mostly separately.
There is no need to go to the extremes and either add or or add nothing.
4) May be add "Development" section to README.md with:
- One line that mention C# version number.
- Link to detailed "Development Principles" file.
?

FlorianRappl:
Sounds good. I would cover the following points:

- Minimum C# version
- Link to CONTRIBUTING.md
- Use braces for any conditional / loop body
- VS2015+ recommended, VSCode with OmniSharp on other platforms
- Use `-Async` when available
- Use VIP ("Var If Possible") style (in C++ called AAA: Almost Always Auto) to place types on the right

Issue #534, opened 2017-02-13
Solving or documenting encoding issues in .NET Core


gabriele-tomassetti:
**First thanks to the authors for this fantastic library.** One thing that I really like is that the API makes really easy to port code from Javascript libraries. Also the detection of encoding is very useful, but I found an issue on .NET Core.

.NET Core doesn't include support for some encodings by default. This lead to the problem that when AngleSharp tries to detect the encoding a lot of old content is forced to default to UTF-8.
The simple way to solve it, as [noted here](http://stackoverflow.com/questions/37870084/net-core-doesnt-know-about-windows-1252-how-to-fix) is:

> to include "System.Text.Encoding.CodePages" as a dependency
> and then call
> Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);

If it's not desirable to add this dependency and the dependencies of the [System.Text.Encoding.CodePages](https://www.nuget.org/packages/System.Text.Encoding.CodePages/) package itself, it may be useful to document the problem.

FlorianRappl:
I think for .NET Core the dependency makes perfect sense.

Could you - by any chance - create a PR? As v0.10 is still in the pipe (shame on me!) the PR could make it in time.

gabriele-tomassetti:
@FlorianRappl For the PR, I just need to clone master, create a branch for the issue and modify the "netstandard1.0" -> "dependencies" section in project.json, isn't it?

jogibear9988:
you can also edit the file directly in github, it will fork, create a branch & pull req automaticaly

FlorianRappl:
Please clone `devel`. `master` is the last release, `devel` is the current (in-development) version.

Also when submitting the PR please do it against `devel`! Thanks 👍 .

LeMoussel:
@jogibear9988 How do you open/edit the file directly in github?

jogibear9988:
https://help.github.com/articles/editing-files-in-your-repository/

gabriele-tomassetti:
I have added the pull request #535 but I didn't notice that adding the System.Text.Encoding.CodePages package would also require to upgrade to netstandard1.3, from netstandard1.0.

FlorianRappl:
Yep that is now a big question. I need to see what platforms would be excluded by such a change!

gabriele-tomassetti:
[...]

PR #535, opened 2017-02-13
Update project.json to solve issue-534

gabriele-tomassetti:
Add System.Text.Encoding.CodePages to solve the issue 534, but to do that it also need to upgrade to netstandard1.3.
This is something that need to be checked.

### DIFF
--- a/src/AngleSharp/project.json
+++ b/src/AngleSharp/project.json
@@ -16,7 +16,7 @@
   },
 
   "frameworks": {
-    "netstandard1.0": {
+    "netstandard1.3": {
       "dependencies": {
         "System.Collections": "4.0.11",
         "System.Diagnostics.Debug": "4.0.11",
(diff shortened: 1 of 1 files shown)


gabriele-tomassetti:
Sorry, I forgot to mention the **really important fact** that this doesn't actually solves the issue (ie, adding Encoding.RegisterProvider). Before doing that I am waiting to know if it's ok to upgrade to netstandard1.3.

FlorianRappl:
Yep, let's discuss this in #534 in detail. No need to rush into it. Thanks for noticing!

304NotModified:
I think the best solution here is to make a specialized build for .NET standard 1 and 1.3 - then the issues could be solved in .NET standard 1.3

FlorianRappl:
Closing due to #534. Thanks for the contribution!

PR #540, opened 2017-02-26
This fixes #539

### LINKED ISSUE #539: Anglesharp can't handle Cookie "Path" properly
Website can set cookie specific to path.  Leatest version of Anglesharp 0.9.9 contains Bug and do not handles cookie path properly.

1. response `http://example.com` sets cookie : `sessionid=ABC1; path=/`
2. response `http://example.com/document` sets cookie: `sessionid=ABC2; path=/document`

Than request `http://example.com/document?somethins=234` should include cookie: `sessionid=ABC; sessionid=ABC2;`

tvildo:
1. We should provide full path to `ICookieProvider `  to handle cookie's properly.
2. We should check for cookie after every response.
3. `HttpWebRequest`  Cookie header should be set manually. We can't use `CookieContainer `  for cookie header extracted from `ICookieProvider`. Because passing cookie like 'sessionid=abc1; sessionid=abc2' again to   `CookieContainer `  makes invalid cookie.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
+++ b/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
@@ -117,7 +117,7 @@
 
             document = await document.QuerySelector<IHtmlAnchorElement>("a").NavigateAsync();
 
-            Assert.AreEqual("UserID=Foo; Auth=Bar", document.Cookie);
+            Assert.AreEqual("Auth=Bar; UserID=Foo", document.Cookie);
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


tvildo:
In my machine all test are passed, Anyone has idea what's wrong ?

FlorianRappl:
Please do not submit PRs against `master`. `devel` is the branch for the upcoming release (see the contribution guide).

Also I think this is a duplicate and already available in `devel`. Please check there first.

(Need to close this as GitHub does not allow re-targeting PRs.)

PR #544, opened 2017-03-11
Do not use recursion for setting Owner and getting BaseUri

### LINKED ISSUE #543: StackOverflowException when parsing many invalid HTML Tags
When parsing html which contains a lot (~21.000) invalid html tags, AngleSharp produces a StackOverflow exception.
Unfortunatly I cannot provide any code to reproduce it, because I am using the [HtmlSanitizer](https://github.com/mganss/HtmlSanitizer) lib, which again uses AngleSharp.

However I can provide the [sample HTML](http://pastebin.com/g4nsaUYh) which produces the stackoverflow and also a [screenshot](https://ibin.co/3ExBplOz9Knt.png) which shows, that the error is caused by AngleSharp itself.

I noticed, that this exception only happens when running the code within an IIS AppPool. When running it within an unitTest instead, the parsing (and sanitizing) runs without any problems in less than one second.

This bug is very crucial, because it can lead to a complete shutdown of an application and stackoverflow-exceptions are not easy to predict/catch.

Can you fix this issue or provide any workaround, so this won't happen when parsing large html within an IIS appPool?

campersau:
fixes #543 

I also added an internal `DescendentsAndSelf` method as suggested [here](https://github.com/AngleSharp/AngleSharp/pull/524#discussion_r98349283) since it is now used in three different places.
(I can also make it public if you want.)

Any other ideas for the test names?
Is this the correct place for these tests?

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/DOM.cs
+++ b/src/AngleSharp.Core.Tests/Html/DOM.cs
@@ -662,5 +662,23 @@
             Assert.AreEqual(data, img.GetAttribute("src"));
             Assert.AreEqual(data, ((IHtmlImageElement)img).Source);
         }
+
+        [Test]
+        public void CreateDeepUnknownElementDocument()
+        {
+            var document = String.Empty.ToHtmlDocument();
+            var html = String.Join("", Enumerable.Repeat("<xyz>", 50 * 1000));
+            document.Body.InnerHtml = html;
+            Assert.True(true);
+        }
+
+        [Test]
+        public void SettingBaseUrlForDeepElements()
+        {
+            var document = String.Empty.ToHtmlDocument();
+            var html = String.Join("", Enumerable.Repeat("<xyz>", 50 * 1000)) + "<img src=\"http://www.example.com/\">";
+            document.Body.InnerHtml = html;
+            Assert.True(true);
+        }
     }
 }
--- a/src/AngleSharp/Css/Dom/SelectorExtensions.cs
+++ b/src/AngleSharp/Css/Dom/SelectorExtensions.cs
@@ -21,16 +21,11 @@
         {
             foreach (var element in elements)
             {
-                if (selector.Match(element, scope))
+                foreach (var descendentAndSelf in element.DescendentsAndSelf<IElement>())
                 {
-                    return element;
-                }
-
-                foreach (var child in element.Descendents<IElement>())
-                {
-                    if (selector.Match(child, scope))
+                    if (selector.Match(descendentAndSelf, scope))
                     {
-                        return child;
+                        return descendentAndSelf;
                     }
                 }
             }
@@ -69,16 +64,11 @@
         {
             foreach (var element in elements)
             {
-                if (selector.Match(element, scope))
+                foreach (var descendentAndSelf in element.DescendentsAndSelf<IElement>())
                 {
-                    result.Add(element);
-                }
-
-                foreach (var child in element.Descendents<IElement>())
-                {
-                    if (selector.Match(child, scope))
+                    if (selector.Match(descendentAndSelf, scope))
                     {
-                        result.Add(child);
+                        result.Add(descendentAndSelf);
                     }
                 }
             }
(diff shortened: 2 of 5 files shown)


FlorianRappl:
Looks great so far. Yes, regarding the tests the location seems fine. I think in the long run this needs to be re-organized / made more structured.

According to AppVeyor some unit tests (10) seem to fail now. All seem to be related to network, i.e., maybe something with the base URL which goes wrong when reading the value.

FlorianRappl:
Regarding the `internal` / `public` one - I think it could indeed be very useful if it is made `public`.

campersau:
I fixed the tests and made `DescendentsAndSelf` public.

FlorianRappl:
Looks great - much appreciated 👍 !

304NotModified:
When will this be released?

FlorianRappl:
There is just one thing missing, which I was / am unable to do (due to time constraints): Provide an elegant way for extension to hook on element insertion.

So far all experiments have suffered from a significant performance loss, which is not an option. Integrating it is easy and straight forward, making it performing is also not terribly difficult, but having an extensible (and elegant) approach that is also performing seems to be non trivial so far.

This is required for the CSS extension (and also scripting, but here its less of a problem) to work properly.

TL;DR: I am behind the schedule but to me its unclear when I'll find the time to release it. I still hope within the next 2 weeks, but I can't guarantee it.

304NotModified:
Can't you move the missing feature to a next release?

FlorianRappl:
Unfortunately not. As described the CSS extension requires it. Otherwise, there is no possibility for users of the new version to bring in CSS support.

Actually, I am already quite pissed that I went too early for the CSS removal (and merged it into `devel`). Yes, the v0.10 release was always thought of as being quite large, but now it is quite massive (lots of fixed and improvements + API changes + CSS outsourced to its own library).

304NotModified:
Clear, thanks. We have to patient then:) success !

vivekjMSFT:
is there a way both these below fixes can be applied on master , until you release v0.10? 
https://github.com/AngleSharp/AngleSharp/issues/406
https://github.com/AngleSharp/AngleSharp/pull/544

both of these issues are catastrophic as one causes all further requests to fail due to corrupt dictionary and other causes the process to crash itself.

FlorianRappl:
Unfortunately not. Cherry picking and a HF would have worked, but the project is lacking the resources to release a HF. Hence, any work should be spent on finalizing the v0.10, such that the fixes are released asap with the v0.10.

vivekjMSFT:
I did send a pull request for master so that this catastrophic issue is fixed with the master. 
https://github.com/AngleSharp/AngleSharp/pull/552

I would suggest if you are ok with the Fix, please release the minor version on top of 0.9.9 so folks can leverage this critical fix. If not I am happy with my fork but please review if you can so I am not breaking anything else :-)

I am sure lot of people facing this issue in their applications.

FlorianRappl:
> There is not going to be a HF for 0.9.9. I know its unfortunate, but the project is lacking the resources to maintain it.

304NotModified:
>  lacking the resources 

Is there something others could help with?

FlorianRappl:
Yes, definitely. Below is a list of all things that need to be done (high-level):

- Homepage
- Documentation
- Releases / QA
- Roadmap / Management
- Development
- Support

Some things are larger blocks (development), others may go hand in hand (homepage, documentation, support).

There are plenty of positions in all of these tasks (e.g., taking full responsibility / being in charge of the homepage, full development of multiple or a single library - could be the core or CSS or ...).

In particular for the v0.10 release the following (mid / low-level) tasks are still to do:

- Finish the latest bits / last failing tests for AngleSharp.Css
- Re-evaluate the service API from the core
- Come up and implement a new proposal for allowing extensions to manipulate the DOM during construction (this is difficult as we need it to be extensible *and* well performing, one of the two would be simple)

So whoever wants to be a part of that / help would be much appreciated.

Issue #546, opened 2017-03-17
Pool with high memory usage on high load


drungrin:
Hi, the Pool implementation can grow really huge on high load. And doesn't go down.
Maybe we can have an option to disable it, or have an idle object limit, or maybe only clearing the StringBuilder  on the ToPoll() can minimize the issue.

Edit: StringBuilder doesn't resize on Clear(). Clearing it still keep the memory.

static ConditionalWeakTable on Node is growing also.

FlorianRappl:
This is definitely right. A cache without an eviction rule is equivalent to a memory leak.

I propose the following:

- Allow defining the maximum of pooled `StringBuilder` instances (library global)
- By default take a max. of 4 `StringBuilder` instances (tbd)
- After a minimum number of `min(2, defined_max)` only consider adding larger instances
- If the maximum number of instances is reached the one with the least capacity is dropped
- `StringBuilder` instances exceeding 2 MB will be dropped

The growth of the weak table would be another issue. However, here the issue can also lie on the user's side (should be checked first). If many documents are created, which are never disposed / released by the user, the weak table will still map the assigned owners.

How does the proposal sound - anything that I missed or should be changed from your point of view?

drungrin:
Sounds good, I believe that'll solve the issue.

About the weak table, I'll need to double check on my side to make sure that I didn't miss something.

drungrin:
The static ConditionalWeakTable on Node is grow was on my side, it is ok for me.

FlorianRappl:
Thanks for the heads up!

FlorianRappl:
Is being fixed in PR #549.

PR #549, opened 2017-03-19
Issue #546

FlorianRappl:
Added more tests to check the novel requirements for the `StringBuilderPool`.

In contrast to #546 the limit has been set to 85k as this is also the LOH limit. That way, such massive string builders are discarded, which should keep the memory clean (LOH fragmentation is still an issue - even though we can now include the LOH for compactification).

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
+++ b/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
@@ -213,7 +213,7 @@
                 var context = BrowsingContext.New(config);
                 var document = await context.OpenAsync(url);
 
-                Assert.AreEqual("k2=v2; k1=v1", document.Cookie);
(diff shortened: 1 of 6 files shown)


jogibear9988:
Wouldn't itbe a good idea, to implement the pool as a non static class, so someone can inject another pool implementation (if needed)?

FlorianRappl:
Problem with that approach is performance. How do you transport that information to all the consumers of this instance? You would essentially do a singleton, where you are free to set the instance. However, that would take away part of flexibility, as you will always set it once for the whole library. The other option comes via the `IConfiguration`, but here the major issue is that some of the consumers are `static` or independent of a browsing context.

This one was not meant to be flexible. That you can configure it now (on a per-library level) is an option if the pool really becomes a problem. Furthermore, it was made `public` such that extension libraries can use the same pool and do not have to rely on their own implementation.

Issue #550, opened 2017-03-20
Wrong "The string did not match the expected pattern" exception


drungrin:
The devel branch is raising this exeption when calling QuerySelector on an empty INodeList. It should return an empty list.

Where:

https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Dom/QueryExtensions.cs#L215-L223

Maybe there and variants should short-return an empty list (or whatever is returned on a not found case) on an empty nodelist

https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Dom/QueryExtensions.cs#L40-L45

FlorianRappl:
You are right here, but the other side (returning the empty list in such cases) may also be wrong. Problem here is that a node is needed to transport the context, which gives access to the parser. Maybe we should use a default parser in such cases - but that could lead to inconsistencies.

I'll see what options come to my mind (I guess we'll need to transport the context explicitly then).

drungrin:
I believe that on master it returned a empty list (not sure). 
You should not need the node to search on an empty node list, just return whatever a "NotFound" would return.

FlorianRappl:
No, that is not the issue. Throwing a syntax exception is. The parser must still be invoked (like in the browser). Hence see my comment above.

drungrin:
Sorry, I didn't got it. Why the parser should be invoked to query a selector over a empty node list? Should not be a optimization to return the "not found" (null on QuerySelector), since the result is known to be "not found"?

FlorianRappl:
Yes, but in order to be *always* consistent the parser must also run *always*. In general, though the `NodeList` is not empty (as such not invoking the parser could be considered a micro optimization). 

Consider `myNodeList.QuerySelector("<invalidSyntax>")`. It should not just return an empty list - as throwing an error would then depend on the state of the `myNodeList` instance. If something is fishy with the syntax it should just throw an exception. BTW: The same is true (and was the motivation) for the `NodeList`.

As a conclusion I think that going for a special exception (invalid operation) for passing in empty `NodeList` may be the right thing to do. It would avoid being inconsistent and it would also avoid being confusing. Again, this part is user invoked (and will never be triggered from the `document` / standard `QuerySelector/All` methods.

drungrin:
Got it about invoking the parser, it should throws on an invalid syntax, and it is needed to check the syntax, being consistent in that way, and I agree with that.

[...]

PR #551, opened 2017-03-20
Issue #550

FlorianRappl:
* Use `scopeNode` as fallback (solves issues with nodes that have no children)
* Throw invalid operation exception if the `NodeList` is empty when no scope has been provided (thus unable to evaluate the syntax)

The latter only applies outside of the "usual" DOM, i.e., in scenarios when we have, e.g., a bare list of nodes that needs to be filtered.

Potentially, it would be better (as this one is outside of the DOM) to just return the empty list (e.g., by returning the `*` selector). However, the issue is, that this way invites inconsistent behavior: We could write code that works until we have a non-empty list. Then a syntax error would be detected. This is quite fishy. 

On the other side we now force users (potentially without knowing) to provide only non-empty lists (or a scope - which is by default `null`, i..e, not demanded).

Maybe there is a third option that stays consistent, is convenient to write, and still allows to avoid passing in non-empty lists or a scope.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/QueryExtensions.cs
+++ b/src/AngleSharp.Core.Tests/Css/QueryExtensions.cs
@@ -14,6 +14,36 @@
         }
 
         [Test]
+        public void QueryOnEmptyNodeListShouldYieldEmptyResult()
+        {
+            var document = GetTestDocument();
+            var result = document.Head.QuerySelectorAll("a");
+            Assert.AreEqual(0, result.Length);
+        }
+
+        [Test]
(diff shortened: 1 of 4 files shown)


FlorianRappl:
In accordance with #550 this was now changed to not throw if an empty list is provided w/o a scope. In such a case either null (match any) or an empty list (match all) is returned. Again, the downside of this approach is that potentially illegal selectors (which would throw otherwise) will not throw. However, given that the other exception would also be unrelated to some syntax validation I guess the given behavior is more sane.

PR #552, opened 2017-03-22
Fix - https://github.com/AngleSharp/AngleSharp/pull/544 for v0.9.0 (M…

vivekjMSFT:
since v0.10 is getting delayed and this is causing catastrophic issue at run time causing process to crash itself, fixing it for master.


FlorianRappl:
There is not going to be a HF for 0.9.9. I know its unfortunate, but the project is lacking the resources to maintain it.

PR #557, opened 2017-05-08
Don't process <img> elements when Source = null or empty

georgiosd:
I was investigating a problem where `ICookieProvider.GetCookie` was called with `origin = null` which didn't seem right. 

Looking into why this happened I discovered that there was a bug where an `img` element with no `src` attribute was getting processed anyway. It only had some classes and a few custom attributes that would later lazy-load the image from the originating page. 

You can see it's a simple fix.

`build.ps1` failed for me but I ran the tests in VS. 5 tests failed but seems to be unrelated to this fix because they fail even if I revert.

### DIFF
--- a/src/AngleSharp/Dom/Html/HtmlImageElement.cs
+++ b/src/AngleSharp/Dom/Html/HtmlImageElement.cs
@@ -133,7 +133,8 @@
         private void UpdateSource()
         {
             var url = this.GetImageCandidate();
-            this.Process(_request, url);
+            if (url != null)
+                this.Process(_request, url);
         }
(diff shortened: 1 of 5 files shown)


FlorianRappl:
Here I definitely cannot accept PRs to `master`. Can you try doing the same on `devel`? Also for `if` statements please add braces - thanks!

The build seems to fail as project.json is no longer valid for the latest .NET Core. Will be updated (migrated) in `devel`.

georgiosd:
Sigh, sorry! I forgot. Will fix.

PR #562, opened 2017-05-15
Bug fixes and enhancements related to AngleSharp.Scripting PR

georgiosd:
So, fixed a bug with `Configuration.Without`, added method to `ErrorEvent` so I can use it from the JS project and fixed the CancellationToken stuff.

### DIFF
--- a/src/AngleSharp/ConfigurationExtensions.cs
+++ b/src/AngleSharp/ConfigurationExtensions.cs
@@ -139,7 +139,7 @@
         /// <returns>The new instance without the services.</returns>
         public static IConfiguration Without<TService>(this IConfiguration configuration)
         {
-            var items = configuration.Services.OfType<TService>();
+            var items = configuration.Services.OfType<TService>().Cast<object>();
             var creators = configuration.Services.OfType<Func<IBrowsingContext, TService>>();
             return configuration.Without(items).Without(creators);
         }
--- a/src/AngleSharp/Dom/Events/ErrorEvent.cs
+++ b/src/AngleSharp/Dom/Events/ErrorEvent.cs
@@ -58,7 +58,7 @@
             private set;
         }
 
-        public void Init(string filename, int line, int column, Exception error)
+        public void Init(String filename, Int32 line, Int32 column, Exception error)
         {
             FileName = filename;
             Line = line;
--- a/src/AngleSharp/Dom/Events/ErrorEvent.cs
+++ b/src/AngleSharp/Dom/Events/ErrorEvent.cs
@@ -52,10 +52,18 @@
         /// Gets the exception describing the error.
         /// </summary>
         [DomName("error")]
-        public DomException Error
+        public Exception Error
         {
             get;
             private set;
+        }
+
+        public void Init(string filename, int line, int column, Exception error)
+        {
+            FileName = filename;
+            Line = line;
+            Column = column;
+            Error = error;
         }
     }
 }
--- a/src/AngleSharp/Io/BaseLoader.cs
+++ b/src/AngleSharp/Io/BaseLoader.cs
@@ -190,6 +190,7 @@
             if (response.StatusCode == HttpStatusCode.Redirect || response.StatusCode == HttpStatusCode.RedirectMethod)
             {
                 method = HttpMethod.Get;
+                headers.Remove(HeaderNames.ContentType);
                 content = Stream.Null;
             }
             else if (content.Length > 0)
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -62,6 +62,7 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>Assets.resx</DependentUpon>
     </Compile>
+    <Compile Include="Auxiliary\ConfigurationsExtensions.cs" />
     <Compile Include="Css\Slickspeed.cs" />
     <Compile Include="Css\WebsiteQuery.cs" />
     <Compile Include="External\MemoryStreamExtensions.cs" />
--- a/src/AngleSharp.Core.Tests/Auxiliary/ConfigurationsExtensions.cs
+++ b/src/AngleSharp.Core.Tests/Auxiliary/ConfigurationsExtensions.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.Linq;
+using AngleSharp.Io;
+using NUnit.Framework;
+
+namespace AngleSharp.Core.Tests.Auxiliary
+{
+    [TestFixture]
+    public class ConfigurationsExtensionsTests
+    {
+        [Test]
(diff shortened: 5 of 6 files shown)


src/AngleSharp/ConfigurationExtensions.cs
```diff
@@ -139,7 +139,7 @@ public static IConfiguration With<TService>(this IConfiguration configuration, F
         /// <returns>The new instance without the services.</returns>
         public static IConfiguration Without<TService>(this IConfiguration configuration)
         {
-            var items = configuration.Services.OfType<TService>();
+            var items = configuration.Services.OfType<TService>().Cast<object>();
```
FlorianRappl, line 142: That change I don't understand. Is that necessary? I don't think so, but maybe I am missing something.

src/AngleSharp/Dom/Events/ErrorEvent.cs
```diff
@@ -52,10 +52,18 @@ public Int32 Column
...
         /// </summary>
         [DomName("error")]
-        public DomException Error
+        public Exception Error
         {
             get;
             private set;
         }
+
+        public void Init(string filename, int line, int column, Exception error)
```
FlorianRappl: Please follow the project's convention here and use the real type names instead of their aliases (`String`, `Int32`).

src/AngleSharp/Dom/Events/ErrorEvent.cs
```diff
@@ -52,10 +52,18 @@ public Int32 Column
         /// Gets the exception describing the error.
         /// </summary>
         [DomName("error")]
-        public DomException Error
+        public Exception Error
```
FlorianRappl, line 55: I would prefer having `DomException` here, as this one would be equivalent to the standard error event. Any other (more generic exception) would be packed into a `DomException`.

FlorianRappl:
Looks good so far! 👍

src/AngleSharp/ConfigurationExtensions.cs
georgiosd: It is. Without the cast I added, the `configuration.Without(items)` line below goes to `Without(object)` when it should go to `Without(IEnumerable<object>` and it fails to remove the object. 

See my test.

src/AngleSharp/Dom/Events/ErrorEvent.cs
georgiosd: https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent says it can be about error from scripts or files. 

I guess it's open to interpretation but I personally think the JS Exception makes more sense.

If you still think `DomException` is the way to go, I'd have to add a `ctor(Exception)`, it doesn't have one currently.

src/AngleSharp/ConfigurationExtensions.cs
FlorianRappl: Good catch! Thanks! *Reminds me to check the chosen overload for indirect matches ...*

src/AngleSharp/Dom/Events/ErrorEvent.cs
FlorianRappl: I guess we can generalize it here - so I'm fine with it 🥇 .

georgiosd:
So:

1) Why don't you pick a name for the event handler and I'll add it to `TaskEventLoop`. Maybe `TaskEventLoop.UnhandledException`? 

2) Is there any reason why you are holding an IEnumerable<object> reference in Configuration? Since it's immutable, it would be safer to hold an object[] - if someone passes an actual IEnumerable<object> there's the danger that it will be enumerated several times. And since the configuration object is not modified a lot, the overhead of creating arrays for every change should be negligible.

3) I also saw that you removed access to the Configuration object in BrowsingContext. I agree it's cleaner but now I don't have the capability to replace services throughout the lifetime of the context. Any smart solutions you can think of? :)

4) Would you object to adding overloads to `DocumentRequest` to accept strings as address too? I hate doing `new Url(address)` all the time :)

FlorianRappl:
1. Sounds good!
2. There was some reason, but I can't remember. I guess it had to do with the quite generic and powerful approach using just the interface. The object array indeed has some advantage, however, note that in v0.10 the original configuration (storing the enumerable) is converted to a fixed array. So you get the flexibility during configuration and then something better / more fixed later.
3. Services should stay during the lifetime of a context. Otherwise, we are doomed as we would need to check for changes at any access. Maybe we can, however, come up with a better system that allows this flexibility and still makes some guarantees if needed.
4. An overload to the constructor or to all construction helpers, e.g., `Get` or `Post`? I guess you are after the latter. The problem here is, that one could argue a string alone is not enough and that `Uri` should also be supported. But then again for both of these types casts and explicit creators for `Url` exist. What is your use case here and wouldn't a helper with extension methods or overloads be a better fit?

georgiosd:
2. So I guess you'd like to keep it that way? The use of the lazy-evaluation did bite me in the behind recently but I can't remember the scenario.

3. I guess you're right. In fact after writing my own loaders with the options I need, it looks like I don't really need it either. 

4. I've just ended with a lot `new DocumentRequest(new Url(foo))` statements because I usually needs to set headers also and the static factories don't have parameters for those. How about an `explicit` string => Url operator plus adding parameters (optional) for headers in the static factories of `DocumentUrl`?

src/AngleSharp/Io/BaseLoader.cs
```diff
@@ -190,6 +190,7 @@ protected static Request CreateNewRequest(Request request, IResponse response)
             if (response.StatusCode == HttpStatusCode.Redirect || response.StatusCode == HttpStatusCode.RedirectMethod)
             {
                 method = HttpMethod.Get;
+                headers.Remove(HeaderNames.ContentType);
```
FlorianRappl, line 193: Can you explain why that is? Where did/does it have an impact?

georgiosd: Well, I have a bunch of tests that check that I create requests that are identical with requests from a SAZ archive (Fiddler) and they kept failing because the `Content-Type` was kept from `POST => redirect (GET)`. 

I think it's semantically wrong - the `Content-Type` is only relevant to a `POST`. If you get redirected which is by definition a `GET`, which has no body, the content type is irrelevant.

Please also give feedback on 2-3-4 above so I can make these changes and we can close this PR :)

FlorianRappl:
Sorry for being late - some feedback:

2. Yes, let's keep it like that for the moment.
3. Keep it then as well.
4. I think there should be already an explicit cast defined. However, Having the overloads may be useful to avoid the explicit cast at all. So please go ahead with the overloads!

src/AngleSharp/Io/BaseLoader.cs
georgiosd: Cool with this too?

georgiosd:
Okie :)

georgiosd:
@FlorianRappl I don't really have time to do the overloads right now - let's merge this one so that it's in sync with the scripting PR and I'll come back to it!

PR #571, opened 2017-06-14
Follow meta refresh (optionally)

georgiosd:
The sanest way I could think to do it :)

### DIFF
--- a/src/AngleSharp/BrowsingContextExtensions.cs
+++ b/src/AngleSharp/BrowsingContextExtensions.cs
@@ -76,16 +76,7 @@
 
             if (loader != null)
             {
-                var download = loader.FetchAsync(request);
-                cancel.Register(download.Cancel);
-
-                using (var response = await download.Task.ConfigureAwait(false))
-                {
-                    if (response != null)
-                    {
-                        return await context.OpenAsync(response, cancel).ConfigureAwait(false);
-                    }
-                }
+                return await loader.OpenAsync(context, request, cancel).ConfigureAwait(false);
             }
 
             return await context.OpenNewAsync(request.Target.Href, cancel).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


FlorianRappl:
First of all thanks for the nice PR 👍 !

This is not how it should be implemented imho. The design I envision is more flexible and would allow more use cases:

- Bring in a new kind of interface, e.g., `IMetaHandler` specifying a function `void HandleContent(IHtmlMetaElement element)`
- On `SetupElement` the `HtmlMetaElement` looks at the available `IMetaHandler` instances (at the configuration of the contained browsing context)
- It invokes all given handlers (i.e., it is possible that multiple handlers can be applied to the same element / data)
- The refresh handler would now actually set the document's location (potentially after a given time; depending on the content of the meta element)

This approach scales well, as the decision if a refresh should occur is purely done on configuration (via a service). It also allows bringing in own refresh handlers (or wrappers of the default one).

The question now is: Do you also think the one given above is a more suitable design or would you disagree that this is the direction this thing should aim for?

georgiosd:
It's definitely much more configurable, I just don't know if it's an overkill for this.

Personally speaking, it's enough to able to say "I want AngleSharp to respect meta refresh (or not)" but I don't know what that answer is for other people or whether my needs will change over time :)

Sleep on it! I needed this for my own project so I thought I'll send a PR anyway.

georgiosd:
One requirement that does up is: I want to be able to switch on/off this behavior per request.

I've also found that some sites use this meta tag inside `<noscript>` to redirect users to a non-JS page but this is obviously problematic. I think search should be limited to `<head>`

FlorianRappl:
> One requirement that does up is: I want to be able to switch on/off this behavior per request.

That you could implement in such a service. Just also include a callback (configurable) to determine if refresh should be stopped.

> I've also found that some sites use this meta tag inside <noscript> to redirect users to a non-JS page but this is obviously problematic. I think search should be limited to <head>

I am not sure. I think the correct behavior is to ignore noscript (content) in JS mode and to consider it in non-JS mode. As all noscript content is ignored in JS mode the thing with the service would work without extra effort.

georgiosd:
Makes sense! I don't know how you're doing with time but if you start a branch and lay out some of this work, I'll help out.

PR #573, opened 2017-06-14
Fixes #572

### LINKED ISSUE #572: BaseRequester events do not properly remove event handlers
In the `devel` branch, the events defined in `AngleSharp.Io.BaseRequester` do not correctly remove event handlers when delegates are removed from the .NET events. On [BaseRequester.cs:20](/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Io/BaseRequester.cs#L20) and [BaseRequester.cs:29](/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Io/BaseRequester.cs#L29), the `remove` handler incorrectly calls `AddEventListener` again instead of calling `RemoveEventListener`.

ArcanoxDragon:
Fix event listener leak caused by `BaseRequester` adding instead of removing event listeners in event `remove` handler.

### DIFF
--- a/src/AngleSharp/Io/BaseRequester.cs
+++ b/src/AngleSharp/Io/BaseRequester.cs
@@ -17,7 +17,7 @@
         public event DomEventHandler Requesting
         {
             add { AddEventListener(EventNames.Requesting, value); }
-            remove { AddEventListener(EventNames.Requesting, value); }
+            remove { RemoveEventListener(EventNames.Requesting, value); }
         }
 
         /// <summary>
@@ -26,7 +26,7 @@
         public event DomEventHandler Requested
         {
             add { AddEventListener(EventNames.Requested, value); }
-            remove { AddEventListener(EventNames.Requested, value); }
+            remove { RemoveEventListener( EventNames.Requested, value); }
         }
 
         /// <summary>


FlorianRappl:
That is a massive finding! Thanks for spotting that issue. 👍

georgiosd:
Nice one @briman0094 !

Hey @FlorianRappl, how about publishing a pre-release nuget? There are many improvements on the project that I'd love to use without having to build the source myself and it looks like there are others in my position :)

FlorianRappl:
Yes, but that pre-release w/o the CSS and with some potential changes is also risky. Also we need to figure out the build system problem.

I think besides these problems it would make much sense.

georgiosd:
I don't really need the CSS parsing so I haven't followed it closely - what's the problem with it? :)

FlorianRappl:
CSS has been outsourced to its own package which has some (~10) failing tests. Most of them are in the area of DOM integration and here I did not find a good solution yet (problem statement was given quite a few times so just a TLDR version: a well performing integration model for arbitrary external code is needed, e.g., in case of adding a new node or removing it; here well performing is with respect to DOM sizes of 10k nodes and more).

georgiosd:
Can u refer me to the description of the problem?

PR #578, opened 2017-06-21
Migrate from .xproj + project.json to .csproj

ArcanoxDragon:
Migrate project to the new VS2017 .NET Core project format. Migrated using official .NET Core migration tool.


ArcanoxDragon:
Project builds with the new project file...I don't know if deployment works but it copies the DLL to the NuGet folder correctly so I would have to assume it should.

Issue #566, opened 2017-06-02
Error running build.ps1: Invalid restore input. UAP projects must contain exactly one target framework.


jpoehls:
I'm trying to run the `build.ps1` script and it is failing. Full output below. Fails on both the `master` and `devel` branches with the same error. Any advice?

```
C:\c\AngleSharp [devel]> .\build.ps1
Feeds used:
  C:\Users\jpoehls\.nuget\packages\
  https://api.nuget.org/v3/index.json
  C:\iwcache\nuget
  C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\

All packages listed in C:\c\AngleSharp\tools\packages.config are already installed.
Analyzing build script...
Processing build script...
Installing addins...
Compiling build script...

----------------------------------------
Setup
----------------------------------------
Executing custom setup action...
Building version 0.10.0 of AngleSharp.
For the publish target the following environment variables need to be set:
  NUGET_API_KEY, GITHUB_API_TOKEN

========================================
Clean
========================================
Executing task: Clean
Cleaning directory C:/c/AngleSharp/src/AngleSharp/bin
Cleaning directory C:/c/AngleSharp/bin/0.10.0
Creating directory C:/c/AngleSharp/bin/0.10.0/nuget
Finished executing task: Clean

========================================
Restore-Packages
========================================
Executing task: Restore-Packages
MSBuild auto-detection: using msbuild version '15.1.1012.6693' from 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\bin'.
All packages listed in packages.config are already installed.
Invalid restore input. UAP projects must contain exactly one target framework. Input files: C:\c\AngleSharp\src\AngleSharp\AngleSharp.Core.csproj, C:\c\AngleSharp\src\AngleSharp\project.json.
An error occured when executing task 'Restore-Packages'.
Error: NuGet: Process returned an error (exit code 1).
```

FlorianRappl:
Thanks to @laurynasr for fixing this one!

PR #579, opened 2017-06-21
Fix memory leak caused by long-duration intervals or timeouts

ArcanoxDragon:
Fixes AngleSharp/AngleSharp.Scripting#41

### DIFF
--- a/src/AngleSharp/Dom/IWindow.cs
+++ b/src/AngleSharp/Dom/IWindow.cs
@@ -10,7 +10,7 @@
     /// https://developer.mozilla.org/en-US/docs/Web/API/Window
     /// </summary>
     [DomName("Window")]
-    public interface IWindow : IEventTarget, IGlobalEventHandlers, IWindowEventHandlers, IWindowTimers
+    public interface IWindow : IEventTarget, IGlobalEventHandlers, IWindowEventHandlers, IWindowTimers, IDisposable
     {
         /// <summary>
         /// Gets a reference to the document that the window contains.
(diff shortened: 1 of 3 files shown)


FlorianRappl:
Thanks for the PR! 👍 
I will look into changing the build.ps1 / Cake for using the csproj again. Will be merged asap then!

🍻

ArcanoxDragon:
I'm actually fiddling around with Cake right now trying to get it to build. It loads the new .NET Core `.csproj` file just fine, but it's looking for the source files in `C:\` instead of the project directory. It may be a local configuration issue for me...I'm still digging around.

ArcanoxDragon:
Technically I believe this can actually be merged in before the project files are migrated, but it'd be good to get those migrated so a prerelease version can be published to NuGet.

FlorianRappl:
Agreed!

ArcanoxDragon:
I just pushed to #578 again...I still can't get the project to build in Cake on my workstation but it builds on my tablet with a fresh clone of my `csproj` branch, so it must be a machine issue.

FlorianRappl:
Looks great - now I will need to look into #566.

ArcanoxDragon:
@FlorianRappl that should be fixed with the migration to `.csproj`, which involved removing the .NET Portable frameworks (as they are effectively deprecated... .NET Core has pretty much replaced portable frameworks).

FlorianRappl:
Hm if it is solved then why is the build failing on AppVeyor? Am I missing something?

ArcanoxDragon:
I changed the project format in a different PR/branch, so this PR is still trying to build with project.json. The code should compile fine in VS but I've noticed all the recent PRs have been failing in AppVeyor because of the .NET Core format change

FlorianRappl:
Yeah sorry I was confused about #578 (two PRs). The last two comments from me actually referred to #578.

FlorianRappl:
This PR of course looks great!
Merged 🥇 !

Issue #512, opened 2017-01-13
Type initializer exception


ChrisMaunder:
I'm getting this occasionally:

> The type initializer for 'AngleSharp.Dom.Html.HtmlStyleElement' threw an exception. (inner: Index was outside the bounds of the array.) 

Stacktrace

```
at AngleSharp.Services.Default.HtmlElementFactory.<>c.<.ctor>b__3_50(Document document, String prefix)
   at AngleSharp.Services.Default.HtmlElementFactory.Create(Document document, String localName, String prefix)
   at AngleSharp.Parser.Html.HtmlDomBuilder.AddElement(HtmlTagToken tag, Boolean acknowledgeSelfClosing)
   at AngleSharp.Parser.Html.HtmlDomBuilder.InHead(HtmlToken token)
   at AngleSharp.Parser.Html.HtmlDomBuilder.InBodyStartTag(HtmlTagToken tag)
   at AngleSharp.Parser.Html.HtmlDomBuilder.InBody(HtmlToken token)
   at AngleSharp.Parser.Html.HtmlDomBuilder.Parse(HtmlParserOptions options)
   at AngleSharp.Parser.Html.HtmlParser.Parse(String source)
```

FlorianRappl:
I think this is a duplicate of #406, which was solved in #489 (PR #491). Should be fixed with the next release.

ChrisMaunder:
We’re on the latest version and we’re seeing this problem today.

Wish I had some data to send you, but our logs don’t have the good bits
saved in this case.

Chris Maunder | Co-founder
chris@codeproject.com | www.codeproject.com

FlorianRappl:
Hi Chris, unless you've build from `devel` and have the 0.10.0 release (which I doubt) sending the data makes no sense. Maybe you have misinterpreted my post (see "Should be fixed with the next release.").

vivekjMSFT:
Hi Chris, did you work out the solution for it, yet? I am thinking to fork the 0.9.9 and apply the fix on top of it since V.0.10 is taking forever to get released and its another bad run time exception.

ChrisMaunder:
No – we've just worked around it.

Funnily enough just last night I was thinking exactly the same thing and I've sent Florian an email today about it: v1 is way overdue.

Cheers
Chris

vivekjMSFT:
[...]

PR #587, opened 2017-07-14
Fixes - 491(406,512), 544 - on top of v0.9.9

vivekjMSFT:
since, V0.10 release is taking forever, these are critical fixes, breaking people's code day & night. I have applied both critical fixes, IMHO , they should be released as hotfix on top v0.9.9 immediately. 

https://github.com/AngleSharp/AngleSharp/issues/512
https://github.com/AngleSharp/AngleSharp/issues/406
https://github.com/AngleSharp/AngleSharp/pull/544


FlorianRappl:
PRs to `master` are normally not allowed. If so it should be submitted to a new branch first (e.g., `hotfix`) and then merged into `master` later (as `devel` is used for v0.10).

Also, you downgrade NUnit. Not sure if its intentional, but it should not be downgraded.

All in all the change should fix the issue as the static constructors are the origin of the problem, but I am not sure of the side effects here. Furthermore, some explanations would be appreciated.

Finally, instead of complaining about v0.9.9 taking long you could either

- fund / sponsor the project,
- contribute to the development, or
- contribute in some other form (e.g., write documentation, motivate others to contribute, ...).

Of course this is all no requirement. Thanks for the nice PR 👍 ! More than welcome. Now we only need to get AppVeyor to work again...

vivekjMSFT:
Thank you for looking at the PR ! Let me know if you are willing to release the hotfix on top of 0.9 , I will go ahead and create hotfix branch and send a pull request for master. Nunit, was giving me issues as I was building with VS and it complains about the compatibility issue with new version. 
I am kinda new to this Appveyor & cake thing but that shouldn't hard to learn. I will be glad to contribute, since I believe this is a fantastic project and have lot of utility in this area. Let me know how can I contribute in (code, documents) to get V1.0 out of the door.

FlorianRappl:
Apparently I can now change the base branch for the PR (it was always freaking me out that I had to close the original PR). So the `hotfix` branch is not the problem anymore.

Anyway, I've also created a milestone for the v0.9.9-hotfix-1 and I want to take this due that seriously. So in one week (or earlier) I want to release this. That includes

- any other PRs / fixes on top of v0.9.9 (no API changes, only fixes)
- a fix for the build system / AppVeyor / .NET standard tooling problem we currently have

I would already merge that one in - but the NUnit change needs to be reverted. Which version of VS are you using? I had no problem using this version of NUnit with VS 2013, 2015 and now 2017. Sometimes I needed the corresponding VS Extension for NUnit 3.

vivekjMSFT:
That would be great ! I am on VS2015.

vivekjMSFT:
ok. I reverted the Nunit change. Please go ahead and merge. Looking forward to the Hotifx release 👍

src/AngleSharp/Common/AttachedProperty.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Runtime.CompilerServices;
```
FlorianRappl: I think a single using statement (inside the namespace) is sufficient.

src/AngleSharp/Dom/Internal/DefaultAttributeObserver.cs
```diff
@@ -0,0 +1,66 @@
+﻿using AngleSharp.Dom.Html;
```
FlorianRappl: Would it be possible to place the using statements inside the namespace? Just to keep the files uniform. Thanks!

src/AngleSharp/Extensions/ParentNodeExtensions.cs
```diff
@@ -1,4 +1,8 @@
-﻿namespace AngleSharp.Extensions
+﻿using System;
```
FlorianRappl: Same here with the using statements.

src/AngleSharp/Foundation/Factory.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace AngleSharp
+﻿using AngleSharp.Dom;
```
FlorianRappl: And here as well!

FlorianRappl:
Hi @vivekscripts thanks for reverting NUnit. I did another sweep over the changes and they look great - thanks. I just found some cosmetic issues with the placement of the using statements. Maybe you can change these few statements - would be awesome, thanks!

👍

vivekjMSFT:
@FlorianRappl , Addressed the review comments. Please go ahead and merge.

vivekjMSFT:
@FlorianRappl are you still going to release the Hotifx or V1.0 is coming with these fixes?

FlorianRappl:
Yes, but the AppVeyor build needs to be done. As long as this one is failing there won't be a release.

PR #604, opened 2017-11-03
Use Encoding.GetEncodings to avoid all the exceptions thrown from missing encodings

zone117x:
This isn't ready for merging (not on dev branch and changes to the csproj format).
Please evaluate only the diff in TextEncoding.cs which avoids the exceptions thrown when encodings are not found. 

VS for Mac (Xamarin Studio) is a huge pain to debug with when catching all exceptions and the AngleSharp lib throws a bunch of them per launch. 

We've been maintaining this single change for a while now on our branch - would be nice to get it upstreamed. Thanks!


FlorianRappl:
What you have in there could not be used in the PCL we have (no such method exists there - I guess you use .NET 4.5 or higher?!).

There have been plenty of discussions about this and most people who suggested such a way - I am of course aware of this and I would have loved to use this, but unfortunately such an option was not given in the decided PCL format.

That being said at the moment (for the 0.10 release and going forward) I am thinking about dropping the PCL + .NET 4 + SL5 support anyway. It has been a pain to support and it seems to be "unwanted" by MS anyway. It will only support .NET Standard (currently 1.0, but maybe a higher version makes more sense to get more APIs). That being written - maybe GetEncodings is then an option again (together with some other things that have been necessary to make it work in the PCL setup).

Issue #637, opened 2018-01-16
The ConditionalWeakTable becomes a bottleneck with parallel html processing


tommymonk:
We've tried master and devel with production work-loads and found that we aren't able to utilize more than ~50% of the CPU.

Our use-case is parsing and running CSS selectors on millions of HTML documents each day. We noticed that no amount of parallelism would allow the majority of the CPU to be utilised.

I decided to try and track down the limitation, the attached code spins up some threads that each parse and query the same html document.

When I profile the application I see that the hot and contentious parts of the code all involve the ConditionalWeakTable used by Node.cs

I'm unable to think of an optimization that would either reduce contention or improve performance.

[cpu_perf_test.txt](https://github.com/AngleSharp/AngleSharp/files/1634853/cpu_perf_test.txt)

FlorianRappl:
There are several things in there to remark.

1. The conditional weak table is obviously also synced / held consistent. So running with multiple simultaneous accessors will obviously have an impact on performance.
2. In real world scenarios IO would be more interesting (i.e., the waiting time to download the page).
3. The question is if running on a single core would actually fully utilize the single core (i.e., 1 thread = 1 / n of the CPU, where we have n cores).

Potentially, the conditional weak table could be replaced with something else. I will need to look into the details, where it is actually used and what impact a replacement would have on the code.

tommymonk:
1) Agreed, I think that the ConditionalWeakTable will have some overhead with high GC.
2) Our application is offloading IO tasks, so this analogue of processing documents continuously is fairly representative.
3) Running on an i7-6700 (4 core, 8 with HT) I see a steady 13% usage which implies it is able to fully utilize the thread

The ConditionalWeakTable seems well suited to it's task, but I wonder how some other collection such as ConcurrentDictionary would compare. It would obviously produce a managed leak should someone forget to dispose documents properly.

FlorianRappl:
> It would obviously produce a managed leak should someone forget to dispose documents properly.

I think this was the original reason for it (not sure anymore though). In the best case there may be some other option that does not come with the described disadvantage.

Joe118:
@FlorianRappl yes, it was indeed the original reason.

[...]

PR #644, opened 2018-02-27
Replace the ConditionalWeakTable tracking Node owners with a simple backing field

comdw:
To address performance issue  #637

### DIFF
--- a/src/AngleSharp/Dom/Node.cs
+++ b/src/AngleSharp/Dom/Node.cs
@@ -1,11 +1,10 @@
-﻿namespace AngleSharp.Dom
+namespace AngleSharp.Dom
 {
     using AngleSharp.Dom.Collections;
     using AngleSharp.Extensions;
     using AngleSharp.Html;
     using System;
     using System.IO;
-    using System.Runtime.CompilerServices;
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


FlorianRappl:
Ok so before approving I wonder how the behavior is of detached nodes that otherwise have not been disposed.

So assuming I do things like `document.Body.InnerHtml = ""` or similar, which essentially removes all nodes from body.

The original reason for introducing the CWT was that .NET / the GC had problems with the cyclic reference (you have document -> node -> document), which could get quite long, e.g., document -> node -> node -> ... -> document. Breaking the cyclic reference by not putting in a direct reference, but rather a CWT solved the issue for me (but as seen, brought in other issues).

Maybe this is gone now anyway; I would just like to see how it behaves in such cases to not introduce a potential memory leak for some performance gain.

comdw:
I've added that line of code to the loop of the performance test and I'm not seeing any difference in the memory usage profile and no evidence of leaks.  Note that this version tends to use more memory during processing when server GC is left off.  With server GC enabled, memory usage is higher for both versions of code.

If you have a more comprehensive test in mind perhaps you'll have chance to run that against this code before approving the PR?

Issue #618, opened 2017-12-14
HTML 5.2


FlorianRappl:
Right now AngleSharp is following the HTML 5.1 specification. With the proposal of HTML 5.2 some (minor) things changed. These should be reflected in AngleSharp as well.

For more information on the changes, see [official W3C specification](https://www.w3.org/TR/html52/changes.html#changes).

PR #655, opened 2018-03-29
implement innerText

campersau:
[`innerText`](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute) is part of the [HTML5.2 spec](https://www.w3.org/TR/html52/changes.html#changes) (See https://github.com/AngleSharp/AngleSharp/issues/618)
See [here](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute) for the documentation of the implementation.

~~There are still tests missing for some CSS cases.~~

### DIFF
--- a/src/AngleSharp/Dom/Html/HtmlElement.cs
+++ b/src/AngleSharp/Dom/Html/HtmlElement.cs
@@ -646,7 +646,7 @@ namespace AngleSharp.Dom.Html
             {
                 var textElement = (IText)node;
 
-                ProcessText(textElement.Data, sb, node.NextSibling == null && node.Parent?.NextSibling == null, parentStyle);
+                ProcessText(textElement.Data, sb, parentStyle);
             }
             else if (node is IHtmlBreakRowElement)
             {
@@ -654,16 +654,26 @@ namespace AngleSharp.Dom.Html
             }
             else if ((node is IHtmlTableCellElement && String.IsNullOrEmpty(elementCss.Display)) || elementCss.Display == "table-cell")
             {
-                if (node.NextSibling != null)
+                var nextSibling = node.NextSibling as IElement;
+                if (nextSibling != null)
                 {
-                    sb.Append(Symbols.Tab);
+                    var nextSiblingCss = nextSibling.ComputeCurrentStyle();
+                    if (nextSibling is IHtmlTableCellElement && String.IsNullOrEmpty(nextSiblingCss.Display) || nextSiblingCss.Display == "table-cell")
+                    {
+                        sb.Append(Symbols.Tab);
+                    }
                 }
             }
             else if ((node is IHtmlTableRowElement && String.IsNullOrEmpty(elementCss.Display)) || elementCss.Display == "table-row")
             {
-                if (node.NextSibling != null)
+                var nextSibling = node.NextSibling as IElement;
+                if (nextSibling != null)
                 {
-                    sb.Append(Symbols.LineFeed);
+                    var nextSiblingCss = nextSibling.ComputeCurrentStyle();
+                    if (nextSibling is IHtmlTableRowElement && String.IsNullOrEmpty(nextSiblingCss.Display) || nextSiblingCss.Display == "table-row")
+                    {
+                        sb.Append(Symbols.LineFeed);
+                    }
                 }
             }
             else if (node is IHtmlParagraphElement)
@@ -724,6 +734,7 @@ namespace AngleSharp.Dom.Html
                 case "IFRAME":
                 case "IMG":
                 case "INPUT":
+                case "LINK":
                 case "METER":
                 case "PROGRESS":
                 case "TEMPLATE":
@@ -805,13 +816,13 @@ namespace AngleSharp.Dom.Html
             }
         }
 
-        private static void ProcessText(String text, StringBuilder sb, Boolean isLastBlock, ICssStyleDeclaration style)
+        private static void ProcessText(String text, StringBuilder sb, ICssStyleDeclaration style)
         {
             var startIndex = sb.Length;
             var whiteSpace = style?.WhiteSpace;
             var textTransform = style?.TextTransform;
 
-            var isWhiteSpace = true;
+            var isWhiteSpace = startIndex > 0 ? Char.IsWhiteSpace(sb[startIndex - 1]) && sb[startIndex - 1] != Symbols.NoBreakSpace : true;
             for (var i = 0; i < text.Length; i++)
             {
                 var c = text[i];
@@ -881,7 +892,7 @@ namespace AngleSharp.Dom.Html
                 sb.Append(c);
             }
 
-            if (isWhiteSpace && isLastBlock) // ended with whitespace
+            if (isWhiteSpace) // ended with whitespace
             {
                 for (var offset = sb.Length - 1; offset >= startIndex; offset--)
                 {
--- a/src/AngleSharp/Dom/Html/HtmlElement.cs
+++ b/src/AngleSharp/Dom/Html/HtmlElement.cs
@@ -10,6 +10,7 @@ namespace AngleSharp.Dom.Html
     using AngleSharp.Services.Scripting;
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Text;
 
     /// <summary>
@@ -518,10 +519,14 @@ namespace AngleSharp.Dom.Html
             get
             {
                 bool? hidden = null;
-                var css = this.ComputeCurrentStyle();
-                if (css != null)
+                if (Owner == null)
                 {
-                    if (!String.IsNullOrEmpty(css.Display))
+                    hidden = true;
+                }
+                if (!hidden.HasValue)
+                {
+                    var css = this.ComputeCurrentStyle();
+                    if (!String.IsNullOrEmpty(css?.Display))
                     {
                         hidden = css.Display == "none";
                     }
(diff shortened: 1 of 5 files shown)


src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +513,270 @@ public Boolean IsTranslated
...
+                        hidden = css.Display == "none";
+                    }
+                }
+                if (!hidden.HasValue)
+                {
+                    hidden = IsHidden;
+                }
+                if (hidden.Value)
+                {
+                    return TextContent;
```
campersau, line 540: Yes, this is strange:
> IF THIS ELEMENT IS NOT BEING RENDERED, OR IF THE USER AGENT IS A NON-CSS USER AGENT, THEN RETURN THE SAME VALUE AS THE TEXTCONTENT IDL ATTRIBUTE ON THIS ELEMENT.

> THIS STEP CAN PRODUCE SURPRISING RESULTS, AS WHEN THE INNERTEXT ATTRIBUTE IS ACCESSED ON AN ELEMENT NOT BEING RENDERED, ITS TEXT CONTENTS ARE RETURNED, BUT WHEN ACCESSED ON AN ELEMENT THAT IS BEING RENDERED, ALL OF ITS CHILDREN THAT ARE NOT BEING RENDERED HAVE THEIR TEXT CONTENTS IGNORED.

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +513,270 @@ public Boolean IsTranslated
...
+                    }
+
+                    ReplaceAll(fragment, false);
+                }
+            }
+        }
+
+        private static void InnerTextCollection(INode node, StringBuilder sb, SortedDictionary<Int32, Int32> requiredLineBreakCounts, ICssStyleDeclaration parentStyle)
+        {
+            if (node is IHtmlTextAreaElement || node is IHtmlInputElement || node is IHtmlVideoElement)
```
campersau: Not sure if there are more elements:

> NOTE THAT DESCENDANT NODES OF MOST REPLACED ELEMENTS (E.G., TEXTAREA, INPUT, AND VIDEO — BUT NOT BUTTON) ARE NOT RENDERED BY CSS, STRICTLY SPEAKING, AND THEREFORE HAVE NO CSS BOXES FOR THE PURPOSES OF THIS ALGORITHM.

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +513,270 @@ public Boolean IsTranslated
...
+            foreach (var child in node.ChildNodes)
+            {
+                InnerTextCollection(child, sb, requiredLineBreakCounts, elementCss);
+            }
+
+            var endIndex = sb.Length;
+
+            if (node is IText textElement)
+            {
+                switch (parentStyle?.TextTransform)
```
campersau: This could be a extension method on `IText` e.g. `GetCssText` but in this case it would decrease perf because it would recalculate the computed styles.

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +513,270 @@ public Boolean IsTranslated
...
+                case "flex":
+                case "grid":
+                case "table":
+                    isBlockLevel = true;
+                    break;
+            }
+            if (!isBlockLevel.HasValue)
+            {
+                // https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements
+                switch (node.NodeName)
```
campersau: This could also be in a separate method or extension method on `IHtmlElement`.

src/AngleSharp/Extensions/WindowExtensions.cs
```diff
@@ -60,6 +61,12 @@ public static CssStyleDeclaration ComputeCascadedStyle(this StyleCollection styl
                 computedStyle.SetDeclarations(inlineStyle.Declarations);
             }
 
+            if (element is IHtmlElement htmlElement && htmlElement.Style != null)
```
campersau: This was a bug! Inline styles from parent nodes were not inherited.

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +514,374 @@ public Boolean IsTranslated
...
+                    requiredLineBreakCounts[sb.Length] = 1;
+                }
+            }
+        }
+
+        private static Boolean HasCssBox(INode node)
+        {
+            switch (node.NodeName)
+            {
+                case "CANVAS":
```
FlorianRappl, line 728: For these groups we already have a quite large collection. Are you sure you do not find a match there? Plus: String switches are ultra slow - can we have hashes instead?

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +514,374 @@ public Boolean IsTranslated
...
+        }
+
+        private static void InnerTextCollection(INode node, StringBuilder sb, Dictionary<Int32, Int32> requiredLineBreakCounts, ICssStyleDeclaration parentStyle)
+        {
+            if (!HasCssBox(node))
+            {
+                return;
+            }
+
+            var elementCss = (node as IElement)?.ComputeCurrentStyle();
```
FlorianRappl, line 615: This will be a problem as in v0.10 AngleSharp.Css is no longer included, but an external lib.

src/AngleSharp/Dom/Html/HtmlElement.cs
```diff
@@ -510,6 +514,374 @@ public Boolean IsTranslated
...
+            get
+            {
+                bool? hidden = null;
+                if (Owner == null)
+                {
+                    hidden = true;
+                }
+                if (!hidden.HasValue)
+                {
+                    var css = this.ComputeCurrentStyle();
```
FlorianRappl, line 528: This will be a problem as in v0.10 AngleSharp.Css is no longer included, but an external lib.

FlorianRappl:
Okay looks nice. I looked into this spec earlier, but was driven away that it is actually a visual spec (i.e., for rendering browsers). So kudos for implementing it.

Now: The question is what to do with it. I would love to have it in 0.9 already, but it will for sure be problematic to have it 1-1 transported to 0.10. Some of the methods you use are just not there. My current proposal would be to carry it actually to AngleSharp.Css and implement it there as an extension method (in JS it would still resolve to the innerText property).

The other thing, however, that we should take care of are these giant switch blocks. They are quite likely very inefficient as these will resolve to primitive string equality. Comparing hashes would be more efficient. I am quite sure that some of these groups already exist and could be reused. What is your opinion?

campersau:
Yes, I can move it to AngleSharp.Css as an extension method if you want. But I am currently using some internal methods. For example [`ReplaceAll`](https://github.com/AngleSharp/AngleSharp/pull/655/files#diff-e665407d7cb5b012c861b087fb3f3f48R599) so we would need `InternalsVisibleTo` for that.

I have actually seen a [some](https://bytes.com/topic/c-sharp/answers/455900-speed-switch-case-vs-hash-table-parsing-xml#post1745170) [posts](https://stackoverflow.com/a/11617866) which say that **larger** `switch` statements are compiled to hash tables. Now the question is if these `switch` statements are large enought. But I can use hash tables or dictionaries directly if you want.

Also I just realized that this implementation does currently not support `&nbsp;`, because this information is not avaiable on `IText` nodes... I have only seen them in the `innerHtml` of the parent node but the parent can obviously have multiple children. Can this information somehow be passed to the `IText` nodes?

FlorianRappl:
> I have actually seen a some posts which say that larger switch statements are compiled to hash tables. 

Cool - I did not know about this optimization. Last time I tried the compiler never optimized for me.

> Also I just realized that this implementation does currently not support &nbsp;, because this information is not avaiable on IText nodes...

As via the spec we are resolving the entities directly - there is no `IEntity` or similar DOM node. The advantage is that reading out text works directly and without much overhead, however, the disadvantage is that we lose this information. There is also no reliable way of restoring this information.

Potentially, we can enhance the tokenizer to remember where the entities have been resolved. Attaching this information to text tokens could attach the information to text nodes, which would now be able to use the info.

> Yes, I can move it to AngleSharp.Css as an extension method if you want. But I am currently using some internal methods. For example ReplaceAll so we would need InternalsVisibleTo for that.

Sounds good. My proposal would be to first have the implementation as proposed here and move it then with 0.10 to AngleSharp.Css. I could actually do the migration. Does that sound fine to you?

campersau:
The cool thing with `&nbsp;` is that it creates `0xa0` chars instead of `0x20` so I was able to add support for it.

So the 0.9.9 version is complete from my side.

src/AngleSharp/Dom/Html/HtmlElement.cs
campersau: What exactly do you mean with:
> For these groups we already have a quite large collection. Are you sure you do not find a match there?

Can you show me where I could find them?

FlorianRappl:
> The cool thing with &nbsp; is that it creates 0xa0 chars instead of 0x20 so I was able to add support for it.

Cool! Very nice job :beer:!

> Can you show me where I could find them?

Yes, of course. Look at [TagNames.cs](https://github.com/AngleSharp/AngleSharp/blob/master/src/AngleSharp/Html/TagNames.cs); it contains combinations of certain elements, which are used throughout the HTML 5 spec. I am not sure if all groups can be found there, but I guess its a start.

campersau:
Sadly these are other groups than the existing ones. `IsBlockLevel` requires `block-level` but the `button` 
 in the [`AllBlocks` group](https://github.com/AngleSharp/AngleSharp/blob/4e24a74548633017b9d49457a80d916c676012fa/src/AngleSharp/Html/TagNames.cs#L835-L840) for example has `display: inline-block` which isn't defined as `block-level`.
But I can add these as new groups if you prefer that.

PR #697, opened 2018-08-06
Do not hide no-response exceptions

justdmitry:
Sometimes target host can't be connected. For example `WebException : No route to host` (with `System.Net.Sockets.SocketException : No route to host` internally). `WebException.Response` becomes null, and `RequestAsync` return null, and any content-related operations return null or empty string, and it's impossible to guess what happened.

I think RequestAsync should re-throw exceptions which does not contain any Response.

### DIFF
--- a/src/AngleSharp/Io/DefaultHttpRequester.cs
+++ b/src/AngleSharp/Io/DefaultHttpRequester.cs
@@ -1,4 +1,4 @@
-﻿namespace AngleSharp.Io
+namespace AngleSharp.Io
 {
     using AngleSharp.Common;
     using AngleSharp.Text;
@@ -180,6 +180,10 @@
                 catch (WebException ex)
                 {
                     response = ex.Response;
+                    if (response == null)
+                    {
+                        throw;
+                    }
                 }
 
                 RaiseConnectionLimit(_http);


justdmitry:
Hmm, that failed test seems not related to my change....

FlorianRappl:
First of all thanks for your contribution.

2 thoughts on the matter:

1. For the failing tests I think they should work (again, at least after pulling in `master` again which diverged). So yes, not related.
2. I'm not sure if it should re-throw. Are there alternatives? I mean a browser also does not crash if no connection cannot be established. On the other hand we need a way to let the developer intercept / react to such an error. Do you think this is the best way or do you see any other option here?

justdmitry:
1) Can't find any new commits in devel and/or master since Jul 16... Why tests should pass now? Or you mean that I should try PR against `master` and not against `devel`?

2) But browser displays its "own" pages with "SSL error", "Can't contact server" etc, not blank page. So, developer here also should have clear sign that "this is not response from target server". Of course one of other options is to make separate/other/fake WebResponse with fake content, describing "problem", but this idea seems very complex and wrong :( "Classic" `HttpClient` has very useful `EnsureSuccessStatusCode` which I can call just to be sure that "everything seems Ok", but I can't see where/how this can be applied in AngleSharp "relatively easy". So re-throwing "unexpected" exceptions seems best choice for me. You may wrap it with your own exception, of course, not sure about pro/cons at this point.

FlorianRappl:
1. yes
2. Again yes - read my post: there must be a way to hook in, but throwing exceptions is definitely not ideal imho; they often get overlooked and force handling on the caller, even though the user may be fine with a default handling (e.g., in case of a browser displaying a generic error in these cases)

justdmitry:
> the user may be fine with a default handling 

We are talking not about end user, but about a developer who want programmatically get some site data. Returning null if "something goes wrong" should not be evaluated from "browser" point of view (browsers do not crash if site is unavailable), but from programming PoV: [exception handling features help you deal with any unexpected or exceptional situations that occur when a program is running](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/). "No connectivity" for parsing library is unexpected situation.

FlorianRappl:
I'm not sure we talking about the same thing @justdmitry. I was not referring to returning null (that's the *current* behavior, not necessarily the *wanted* behavior), but rather different error handing mechanisms. You seem to get stuck with exceptions only. An exception should be a last resort if it cannot continue otherwise, but in here we are free to determine how to continue (and again, a full browser engine may have a default handling for this case inbuilt).

In our case an exception would be pretty devastating, e.g., if the source of an iframe should be loaded the whole iframe crashes (with no possibility of handling this). The biggest issue with exceptions (not only in this case) is the unwinding of the call stack. So unless it's handled directly in library code it will bubble up and destroy potential follow-up actions. Hence the need for a different mechanism for handling errors in that case.

So I'd rather think about this first.

justdmitry:
You mean that somewhere inside library all IFRAMEs are auto-loaded and calling code "does not care" if they loaded successfully or not, and you don't want put try/catch here because of "exception throwing cost"? Ok.

This lead us to decision that calling code should clearly indicate does if it want exceptions to happen, or to be silently ignored, or to be handled in "special way". Something like replacing `IBrowsingContext.OpenAsync` extensions with `OpenOrDieAsync` and `OpenOrRelaxAsync`, so developer should decide which one he wants. Internally, it may be some bool/action property on `DocumentRequest`, I'm not sure about all library internals.

FlorianRappl:
> You mean that somewhere inside library all IFRAMEs are auto-loaded and calling code "does not care" if they loaded successfully or not, and you don't want put try/catch here because of "exception throwing cost"? Ok.

No - I've written an iframe is just loaded by code and if an exception would be thrown from the requester there is no way this could be handled in the library. This has nothing to do with the try-catch cost, rather with the lack of knowledge inside the library (why was an exception thrown? what is the best way to handle the exception? how to inform the user? how can the user handle it there?).

I disagree with the approach - `OpenAsync` is just a helper. You want to cure a problem in the core with a function that lives on the outer shell. That does not work. We need a proper (error) handler in the requester imho.

justdmitry:
I don't understand about IFRAME. You call OpenAsync with url that contain html that contain iframe with src that fails to load. Who parsed root html and initiated iframe load - library? Then code that do this should decide what it should do if loading fails - ignore "problem" or "fail" whole page loading. Of course there is no "one correct decision" for all cases, answer will depend of calling user wish/settings. Such "setting" will come from top-level code that initiated/asked for root page load. That's why I wrote about 'OpenAsync' - this extensions are just helpers and they are source for such params (by function name or by optional param - it's minor details now). Of course this params should dive deeper (where requester will receive it and use inside own try/catch) via something "low-level", may be DocumentRequest (it seems not good enough for such info, but I don't see better one now).

FlorianRappl:
Don't get stuck on the iframe - it was just one example. Really anything that goes through the whole engine and starts at a requester would also fit for the example.

Again, that code cannot do anything here, because if it fails for a subresource all subsequent actions are already invalid (e.g., a script resource fails to load, but then the whole page is destroyed and there is no way to recover from this; even though you would have known how to handle this).

Reason is simple (as stated already): Exceptions unwind the call stack, i.e., in a nesting A B C D of calls, where A was the outer lib function called if something fails at D then also C and B are skipped (even though there was B1 ... Bn statements, if we throw we will skip any remaining parts). Hence the catch on the outside only works if no inner levels with zero follow-up actions exist; this is certainly not the case for a complex browser engine where everything has to follow up according to some spec.

I think the correct way to solve this in the current particular case is by extending the default requester with an optional error handling callback. Then the callback could be requested in all error cases (timeout, network failure, DNS resolution failed, ...).

PR #705, opened 2018-08-21
Add Element DOM method "Closest"

keith-hall:
Hi,

I saw from the project readme that the vision for the project is to have all main JavaScript DOM methods available from C#, and in my uses I found that the standardized [`Element.closest` method](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) was currently unimplemented in AngleSharp.

I implemented this method for my own use, and would like to share it upstream, in the hope that it will get this useful project a little closer to your vision. :)


304NotModified (approved):

304NotModified:
@FlorianRappl maybe add a milestone for this PR? (I see some of the merged PR has a milestone)

PR #706, opened 2018-08-25
invalid character reference &#11111111111111111; 

junyugit:
invalid character reference  &#11111111111111111;  less than 0,   int.ConvertFromUtf32 convert error.

https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Html/Parser/HtmlTokenizer.cs#L548


304NotModified:
I think this change should have an unit test.

Gordory:
Hmm, my pull request isn't first, ok. I will add unit tests to my [PR](https://github.com/AngleSharp/AngleSharp/pull/716).

FlorianRappl:
Fixed already in PR #716 - thanks for your contribution!

Issue #679, opened 2018-06-18
Build on VS2017


atlastodor:
Would it make sense to convert the projects and the build script to VS2017 and .Net Standard Library instead of the deprecated PCL?

I have already branched "devel" branch locally and converted the projects to VS2017 and the cake script (otherwise it wouldn't build). It works for .Net 4.5 and the tests are green. The other platforms are compiled, but I can't say for sure if they work correctly. Also, I can't say if the cake script works on AppVeyor or compiles on non-Windows platforms. This fixes the Nuget version-hell problems. One disadvantage is that VS2017 does not support Silverlight 5 - is this still needed?

Anyways, is there interest to in integrating those changes into the branch? If so, what should I do? I am not used to working on open-source projects on Git-Hub.

FlorianRappl:
Hi @atlastodor thanks for the request. Yes, what you describe is the ~long term~ goal (actually it was, now it should be feasible). There are reasons for the current approach (namely broadest possible compatibility, incl. .NET 4.0 and SL5), however, the downsides etc. are just too high. As a consequence what you describe is wanted (**yes**, we drop support for .NET 4.0 and SL5 and go to a higher .NET Standard than 1.0).

A PR would be appreciated (especially if its from the current `devel` snapshot). Thanks!

tbolon:
Hi,

I have started working on migrating this project to the new csproj format & vs2017.

I just created a PR for the WIP, and would like some feedback from the mainteners of this project.

Known issues:

**some tests are failing**. Two when running using VS 2017, 126 on commandline/appveyor. Need to discover why.

**netstandard 2.0 is the minimum standard version this project can target**. This project can't can't target netstandard1.0 because HttpWebRequest is only available starting [netstandard2.0](https://apisof.net/catalog/System.Net.HttpWebRequest). The previous project was referencing the netcore nuget package which is not recommended (can cause lot of troubles for library consumers. I have targeted both netstandard2.0 and net45 so we need to decide if lower standard must be supported (by changing code to use HttpClient (available on netstandard1.1 for example).

**Need appveyor testing** I can't test an appveyor build for now. Need maintainer help.

tbolon:
Ok found the difference: using command-line, 2 tests fails on Debug configuration, 122 on Release.

tbolon:
Some tests are skipped on my machine, because of the network detection algorithm: "Message: No network has been detected. Test skipped."

tbolon:
[...]

PR #715, opened 2018-10-30
[WIP] Migrate to new csproj and netstandard

tbolon:
I am trying to try to help mgirate this project to netstandard format, using the dotnet crossplatform tooling.

This is a work in progress.

Tracking Issue: #679

- [X] Convert projects to new csproj SDK format
- [X] Remove global.json SDK constraint
- [X] Migrate tests
- [X] Fix tests can not be debugged in VS2017 (breakpoints not hit) -> switch to xunit ?
- [X] Fix tests which does not run on netcore 2.1 (encoding)
- [X] Fix networking tests which does not work anymore (dependency changed)
- [ ] Fix Html.NoScriptTests.NoScriptEatsTooMuch_Issue681 failing


FlorianRappl:
> Fix tests can not be debugged in VS2017 (breakpoints not hit) -> switch to xunit ?

Is xUnit a real fix here? I mean in the end I do not care much about NUnit vs xUnit, just that I want to understand *why* one is working / a better choice than the other before going over all files to change them.

> Fix networking tests which does not work anymore (dependency changed)

Here I need to understand what is failing - then I can support you. Some tests (e.g., the ones going to httpbin) fail regularly when they service operator makes some changes (here ideally we would use our own web app to have a guaranteed stable API). Others may fail due to proxy (e.g., the streaming one can fail here) or other settings.

tbolon:
You can check the last appveyor build, I have added a logger to populate the test result tab. It seems the format of the returns from some sites have changed, need to fix the expected result I think in most cases.

I don't know if it's easier for you, but perhaps you could add my branch as a remote and run the tests on your side ?

I will certainly update the tests to exclude the failing encoding tests from the .net core target.

I will also change the name of the test assembly based on the target framework, it will allow us to track Wich framework test failed when browsing failed tests on appveyor.

tbolon:
> I will also change the name of the test assembly based on the target framework, it will allow us to track Wich framework test failed when browsing failed tests on appveyor.

Reverted back: I can't easily change the name of the test assembly, because of the InternalsVisibleTo. For now there is no way to separate test results from netcore vs netfx.

tbolon:
Ok I have managed to fix all net46 issues, the only test left is NoScriptEatsTooMuch_Issue681 which throw an ArgumentNullException. I can't see how it could pass... I need your help on this one.

I will try to find a way to exclude encoding tests on netcore platform.

The tests still can't be debugged in VS. I can't get why, because I am able to create a project from scratch with the same platform and debug it... Must try to find what is different, perhaps the strong signing.

After that I will certainly look on the cost of replacing HttpWebRequest with HttpClient. If possible, it will allow to target netstandard1.1 instead of netstandard2.0, then most of the job will be done.

FlorianRappl:
I think I will have some time tomorrow or on Thursday to look at all this in detail. Right now what I can tell you regarding HttpClient is that migration should be fairly simple - there is already AngleSharp.Io which uses HttpClient, so we have a working implementation of the IRequester.

Thanks for your work on this so much, worth 1000s of :beers: !

tbolon:
I have finally found the reason why debugging was not working anymore: it was caused by two tests which were using very long constant strings. Moved them to embedded resources did the trick.

There is still one test failing, Html.NoScriptTests.NoScriptEatsTooMuch_Issue681. Since the test provide a default(IScriptingService) as parameter which is always null and the method does a check for null, I do not see how to fix this. I will let you dig this.

alexanderuv:
@FlorianRappl is this PR ready to go? If not, do you guys need any help to finish it up?

alexanderuv:
@FlorianRappl , @tbolon I created another PR #724 which fixes the encoding issues plus enables cross-platform builds (macOS can't have net462 or similar)

PR #716, opened 2018-11-07
Added to HtmlEntityService.IsInvalidNumber method check for 0x0.

Gordory:
This case will shoot when parsing hex code that is bigger than `Int32,MaxValue` from html page.
For examle `<p> This is shoot you down: &#80000000; </p>`.

Actually I would be so happy if you accept this as fast as you can. I need this code and I don't want to use my fork at project. Also I will start an issue describing base bug with overflow.

Also I did`t found any tests on this static class, so I just left as is.

### DIFF
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -107,6 +107,7 @@
     <Compile Include="Html\Doctype.cs" />
     <Compile Include="Html\DomManipulation.cs" />
     <Compile Include="Html\HtmlConstruction.cs" />
+    <Compile Include="Html\HtmlEntityServiceTests.cs" />
     <Compile Include="Html\MainElement.cs" />
     <Compile Include="Html\Namespace.cs" />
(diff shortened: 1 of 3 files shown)


Gordory:
I will add unit tests.

Gordory:
Added unit tests

Gordory:
I can't build solution on local machine, so I can't execute all tests and check that they are corrent. I will try fix my problem and build this. But anyway can you run unit tests on my changes?

Gordory:
Ok, I handle it.
But guys, You are have some failing tests (5) and it's seems to be not my fault.

FlorianRappl:
Yep, I think these tests have something to do with the network -let's ignore them for the moment as they are fixed in develop. I will accept the PR for now going to master (it should go to develop) and take care of all the rest (today I wanted to spend some time finally on the other larger PR to devel).

Thanks for your contribution - much appreciated.

Next steps:

- [x] Make the tests pass again
- [x] Release a new version
- [x] Merge back to develop
- [ ] Merge outstanding PRs to develop
- [ ] Prepare the next release (v0.10)

Hopefully today I can at least do the first three.

Gordory:
Hi there. How release are going? :)

Issue #721, opened 2018-11-17
Create public interfaces for all internal DOM classes


doominator42:
Most of DOM element classes have an interface (e.g. HtmlAnchorElement : IHtmlAnchorElement), but some (e.g. `HtmlStrongElement`, `HtmlItalicElement`, `HtmlNoScriptElement`) does not have one. As those classes are internal, we can't identify which element it is (e.g. `element is IHtmlStrongElement`).

For now, I forked the project and will create the interfaces on my own.
I will create a pull request when done if this change is accepted.

FlorianRappl:
Why do you want to distinguish on the class? What benefit does the interface give you (afaik there is no new functionality)? You can distinguish via, e.g., `TagName` or `NodeName`.

doominator42:
> Why do you want to distinguish on the class?

I code a scraper. It consist of multiple scraper types for each DOM element types. When parsing DOM it calls the appropriate scraper based on element type.

> You can distinguish via, e.g., `TagName` or `NodeName`.

AngleSharp already does this classifying thing by creating appropriate objects for specific tag names.
Then it's natural to do `element is IHtmlSomeElement` rather than `string.Equals(element.TagName, TagNames.Some, StringComparison.OrdinalIgnoreCase)`.
Also some types have multiple associated tag names (e.g. HtmlSemanticElement), so I would have to hardcode a list of tag names for elements which is already done in AngleSharp.

And what's the purpose of having a specific element type if we can't do anything with it because it's internal and it doesn't have a public interface?

FlorianRappl:
> AngleSharp already does this classifying thing by creating appropriate objects for specific tag names.
Then it's natural to do element is IHtmlSomeElement rather than string.Equals(element.TagName, TagNames.Some, StringComparison.OrdinalIgnoreCase).

These classes are internal because they may change - we only publish interfaces that contain functionality and are considered stable (i.e., by following the offical specs).

> Also some types have multiple associated tag names (e.g. HtmlSemanticElement), so I would have to hardcode a list of tag names for elements which is already done in AngleSharp.

This is true, but then you could do the same via reflection and getting the name of the base class. But this is a valid point.

> And what's the purpose of having a specific element type if we can't do anything with it because it's internal and it doesn't have a public interface?

See point above - in the earlier versions we published all the classes but this became a true show stopper (like other things now). Once its out there it needs to be maintained and compatibility for moving things is always an issue.

[...]

PR #722, opened 2018-11-18
Add public interfaces for all internal DOM classes

doominator42:
PR for #721

Please tell if anything is missing or incorrect.
I'm not sure if I used the `DomName` and `DomHistorical` attributes correctly.


doominator42:
Failed tests does not seem to be related to my changes.
3 tests failed because of JSON serialization formatting.
The 2 others are related to POST and PUT requests.

src/AngleSharp/Interfaces/Html/IHtmlFieldSetElement.cs
```diff
@@ -7,26 +7,9 @@
     /// Represents the fieldset HTML element.
     /// </summary>
     [DomName("HTMLFieldSetElement")]
-    public interface IHtmlFieldSetElement : IHtmlElement, IValidation
+    public interface IHtmlFieldSetElement : IHtmlFormControlElement
```
FlorianRappl, line 10: Now this is strange - it gets the same properties like, e.g., `IHtmlButton` Element even though Autofocus and Labels should be missing.

doominator42: This is because `HtmlFieldSetElement` inherits `HtmlFormControlElement`, so it should have the members of `HtmlFormControlElement`, no?

FlorianRappl: Not necessarily. For convenience we made this internally, but publicly only the W3C specified members make sense. A field set does not have labels, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset.

As a rule of thumb - do not expose more than necessary. We can certainly discuss adding some useful public members (e.g., getting the current download for a frame makes sense - this is non-W3C anyway), but do not expose W3C members (e.g., labels on a fieldset) that are simply not there / specified.

FlorianRappl:
Unfortunately after carefully reviewing this PR I cannot accept it. Right now the interfaces represent the W3C interfaces, which are also annotated to guarantee that, e.g., a JS engine build on top of AngleSharp using these interfaces, gets the right IDL directly.

Inserting these "artificial" interfaces shifts the interface tree and places some IDL specified properties on different interfaces. Long story short: Everything breaks apart in this region.

I added the interface for the marquee element though - this one makes sense (and, e.g., Webkit also has this). Thanks for the contribution!

PR #724, opened 2018-11-22
Migrate to .NET Core (netstandard2.0)

alexanderuv:
hi @FlorianRappl, this is based on the work done by @tbolon, I made sure you can build this on macOS and fixed the encoding related failures.

One caveat is that `document.Charset` will not return "Windows-1252" anymore but "windows-1252". This has been adjusted in the tests, and it runs successfully on both Windows and macOS. I don't know if this is something to give more thought to, though.


tbolon:
Are you sure removing the net45 target is mandatory for AngleSharp.Core is mandatory ? There are a lot of known issues regarding netstandard 2.0 and full .NET framework with versions < 4.7.1. Recommendation is to keep delivering multi-targeting versions when possible.

To support building on non windows os, can't we just select the targeting framework when builing on the IDE instead ?

FlorianRappl:
One question: Why do we have 2 PRs now. @alexanderuv do you mind working together with @tbolon to finish this PR? Thanks for your contribution much appreciated!

alexanderuv:
@tbolon I removed it for the sake of consistency. Some projects had net462 and the main one has net45. Since net462 is a superset of netstandard2.0, I chose to just use netstandard2.0.

I was not aware of the netstandard issues. Now, considering that, I updated the code to build both a .NET Framework target and a .NET Standard target. However, in order to keep the encoding issues resolved I figured we should keep the same implementation for both. But in order to do that, I had to upgrade to net46. 

How big is the net45 users base, anyway? Maybe net46 should be fine?

Another thing is - should we now keep two test projects with shared code? One for netstandard and another one for netframework? Since these tests are currently targeting netstandard.

PS: The build in AppVeyor failed due to an env issue? 
```
C:\projects\anglesharp\src\AngleSharp\AngleSharp.Core.csproj(745,3): error MSB4019: The imported project "C:\Program Files\dotnet\sdk\2.1.403\Microsoft\Portable\Microsoft.Portable.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
```

tbolon:
@alexanderuv here is the original issue regarding net461 and netstandard2.0, and all the problems related: https://github.com/dotnet/standard/issues/481

alexanderuv:
@tbolon that's quite the can of worms! Also I reproduced the build issue on my mac. I'll fix this and resubmit

alexanderuv:
@FlorianRappl this is ready for your review now.
I moved the minimum version supported to .net 4.6 as it was the min common version for the added reference. I hope that's ok. Tests are passing in macOS and Windows. Also the Encoding name in the unit tests is now in lowercase (i.e. "windows-1252" instead of "Windows-1252") but it's coming from System.Text.Encoding so it should be consistent between .net framework and .net core

FlorianRappl:
Great one - thanks a lot!

🍻

Issue #761, opened 2019-02-26
Add support for parsing XML processing instructions


daveaglick:
### Description

XML processing instructions are currently commented out during parsing and added to the DOM as comment nodes. The following:

```
<?php foo >
```

Becomes:

```
<!--?php foo -->
```

This proposal is to add optional support for creating `ProcessingInstruction` nodes in the DOM instead of commenting out processing instructions.

### Background

XML processing instructions are a bit of a gray area. They're not technically valid HTML, but being able to parse them and add `ProcessingInstruction` nodes (which already exist) to the DOM would be useful in certain scenarios. As prior art, many web-processing engines already have special support for processing instructions. For example, [CommonMark](https://spec.commonmark.org/0.28/#processing-instruction) has processing instruction support written into the specification.

See https://github.com/Wyamio/Wyam/issues/784

### Specification

This behavior is outside the HTML specification so it will be enabled via a switch in `HtmlParserOptions`.

Note that based on previous conversations I intend to implement this as a PR unless instructed otherwise.

FlorianRappl:
Sounds good! As discussed any contribution on this one would be appreciated. If you need help / want some (design) discussion or where to place / find things - just message either here, on Twitter, or via the Gitter channel (I re-activated it; I'll be there).

FlorianRappl:
Landed in devel - scheduled for next release.

PR #762, opened 2019-02-26
Processing instruction parsing support

daveaglick:
## Types of Changes

This change implements support for parsing XML processing instructions (#761)

### Prerequisites

### Contribution Type

### Description

This introduces a new option `HtmlParserOptions.IsSupportingProcessingInstructions` which causes the parser to emit `ProcessingInstruction` nodes when `<? ... >` tokens are encountered instead of treat them as "bogus comment" content.

Happy to address any feedback you have. Particularly interested in a look at how I implemented tokenizing the processing instruction and splitting it's target/data after tokenization in a `ProcessingInstruction.Create()` factory method - there might be a more "correct" way to do this. Also happy to add any additional tests you think might be warranted.

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/ProcessingInstructions.cs
+++ b/src/AngleSharp.Core.Tests/Html/ProcessingInstructions.cs
@@ -1,0 +1,33 @@
+namespace AngleSharp.Core.Tests.Html
+{
+    using AngleSharp.Html.Parser;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class ProcessingInstructions
+    {
+        [Test]
+        public void ParsingWithSupportForProcessingInstructionsShouldProduceNode()
+        {
+            var source = @"<?foo bar>";
+            var parser = new HtmlParser(new HtmlParserOptions
+            {
+                IsSupportingProcessingInstructions = true
+            });
+            var document = parser.ParseDocument(source);
+            Assert.AreEqual(Dom.NodeType.ProcessingInstruction, document.ChildNodes[0].NodeType);
+        }
+
+        [Test]
+        public void ParsingWithoutSupportForProcessingInstructionsShouldCommentThemOut()
+        {
+            var source = @"<?foo bar>";
+            var parser = new HtmlParser(new HtmlParserOptions
+            {
+                IsSupportingProcessingInstructions = false
+            });
+            var document = parser.ParseDocument(source);
+            Assert.AreEqual(Dom.NodeType.Comment, document.ChildNodes[0].NodeType);
+        }
+    }
+}
--- a/src/AngleSharp/Dom/Internal/ProcessingInstruction.cs
+++ b/src/AngleSharp/Dom/Internal/ProcessingInstruction.cs
@@ -40,6 +40,20 @@ namespace AngleSharp.Dom
             return node;
         }
 
+        /// <summary>
+        /// Creates a processing instruction by splitting data into the name/target and data.
+        /// </summary>
+        internal static ProcessingInstruction Create(Document owner, String data)
+        {
+            int nameLength = data.IndexOf(' ');
+            var pi = new ProcessingInstruction(owner, nameLength <= 0 ? data : data.Substring(0, nameLength));
+            if (nameLength > 0)
+            {
+                pi.Data = data.Substring(nameLength);
+            }
+            return pi;
+        }
+
         #endregion
     }
 }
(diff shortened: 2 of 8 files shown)


FlorianRappl:
Great PR! All in all looks quite good. The only thing I would love to discuss is the duplicated sections in the DOM builder.

I think it may be a better design to no introduce a PI token, but rather to extend the comment token (e.g., with a `IsProcessingInstruction` property). Then the DOM builder could 100% stay the same (minus setting up the tokenizer etc.) and then use:

```cs
public static void AddCommentOrProcessingInstruction(this Document parent, HtmlToken token)
{
  if (token.IsProcessingInstruction)
  {
    parent.AddNode(ProcessingInstruction.Create(parent, token.Data));
  }
  else
  {
    parent.AddNode(new Comment(parent, token.Data));
  }
}
```

(*Note*: Coded in the comment box, no guarantee on actual compilation given, e.g., I know the passed type may be wrong / we need to use (or introduce?) a specialized token class here.)

What do you think? Would that make sense? (I am just afraid that if we have duplicates a change in the HTML spec will eventually lead to forgetting this non-spec thing to be properly maintained - hence I try to make it as easy / implicit as possible)

daveaglick:
Thanks for the quick feedback. Extending the comment token to support PIs makes perfect sense - it may not be _technically_ correct spec-wise since there's no relationship between PIs and comments, but practically speaking it should greatly reduce the amount of code needed to handle PIs if that option is turned on.

I got swamped at the day job so lost momentum on this, but I'll come back to it and rework based on your feedback as soon as I can (hopefully early next week).

daveaglick:
Finally got a chance to revisit this PR. I'm a little stuck with the feedback though. As far as I can tell, all tokens are implementations of the `HtmlToken` class which has a `Type` property of type `HtmlTokenType`. So "introducing a new token type" meant I created a new value in the `HtmlTokenType` enumeration and then keyed off of that in the DOM builder. If I represent processing instructions as comment tokens with a special `IsProcessingInstruction` flag, where does that flag go? Were you thinking it would be a property in `HtmlToken` that's only relevant when `HtmlToken.Type == HtmlTokenType.Comment` (and ignored otherwise)? I think that makes sense, but just wanted to be sure.

daveaglick:
Okay, went ahead and refactored as discussed above before I forgot what I was doing. I left the old commit there in case I need to go back. Let me know if you want me to squash before you merge.

FlorianRappl:
I think looks great. Let's go ahead and merge? :beers:

daveaglick:
Thanks!

FlorianRappl:
(I planned this for v1, but if something comes up to stop v1 I would release it earlier - its in `devel` anyway, which means its in the next release...)

(next release would be eo March or eo April - if you want me I can try to release a preview NuGet as well - earlier)

daveaglick:
Nope, no need for an early release. I've got a workaround in place with a note to remove it once this ships, so I'm good.

daveaglick:
Quick question about versions - my AngleSharp reference has finally been updated and it seems like this didn't make it into 0.14.0? I've got a custom formatter that works around the issue, but when I remove it after the update, the processing instructions are still being commented (which is what this PR was intended to resolve). It's no problem at all for me to continue using the custom formatter if this change didn't make it into 0.14.0, just want to make sure I'm not holding it wrong. Thanks!

FlorianRappl:
Well it should be there. It was in `develop` and the PR was over a year ago.

Let me check.

FlorianRappl:
Well, I see it in the `HtmlParserOptions` etc.

https://github.com/AngleSharp/AngleSharp/blob/master/src/AngleSharp/Html/Parser/HtmlParserOptions.cs#L54

Hm, I wonder where the difference is etc. -> the PR seems to be in there.

daveaglick:
Ah, so I _was_ holding it wrong :)

I do see the option, maybe the problem is that it's opt-in:

image

Let me try setting the option and see if that removes the need for my formatter...

daveaglick:
Doh! Setting the option _does_ change the behavior...

Without `IsSupportingProcessingInstructions`:

image

With:

image

But it looks like it's stripping the angle brackets from around the processing instruction when writing with `.ToHtml()` and the default `HtmlMarkupFormatter.Instance`. Unless you've got any quick insight, I'll start digging deeper to try and uncover what's going wrong.

FlorianRappl:
Yeah its opt-in by design, because out of the box AngleSharp *has* to W3C compliant (and this is not in the spec. / valid HTML5).

Regarding the issue. I have no direct clue, but I think its somewhere in the comment representation. It takes the inner part. So we need to adjust this comment with the brackets I guess.

I can also try to solve it if we agree on the desired behavior. If you could stretch out 4-5 cases (input -> serialization) I'd implement the test cases and make them succeed!

PR #785, opened 2019-04-29
Creating a selector for the element

lvkins:
Added an extension method for the elements that creates a unique selector path used to locate the element in the DOM.


FlorianRappl:
Looks great - thanks for your contribution :beers:!

PR #799, opened 2019-06-06
Bugfix: invalid characters after idn-mapping

hyspace:
## Types of Changes

Bugfix

### Contribution Type

### Description

1. removed reading string to to byte and reading byte back to string since it is not doing anything useful
2. separate the combined parsing steps since it was generating wrong result.
3. changed `#if NETSTANDARD2_0 || NET46` to `#if !NETSTANDARD1_3` for better forward compatibility

### DIFF
--- a/src/AngleSharp.Core.Tests/Urls/Url.cs
+++ b/src/AngleSharp.Core.Tests/Urls/Url.cs
@@ -209,6 +209,13 @@ namespace AngleSharp.Core.Tests.Urls
         }
 
         [Test]
+        public void PunycodeSquareReplacement_Issue797_InvalidAfterMapping()
+        {
+            var url = new Url("http://www.examp？le.com/");
+            Assert.IsTrue(url.IsInvalid);
+        }
+
+        [Test]
         public void InvalidRelativeUrlAsDifferentProtocolScheme()
         {
             var baseUrl = new Url("http://some.domain.com");
--- a/src/AngleSharp/Url.cs
+++ b/src/AngleSharp/Url.cs
@@ -1162,11 +1162,7 @@ namespace AngleSharp
                         sanatizedHostName = hostName.Substring(start, length);
                         return false;
                     default:
-                        if (cc.IsAlphanumericAscii() || cc.IsOneOf(Symbols.Minus, Symbols.Underscore, Symbols.Dot))
-                        {
-                            buffer.Append(Char.ToLowerInvariant(cc));
-                        }
-
+                        buffer.Append(Char.ToLowerInvariant(cc));
                         break;
                 }
             }
--- a/src/AngleSharp/Url.cs
+++ b/src/AngleSharp/Url.cs
@@ -1132,6 +1132,7 @@ namespace AngleSharp
             }
             catch (ArgumentException)
             {
+                buffer.ToPool();
                 sanatizedHostName = hostName.Substring(start, length);
                 return false;
             }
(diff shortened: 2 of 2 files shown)


FlorianRappl:
With this PR we are not aligning to WHATWG anymore as

```cs
            // https://url.spec.whatwg.org/#host-parsing 3.5.7
            // forbidden host code point check
```

and follow up code is for NETSTANDARD_13 only.

Example:

```cs
new System.Globalization.IdnMapping() { AllowUnassigned = false, UseStd3AsciiRules = false }.GetAscii("foo:yo")
```

results in `foo:yo`. Likewise for other characters such as `:`, `@`, `_`, ` `, ... that was the reason for still iterating over the characters.

Otherwise looks good to me!

hyspace:
There is a confusion here.
```cs
new System.Globalization.IdnMapping() { AllowUnassigned = false, UseStd3AsciiRules = false }.GetAscii("foo:yo")
```
will return `foo:yo` first, but then `:` will cause failure [here](https://github.com/AngleSharp/AngleSharp/pull/799/files#diff-6c477f3f94502e73400990fc5a620b05R1153)

The output change in this PR is:
For `.Net 2.0` and `.Net framework 4.6`, when there is any invalid characters after `percent decoding` and `IdnMapping`, `AngleSharp.Url` will return a failure instead of those invalid characters being ignored. Note that the invalid character can be introduced by `percent decoding` or `IdnMapping`, which is the missing part of original code. For example, `%0A` will become `\n` and then cause failure. `？` will be mapped to `?` and then cause failure. That is why the *forbidden host code point check* need to be done at the end.
(As we discussed last time, `_` will still be considered valid for loose validation.)

Code block for `.Net 1.3` will filter out all invalid characters (also valid Unicode which is not defined in [hard coded mapping](https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Text/Punycode.cs#L26)). Line 1139-1174 is going to do nothing in `.Net 1.3`, since all mapped character defined in *hard coded mapping* are valid characters.

There should not be any other output change in this PR.

FlorianRappl:
Right - thanks for the clarification!

hyspace:
Could you verify if I'm using `StringBuilderPool` correctly in [Line 1159](https://github.com/AngleSharp/AngleSharp/pull/799/files#diff-6c477f3f94502e73400990fc5a620b05R1159)?

Not sure if I should put it back to pool before return failure.

FlorianRappl:
This usage is fine. Note - there cannot be a memory leak, just a performance degradation. I'll go over the full use to see if we properly return the `StringBuilder` instance to the pool.

FlorianRappl:
Looks fine - on line 1133 I would insert another `ToPool` call (otherwise in this exception case we are not returning the instance back to the pool).

hyspace:
Thanks. I will implement ipv4 and ipv6 soon.

hyspace:
Please hold on this PR, I saw we did not include some tests about hostname in [WPT](https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json) since we are missing the ability to parse Unicode. I will add those tests back and update this PR.

FlorianRappl:
Yes - but maybe split the PR in two parts (have the IPv4/6 in at least one other PR). Note my comment above; currently we are not returning the `StringBuilder` to the pool in one possible branch (line 1133).

hyspace:
Sorry I missed your comment. Fixed.

FlorianRappl:
No problem - thanks for the fix. Shall we merge this one?

hyspace:
Yes, I will make IPV4 and IPV6 separate PR.

hyspace:
Resolved conflicts

PR #806, opened 2019-06-27
Percentage decoding hostname

hyspace:
### Contribution Type

### Description
With the [help from WPT tests](https://github.com/web-platform-tests/wpt/issues/17420) we are able to fix [failing test](https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json#L4689) of input `https://%e2%98%83`.

Root cause is my previous PR https://github.com/AngleSharp/AngleSharp/pull/799 is wrong in *percent decoding*.

I removed the string to byte and byte to string operation in that PR, but actually it is required but in a different way (this test will fail even before https://github.com/AngleSharp/AngleSharp/pull/799).

### DIFF
--- a/src/AngleSharp.Core.Tests/Urls/UrlValidation.cs
+++ b/src/AngleSharp.Core.Tests/Urls/UrlValidation.cs
@@ -4182,6 +4182,25 @@ org");
         }
 
         [Test]
+        public void DocumentUrlShouldDoUtf8PercentDecoding()
+        {
+            var document = Html("<base id=base>");
+            var element = document.GetElementById("base") as HtmlBaseElement;
+            Assert.IsNotNull(element);
+            element.Href = @"about:blank";
+            var anchor = document.CreateElement<IHtmlAnchorElement>();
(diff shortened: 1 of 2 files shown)


FlorianRappl:
Another great PR - you rock :beers:!

I'll try to publish 0.13 of AngleSharp in about 1-2 weeks; any blockers you see there?

hyspace:
If you took a look at the WPT issue I raised, there still one failing test about IDN mapping of input
`https://faß.ExAmPlE/`

I don't think it is a blocking issue since even Chrome have wrong parsing result today.

BTW, I'm tracking WPT test failures [here](https://gist.github.com/hyspace/8f79320fd0b6d56af54505400f268e49), and I'm trying to label each failure with tags like [critical] [custom scheme] and I will fix them in order of importance.

I have already fixed all failure with [critical] tag except for `https://faß.ExAmPlE/`.

If you have any concern, you can go through current failures and see if we need to fix any of them before new release.

PR #824, opened 2019-09-20
Update Examples.md

xp44mm:
the `Text()` in `AngleSharp.Dom`.

### DIFF
--- a/doc/Examples.md
+++ b/doc/Examples.md
@@ -142,7 +142,7 @@ static async Task SingleElements()
     Console.WriteLine("Only from C# / AngleSharp:");
     Console.WriteLine();
     Console.WriteLine(emphasize.ToHtml());   //<em> bold <u>and</u> italic </em>
-    Console.WriteLine(emphasize.TextContent);   // bold and italic
+
 
     Console.WriteLine();
     Console.WriteLine("From the DOM:");
@@ -153,9 +153,9 @@ static async Task SingleElements()
 }
 ```
 
-The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags), the other two variants behave like the official W3C specified properties, leaving everything as it is.
+The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. ~~The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags),~~ the other two variants behave like the official W3C specified properties, leaving everything as it is.
 
-Extension methods such as `ToHtml()` and `Text()` can be found in the namespace `AngleSharp.Extensions`.
+~~Extension methods such as `ToHtml()` and `Text()` can be found in the namespace `AngleSharp.Extensions`.~~
 
 ## Connecting JavaScript Evaluation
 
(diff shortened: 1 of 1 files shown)


xp44mm:
@CLAassistant 
> ✅ xp44mm
> ❌ cuishengli
is same person. but cuishengli email is virtual. no way to add it.

FlorianRappl:
Thanks for the PR / corrections :beers:!

Who is `cuishengli`?

> is same person. but cuishengli email is virtual. no way to add it.

Yes, but legally we require that the email / origin is clear. What you can do is to change the commits to come from the same email address.

doc/Examples.md
```diff
@@ -142,7 +142,7 @@ static async Task SingleElements()
     Console.WriteLine("Only from C# / AngleSharp:");
     Console.WriteLine();
     Console.WriteLine(emphasize.ToHtml());   //<em> bold <u>and</u> italic </em>
-    Console.WriteLine(emphasize.TextContent);   // bold and italic
+
```
FlorianRappl: Why is there now a blank line? Can you please keep the `emphasize.Text()` that you used earlier? I think it would make the most sense.

doc/Examples.md
```diff
@@ -153,9 +153,9 @@ static async Task SingleElements()
 }
 ```
 
-The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags), the other two variants behave like the official W3C specified properties, leaving everything as it is.
+The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. ~~The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags),~~ the other two variants behave like the official W3C specified properties, leaving everything as it is.
 
-Extension methods such as `ToHtml()` and `Text()` can be found in the namespace `AngleSharp.Extensions`.
+~~Extension methods such as `ToHtml()` and `Text()` can be found in the namespace `AngleSharp.Extensions`.~~
```
FlorianRappl: I would not strike through anything. Either leave it or remove it. I like the earlier change using the `AngleSharp.Dom` namespace more.

doc/Examples.md
```diff
@@ -153,9 +153,9 @@ static async Task SingleElements()
 }
 ```
 
-The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags), the other two variants behave like the official W3C specified properties, leaving everything as it is.
+The output commands try to demonstrate the differences between several ways of getting back strings from nodes. In fact the DOM property `OuterHtml` uses the `ToHtml()` version to generate the HTML code. ~~The other variants are all different. While `Text()` is just a helper to strip text (and omit unwanted text contents like in `<script>` tags),~~ the other two variants behave like the official W3C specified properties, leaving everything as it is.
```
FlorianRappl: I guess this one needs to be rewritten.

FlorianRappl (approved):

FlorianRappl:
Thanks for the changes :beers:!

xp44mm:
maybe when i reset the commits, something was go wrong. my repo's file is as same as earlier, and i have reviewed myself.  do you note that your window's top has  yellow `outdated`?

FlorianRappl:
> maybe when i reset the commits, something was go wrong. my repo's file is as same as earlier, and i have reviewed myself. do you note that your window's top has yellow `outdated`?

Outdated on GitHub means that the review was done for an earlier version than the current state and that the respective position changed.

In this case all the review points I mentioned have changed, so all of these snippets are outdated.

xp44mm:
thank you let my known those, next i will work correctly.

Issue #831, opened 2019-10-30
Consider updating the .NET dependency version to 4.6.1


skolima:
# Bug Report

## Prerequisites

- [ ] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version of AngleSharp?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository? (there are multiple AngleSharp libraries, e.g., `AngleSharp.Css` for CSS support)
- [x] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

## Description

The package https://www.nuget.org/packages/AngleSharp declares NuGet dependencies for .NET Framework 4.6 and .NET Standard 2.0, without declaring 4.6.1 dependencies. This is against https://docs.microsoft.com/en-us/dotnet/standard/net-standard guidelines and sometimes results in difficult to hunt down build/restore issues.

For me, this breaks PreMailer update - the previous one depended on AngleSharp 0.9.1, which declared only pre-.NETStandard2.0 dependencies and restored correctly. The current AngleSharp triggers the buggy restore mechanisms for .NETFramework4.6

**Expected behavior:** NuGet gallery dependencies listed for .NETFramework 4.6.1 and .NETStandard 2.0

**Actual behavior:** .NETFramework 4.6 is listed as a valid target, while it has known bugs when restoring packages

## Possible Solution

Update build and NuGet package targets to net461 instead of net46.

FlorianRappl:
Where do we start where do we end? I mean this sounds like a bug in Nuget with the package resolution.

Our intention here was to say "if its .NET Framework 4.6 or higher you can use this".

If you have a specific proposal then please make a PR - any contribution / discussion welcome :beers:!

FlorianRappl:
Landed in `devel`.

PR #832, opened 2019-10-30
Target .NETFramework4.6.1 instead of 4.6

skolima:
Resolves https://github.com/AngleSharp/AngleSharp/issues/831

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -3,7 +3,8 @@ var projectName = "AngleSharp";
 var solutionName = "AngleSharp.Core";
 var frameworks = new Dictionary<String, String>
 {
-    { "net461", "net461" },
+    { "net46", "net46" },
+    { "net472", "net472" },
     { "netstandard1.3", "netstandard1.3" },
     { "netstandard2.0", "netstandard2.0" },
 };
--- a/build.cake
+++ b/build.cake
@@ -3,7 +3,7 @@ var projectName = "AngleSharp";
 var solutionName = "AngleSharp.Core";
 var frameworks = new Dictionary<String, String>
 {
-    { "net46", "net46" },
+    { "net461", "net461" },
     { "netstandard1.3", "netstandard1.3" },
     { "netstandard2.0", "netstandard2.0" },
 };
--- a/src/AngleSharp.nuspec
+++ b/src/AngleSharp.nuspec
@@ -22,8 +22,11 @@
       <group targetFramework="netstandard2.0">
         <dependency id="System.Text.Encoding.CodePages" version="4.5.0" />
       </group>
-      <group targetFramework="net461">
+      <group targetFramework="net46">
         <dependency id="System.Text.Encoding.CodePages" version="4.5.0" />
+      </group>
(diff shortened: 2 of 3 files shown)


FlorianRappl:
What about .NET 4.6 users? This seems to be rather hostile (and is breaking!).

skolima:
It is due to tooling bug which shipped in "plain" 4.6. If you care much about backward compatibility, 4.5 target can also be included safely. The "net46" is broken, though, as it shipped before .NETStandard2.0 was fully finalized. Microsoft has been so far suggesting bumping to 4.6.1 in order to force users off the broken 4.6.0.

FlorianRappl:
Just for completeness - if we could just target 4.5 we would do it. There is a reason why we went to 4.6 and not 4.5. Regarding 4.6.1 - why would we bother with this and not go to, e.g., 4.7.2 directly?

I can also rephrase: Why should we ship something for a broken system any way?

Now that is quite controversial, but where I was going with this is that there are existing users with 4.6 - and they should still be able to get AngleSharp.

We don't want to force users in any direction. We want to support what is possible (and doable / manageable) to support.

Do you have any link regarding the net46 claim? I don't see any evidence and this was IIRC never discussed (keep in mind I'm an MVP and I'm in the internal mailing lists; even though possible it is in my opinion unlikely that an issue of that magnitude was not discussed in greater detail). I would love to learn more. What I know is that 4.6.1 has problems with the NS2.0 compatibility. The compatibility was added later and that was a mistake - so some things broke / are broken.

skolima:
I'd be also 100% happy if you'd opt for net46+net472 where the kinks have been iron out, this works for me as well (tried with several other packages, and our internal tooling - went all in on net472 as we don't have internal users on 4.6x). The problematic scenario for me is if package advertises (net46+netstandard20) - in this case 4.7.x tooling goes for net46 as better match.

I've got trouble finding links, unfortunately - my written sources in this case are based mostly on twitter conversations between David Fowler, Immo Landwerth, Marc Gravell ( https://www.nuget.org/packages/Dapper/ ) and Bruno Garcia ( https://www.nuget.org/packages/Sentry ) and what came out of it in terms of targeting their packages - but obviously twitter is terrible for back-references.

Ah, found a decent link, I guess: https://github.com/getsentry/sentry-dotnet/issues/135#issuecomment-438599084

FlorianRappl:
Thanks for taking the time to dig deeper. I think this is the issue I'm also recalling (thus my reference to 4.6.1 being broken).

I guess the way out of this could be to:

- Stay with the NS1.3 and NS2.0
- Stay with the 4.6
- Have a build for 4.7.2

Thus 4.6 and 4.6.1 users will get the same experience, but for 4.7.2 onwards the experience should be better.

Would you agree or do you have a different approach in mind?

skolima:
That works for me. I'll update the PR.

FlorianRappl:
LGTM - let me know if that works!

Also thanks for the contribution - much appreciated 🍻 !

Issue #850, opened 2020-03-10
Url.GetHashCode is not in agreement with Url.Equals


alex-fomin:
# Bug Report

## Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version of AngleSharp?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository? (there are multiple AngleSharp libraries, e.g., `AngleSharp.Css` for CSS support)
- [x] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

## Description

`Url.GetHashCode()` returns `base.GetHashCode()` which is based on object reference, while `Equals` actually compare url's parts.

## Steps to Reproduce
```
var url1 = new Url("https://example.com");
var url2 = new Url("https://example.com");
Console.WriteLine(url1.GetHashCode() == url.GetHashCode()); // False
Console.WriteLine(url1.Equals(url); // True

```
## Possible Solution

Implement `GetHashCode` based on `Equals` method.

FlorianRappl:
Available in `devel` and the preview version.

PR #851, opened 2020-03-10
GetHashCode now corresponds with Equals

alex-fomin:
# Types of Changes

## Prerequisites

## Contribution Type

## Description
Fixing #850 

`Url.GetHashCode()` returns `base.GetHashCode()` which is based on object reference, while Equals actually compare url's parts.

### DIFF
--- a/src/AngleSharp/Url.cs
+++ b/src/AngleSharp/Url.cs
@@ -321,7 +321,19 @@ namespace AngleSharp
         /// <returns>A hash code for the current url.</returns>
         public override Int32 GetHashCode()
         {
-            return base.GetHashCode();
+            unchecked
+            {
+                var hashCode =  _fragment != null ? StringComparer.Ordinal.GetHashCode(_fragment) : 0;
+                hashCode = (hashCode * 397) ^ (_query != null ? StringComparer.Ordinal.GetHashCode(_query) : 0);
+                hashCode = (hashCode * 397) ^ (_path != null ? StringComparer.Ordinal.GetHashCode(_path) : 0);
+                hashCode = (hashCode * 397) ^ (_scheme != null ? StringComparer.OrdinalIgnoreCase.GetHashCode(_scheme) : 0);
+                hashCode = (hashCode * 397) ^ (_port != null ? StringComparer.Ordinal.GetHashCode(_port) : 0);
+                hashCode = (hashCode * 397) ^ (_host != null ?  StringComparer.OrdinalIgnoreCase.GetHashCode(_host) : 0);
+                hashCode = (hashCode * 397) ^ (_username != null ? StringComparer.Ordinal.GetHashCode(_username) : 0);
+                hashCode = (hashCode * 397) ^ (_password != null ? StringComparer.Ordinal.GetHashCode(_password) : 0);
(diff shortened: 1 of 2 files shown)


src/AngleSharp/Url.cs
```diff
@@ -321,7 +321,19 @@ public String Query
         /// <returns>A hash code for the current url.</returns>
         public override Int32 GetHashCode()
         {
-            return base.GetHashCode();
+            unchecked
+            {
+                var hashCode =  _fragment != null ? StringComparer.Ordinal.GetHashCode(_fragment) : 0;
+                hashCode = (hashCode * 397) ^ (_query != null ? StringComparer.Ordinal.GetHashCode(_query) : 0);
```
FlorianRappl, line 327: Wouldn't a hashcode of the computed `Href` string be as accurate, but simpler?

FlorianRappl:
Looks good to me - my only question is whether just taking the hash code of the computed string representation would be good enough.

src/AngleSharp/Url.cs
alex-fomin: Yes it will, but `GetHashCode` should be as fast as possible and Href is too 'heavy' for calculating hash code.

FlorianRappl: Can you back this statement up? Because I see many many calls in the given code and after all Href should pretty much do the same work from the null checks POV.

In the end a simpler solution should be preferred unless its really not performing well. So I would love to understand why its made so much more complicated than just  `StringComparer.Ordinal.GetHashCode(Href)`.

Note that `Href` will be called in the usual use case anyway and could be cached, too.

alex-fomin: Here is Benchmark.Net results:
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i7-9700 CPU 3.00GHz, 1 CPU, 8 logical and 8 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT


```
|            Method |     Mean |    Error |   StdDev |
|------------------ |---------:|---------:|---------:|
|   HrefGetHashCode | 92.54 ns | 0.353 ns | 0.295 ns |
| CustomGetHashCode | 39.85 ns | 0.108 ns | 0.101 ns |


```
    public class GetHashCode
    {
        private readonly Url _url;

        public GetHashCode()
        {
            _url = new Url("http://example.com/test?q=a#tag");
        }

        [Benchmark]
        public int HrefGetHashCode() => _url.Href.GetHashCode();

        [Benchmark]
        public int CustomGetHashCode() => _url.GetHashCode();
    }
```

`Href` builds a string in a correct but too complicated way for getting hash code. So `Href.GetHashCode()` is 2 times slower.

FlorianRappl: Alright not significant in this case, but convincing enough.

Thanks for the investigation! Let's ship it! :rocket:

FlorianRappl (approved):

PR #861, opened 2020-04-01
Fix MinifyMarkupFormatter removes quotes when it shouldn't

### LINKED ISSUE #860: MinifyMarkupFormatter removes attribute quotes when it should not
# Bug Report

## Description

[MinifyMarkupFormatter](https://github.com/AngleSharp/AngleSharp/blob/master/src/AngleSharp/Html/MinifyMarkupFormatter.cs#L208) removes attribute quotes when [the spec says not to remove them](http://w3c.github.io/html-reference/syntax.html#attr-value-unquoted). MinifyMarkupFormatter only checks for whitespaces but the spec indicates other characters: 

> In addition to the general requirements for attribute values, an unquoted attribute value has the following restrictions:
> 
> - must not contain any literal space characters
> - must not contain any """, "'", "=", ">", "<", or "`", characters
> - must not be the empty string

In the following example the quotes are mandatory:

````
<div a="b="></div>
````

## Steps to Reproduce

````
// Should preserve quotes or add a space after
var document = parser.ParseDocument("<div a='/'></div>");
document.ToHtml(Console.Out, new MinifyMarkupFormatter { ShouldKeepAttributeQuotes = false });
// <div a=/></div>

// Should preserve quotes
var document = parser.ParseDocument("<div a='='></div>");
document.ToHtml(Console.Out, new MinifyMarkupFormatter { ShouldKeepAttributeQuotes = false });
// <div a==></div>
````

**Expected behavior:** It should preserve quotes as indicated in the spec 😊

**Actual behavior:** It removes quotes

**Environment details:** Windows 10, .NET Core 3.1.3

## Possible Solution

https://github.com/AngleSharp/AngleSharp/blob/master/src/AngleSharp/Html/MinifyMarkupFormatter.cs#L208

````
if (ShouldKeepAttributeQuotes || result.Any(c => CharExtensions.IsWhiteSpaceCharacter(c) || c == '"' || c == '\'' || c == '=' || ...)) // Add all case
// It should also check if the value ends with a '/' and add a space after if needed
````

meziantou:
Fix #860

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/MinifyFormatter.cs
+++ b/src/AngleSharp.Core.Tests/Library/MinifyFormatter.cs
@@ -72,6 +72,36 @@ namespace AngleSharp.Core.Tests.Library
         }
 
         [Test]
+        [TestCase("=")]
+        [TestCase("`")]
+        [TestCase("<")]
+        [TestCase(">")]
+        [TestCase("'")]
+        [TestCase("&quot;")]
+        public void MinifiesPreservesQuotesWhenNeeded(string attributeValue)
+        {
+            var input = $"<div a=\"{attributeValue}\"></div>";
+            var output = Minify(input);
+            Assert.AreEqual(input, output);
+        }
+
+        [Test]
+        public void MinifiesAddSpaceAfterAttributeValueWhenItEndsWithASlash()
+        {
+            var input = "<div a='abc/'></div>";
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):

FlorianRappl:
Looks good to me!

:+1:

PR #894, opened 2020-07-17
Provide SPDX license identifier

mburtscher:
This change adds the SPDX license identifier to the generated NuGet packages by adding the appropriate [license](https://docs.microsoft.com/en-us/nuget/reference/nuspec#license) tag.

**Why?** Companies use the license information provided by package managers to ensure compliance across large projects. Providing an SPDX license identifier makes it simpler for people to work with automated tools like [LicenseFinder](https://github.com/pivotal/LicenseFinder) or GitLab's license scanning feature.# Types of Changes

### DIFF
--- a/src/AngleSharp.nuspec
+++ b/src/AngleSharp.nuspec
@@ -5,6 +5,7 @@
     <version>$version$</version>
     <authors>AngleSharp</authors>
     <owners>Florian Rappl</owners>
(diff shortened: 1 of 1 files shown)


FlorianRappl (approved):
Sounds like a useful addition to me!

Thanks for the contribution :beers:!

PR #865, opened 2020-04-10
Release 0.1.5.0

FlorianRappl:
# Types of Changes

## Prerequisites

Please make sure you can check the following two boxes:

- [x] I have read the **CONTRIBUTING** document
- [x] My code follows the code style of this project

## Contribution Type

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [x] Bug fix (non-breaking change which fixes an issue, please reference the issue id)
- [x] New feature (non-breaking change which adds functionality, make sure to open an associated issue first)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## Description

~This is the long awaited 1.0.0.~

~:warning: Not yet ready to merge. First the new documentation needs to be rolled out. Furthermore, some issues (planned for 1.0.0) should be merged in, too.~

~The PR is rather already here to indicate the desire for 1.0.0 to come ASAP and to easily follow up on the changes.~

**Update**: This is an intermediate release before releasing 1.0.0. Mostly to try out the new pipeline, but also to make a step forward and *finally* have the 1.0.0.

PR #904, opened 2020-09-25
Check for cancelation when parsing html async

The-Nutty:
# Types of Changes

## Prerequisites

## Contribution Type

- [x] All new and existing tests passed (Cookies ones where failing before i think)

## Description
Currently there is no way to cancel parsing in the HtmlDomBuilder so when parsing the document check the already existing cancelation token for if cancelation is requested. This is an issue as if you are parsing a large document in a low cpu environment it can cause the program to hault with no way to cancel. 

This just calls `cancelToken.ThrowIfCancellationRequested` in `ParseAsync` very simple, i have not added any tests as im not sure what/if any tests are worth adding.

I did a few benchmarks to confirm that this was no slower and all the results where with in margin of error of the test.

### DIFF
--- a/src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
+++ b/src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
@@ -111,6 +111,7 @@ namespace AngleSharp.Html.Parser
                     await source.PrefetchAsync(8192, cancelToken).ConfigureAwait(false);
                 }
 
+                cancelToken.ThrowIfCancellationRequested();
                 token = _tokenizer.Get();
                 Consume(token);
 


FlorianRappl (approved):
LGTM - thanks for providing this enhancement!

:beers:

The-Nutty:
No problem, any idea when this might land in a build? As its something thats causing us a few issues.

FlorianRappl:
> No problem, any idea when this might land in a build? As its something thats causing us a few issues.

Actually the preview *should* have been out now, however, due to one failing test (unfortunately, httpbin changing their APIs/behavior...) it was not published.

I am going to push a working version tonight, so on Monday you should be able to include the next v1 preview. Would that be fine, @The-Nutty ?

The-Nutty:
I think so, the only issue is we depend on a package using V0.14 (as its the latest table version), has the interface changed for V1? Do you think this will cause us any issues? Im also assuming it could be a while before V1 goes stable?

FlorianRappl:
There are no breaking changes in v1 - its just a stability indicator (for AngleSharp as a whole). The only reason for not publishing it yet is that I still want to publish the new documentation, but I have serious trouble finding time to make that beautiful...

See the changes: https://github.com/AngleSharp/AngleSharp/pull/865/files

PR #911, opened 2020-10-14
GetSelector - id with leading digits or special characters

### LINKED ISSUE #909: GetSelector - elements with number ids
# Bug Report

## Prerequisites

- [x] Are you running the latest version of AngleSharp?  (_non-alpha_, 0.14.0)
- [ ] Did you check the FAQs to see if that helps you?
- [ ] Did you perform a search in the issues? 

For more information, see the `CONTRIBUTING` guide.

## Description

`GetSelector` provides an "#id" selector starting with a number, which is invalid if you try to use it later.

(Reference: ) https://css-tricks.com/ids-cannot-start-with-a-number/

## Steps to Reproduce

See tests in branch.

**Expected behavior:** `<div id="1st">` --> `[id='1st']`

**Actual behavior:** `<div id="1st">` --> `#1st`

**Environment details:** .NET Core, Windows/Linux

## Possible Solution

Hi, I've used the `GetSelector` utility many times on a few different websites.

I recently tried a new website with some funky web standards.

This page has many elements that begin with a number, or are all numbers. (Like id = '0ab12cd3', possibly generated by some tool.)

I have a fork with a branch for a possible fix:
https://github.com/AngleSharp/AngleSharp/compare/master...robert-baldini:getselector-id-with-numbers

CSS id selectors that start with a number need to be escaped. (And sure enough, if you try `QuerySelector("#1st")`, you'll get an error. It fails for other tools also.)

The easiest way to escape it would be like: `[id='0ab12cd3']` or `[id='1st']`

I tried to keep it consistent with the existing behavior, so id will normally be passed back as "#id", unless it's neccesary to escape. The alternative would be to pass back the id with always using the square brackets.

### LINKED ISSUE #910: GetSelector - pages with duplicate element ids
# Bug Report

## Prerequisites

- [x] Are you running the latest version of AngleSharp?  (_non-alpha_, 0.14.0)
- [ ] Did you check the FAQs to see if that helps you?
- [ ] Did you perform a search in the issues?

For more information, see the `CONTRIBUTING` guide.

## Description

`GetSelector` can produce a non-unique id, which is invalid if you try to use it later.

## Steps to Reproduce

See tests in branch.

**Expected behavior:** 

```csharp
<div id="something">First</div>
<span id="something">Second</span>

var span = parent.QuerySelector("span");
var spanSelector = span.GetSelector(); // spanSelector should be something like "body>span"
var queryResult = parent.QuerySelector(spanSelector); // gets the span
```

**Actual behavior:** 

```csharp
<div id="something">First</div>
<span id="something">Second</span>

var span = parent.QuerySelector("span");
var spanSelector = span.GetSelector(); // spanSelector is "#something"
var queryResult = parent.QuerySelector(spanSelector); // gets the div!
```

**Environment details:** .NET Core, Windows/Linux

## Possible Solution

Hi, I've used the `GetSelector` utility many times on a few different websites.

I recently tried a new website with some funky web standards.

This page has many elements with the same ID. (Like id = "RULE_SET", same id, all over the place. Possibly generated by some tool.)

I have a fork with a branch for a possible fix:
https://github.com/robert-baldini/AngleSharp/commit/28782f9200179a1b7ec5c2e447aec592571956e7

This fix will only stop at the id selector if the id selector is, in fact, unique.

Possible side effect: if 'Owner' is null, `isUnique` would be false, and it would probably go back to the root in that case. But it looks like 'Owner' is usually not null.

robert-baldini:
# Types of Changes

## Prerequisites

## Contribution Type

 **( ? )** Breaking change (fix or feature that would cause existing functionality to change)

## Description

Closes #909 

See [this blog post](https://mathiasbynens.be/notes/css-escapes) for more details.

### DIFF
--- a/src/AngleSharp/Dom/ElementExtensions.cs
+++ b/src/AngleSharp/Dom/ElementExtensions.cs
@@ -1528,11 +1528,13 @@ namespace AngleSharp.Dom
                 // Always lowercase node name in the path
                 var name = element.LocalName;
 
-                // If node has id attribute...
+                // Id is unique in the DOM, so we can use it to locate the element and skip other parents
                 if (hasId)
                 {
-                    // Id is unique in the DOM, so we can use it to locate the element and skip other parents
-                    name = "#" + element.Id;
+                    var isFirstDigitNumeric = element.Id.Length > 0 && Char.IsNumber(element.Id[0]);
+                    name = isFirstDigitNumeric
+                        ? "[id='" + element.Id + "']"
+                        : "#" + element.Id;
                 }
                 // If node has siblings of the same type...
                 else if (parent != null && !element.IsOnlyOfType())
--- a/src/AngleSharp/Dom/ElementExtensions.cs
+++ b/src/AngleSharp/Dom/ElementExtensions.cs
@@ -1528,13 +1528,11 @@ namespace AngleSharp.Dom
                 // Always lowercase node name in the path
                 var name = element.LocalName;
 
-                // Id is unique in the DOM, so we can use it to locate the element and skip other parents
+                // If node has id attribute...
                 if (hasId)
                 {
-                    var isFirstDigitNumeric = element.Id.Length > 0 && Char.IsNumber(element.Id[0]);
-                    name = isFirstDigitNumeric
-                        ? "[id='" + element.Id + "']"
-                        : "#" + element.Id;
+                    // Id is unique in the DOM, so we can use it to locate the element and skip other parents
+                    name = "[id='" + element.Id + "']";
                 }
                 // If node has siblings of the same type...
                 else if (parent != null && !element.IsOnlyOfType())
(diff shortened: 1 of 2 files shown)


src/AngleSharp/Dom/ElementExtensions.cs
```diff
@@ -1532,7 +1532,7 @@ public static String GetSelector(this IElement element)
                 if (hasId)
                 {
                     // Id is unique in the DOM, so we can use it to locate the element and skip other parents
-                    name = "#" + element.Id;
+                    name = "[id='" + element.Id + "']";
```
FlorianRappl, line 1535: I think there is somewhere a test if its a valid ID in AngleSharp. Maybe we can use this.

I think the identified selector should as "nice" as possible - what's your opinion?

FlorianRappl:
Landed in `devel`.

PR #941, opened 2021-04-21
Drop .NETSTANDARD1.3, cross target NETCOREAPP3.1, and prepare to introduce span

iamcarbon:
# Types of Changes

This PR is to gather feedback and interest in dropping support for NETSTANDARD1.3 in preparation to take a dependency on System.Memory and introduce Span into the library. There's many paths that we utilize span in to reduce allocations and improve performance.

## Prerequisites

## Contribution Type

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -6,8 +6,8 @@ var frameworks = new Dictionary<String, String>
     { "net46", "net46" },
     { "net461", "net461" },
     { "net472", "net472" },
-    { "netstandard1.3", "netstandard1.3" },
     { "netstandard2.0", "netstandard2.0" },
+    { "netcoreapp3.1", "netcoreapp3.1" },
 };
 
 #load tools/anglesharp.cake
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -1,10 +1,10 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
     <IsPackable>false</IsPackable>
-    <LangVersion>7.1</LangVersion>
+    <LangVersion>9</LangVersion>
     <AssemblyName>AngleSharp.Core.Tests</AssemblyName>
   </PropertyGroup>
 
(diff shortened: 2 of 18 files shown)


FlorianRappl:
Sounds great!

Quick question: Have you done experiments with `Span` in AngleSharp yet? I'm asking because I did some experiments right after `Span` landed in .NET Core and I got disappointing results. The reason is that not many areas can be optimized. The W3C spec does a lot of normalization, which means you cannot reuse parts of the original source. Therefore, a `Span` to it will not be benificial. I then went on to apply to less sophisticated parts such as `Url`, which also do some parsing. But its more or less the same here.

So before removing support for .NET Standard 1.3 (which may hurt some users) I'd like to know if we really benefit from `Span` / can leverage this. Thanks!

ViktorHofer:
> So before removing support for .NET Standard 1.3 (which may hurt some users)

Just wanted to share my stance on this. I think the number of devs that would be hurt by dropping .NET Standard 1.3 would be quite low, as AngleSharp already supports .NET Standard 2.0 which runs on:
- .NET Framework 4.6.1 and above (4.6.1 is the minimum supported version afaik)
- .NET Core 2.0 and above (2.1 is the minimum supported version)
- UAP10.0.16299 (latest)

And most libraries are already .NET Standard 2.0 as well.

iamcarbon:
@FlorianRappl Most of the gain are certainly small, but can certainly add up -- especially in the hotter parsing paths.

Just looking at Url -- we can eliminate at least 3 additional allocations from Utf8PercentEncode, 2 allocations from TrySanatizeHost, an allocation from ParseSchema.

In StringExtensions, we can eliminate 2 allocations in ReplaceFirst.

We can eliminate an allocation in Punycode. At least 4 in BaseInputType. 1 allocation in HtmlToken.TrimStart. etc.

FlorianRappl:
Yeah that was my assumption / hope, too, but in the end times remained almost equal - sometimes even worse.

All I'm after is that we first confirm potential gains before doing many changes and then realizing that nothing is to be gained. I may be too cautious here (and sorry for that), but it would be superb to first see some enhancements (could be this branch) before committing 100% to it.

And I fully agree - every dropped allocation is a good thing. Presumably many of these functions are called thousands of times for larger documents, so it definitely adds up.

iamcarbon:
@FlorianRappl  -- We're pretty likely regress performance using Span on NET46x and I don't expect to see gains > 2-3% on the modern runtimes. Even these will introduce additional complexity and conditional code paths into the library -- which have a cost.

Agreed on your caution! 
 
I'll spend a few hours later this week seeing if I can make a more compelling argument.

Regardless of whether we choose to adopt Span, I still believe we should consider dropping NETSTANDARD1.3 before the 1.0 release.

FlorianRappl:
Fully agreed!

I think this is settled; we'll drop .NET Standard 1.3 for the 1.0.0.

ViktorHofer:
> I think this is settled; we'll drop .NET Standard 1.3 for the 1.0.0.

Happy to help with that. Just let me know.

PR #942, opened 2021-04-21
GitHub Action

campersau:
# Types of Changes

Migrate from AppVeyor to GitHub Actions

## Prerequisites

## Contribution Type

## Description

This migrates the CI builds from AppVeyor to GitHub Actions as suggested here https://github.com/AngleSharp/AngleSharp/pull/865#issuecomment-823926228

- updated build and test dependencies
- running builds on linux and windows
- publishing the nupkg as an artifact
- put logo.png inside nuget package

TODO:
- [x] check if `Publish` and `PrePublish` targets still work
- [x] conditional run `Publish` and `PrePublish` targets
- [ ] since the `RunNumber` now starts at `1` again we would need to also increase the version or come up with another solution

You can find example runs here: https://github.com/campersau/AngleSharp/actions/runs/771027266


FlorianRappl (approved):
You sir, are a hero! 🥇 

Now I only need to set up the right env variables. Let me see what's needed ...

FlorianRappl:
Can you change the `GITHUB_API_TOKEN` to `GITHUB_TOKEN` @campersau ? This one is automatically available by GitHub Actions. I've added the NuGet key / secret.

FlorianRappl (approved):
LGTM - thanks a lot!

campersau:
What do you think about the TODOs? Especially the last one? Or should we do them in a separate PR?

FlorianRappl:
Regarding the open points:

- check if Publish and PrePublish targets still work

Definitely. We'd start anyway with the prepublish due to pushing into `devel`. Then I'd suggest making another release (0.14.1?) to see if that works. Let's re-run until we have success.

- conditional run Publish and PrePublish targets

Can you specify what you mean here?

- since the RunNumber now starts at 1 again we would need to also increase the version or come up with another solution

True. It only affects the preview version though. Since we aim to have a fast 0.14.1 to test the publish pipeline I'd say it does not matter that this was reset. We should not run into conflicts.

FlorianRappl:
(I'm usually quite bad at these CI/CD things. I usually just push and reiterate until I have the pipeline I'd like to have. This has many red runs... Maybe you are better, but in either case its fine for me that we test the pipeline with merges / iterate fast on the real system.)

campersau:
The old code looked like this:
https://github.com/AngleSharp/AngleSharp/blob/1cd9dcfdb5d8ca27e32c87ef1abc14557b984e23/appveyor.yml#L12-L18

Where depending on the branch and if it was a PR or not the `Publish` and `PrePublish` targets were run instead of the default one.

---

It took me also some tries: https://github.com/campersau/AngleSharp/actions :)

FlorianRappl:
Yeah so; I think we use a similar mechanism.

- I would remove publishing the NuGet package from the GitHub workflow; this is done from Cake.
- The master branch should publish, the devel branch should pre-publish; any other branch (e.g., PR review) should just build and test
- Linux would always just build and test and would be disabled for publish or prepublish (this way PRs are properly checked against all platforms, but actual building is done exclusively on one)

campersau:
It was just uploading it as an artifact (not publishing to a feed) so one could download the nupkg and test it locally. But I have removed it for now. E.g. here under "Artifacts" https://github.com/campersau/AngleSharp/actions/runs/771027266

The conditions should work now.

FYI: This test seems to fail sometimes (it always fails on my machine) because google now shows the cookie consent stuff:
https://github.com/AngleSharp/AngleSharp/blob/3968eb050e142b1d94550fba407afe772232b126/src/AngleSharp.Core.Tests/Examples/Forms.cs#L12-L22

FlorianRappl:
Yeah I think we should remove the Google stuff (and other things like httpbin). They have been useful up to some point, but right now they are problematic from a maintenance point of view.

Great PR - I'll merge it now in and remove all ties to AppVeyor. Welcome to our new GitHub overlords!

campersau:
Cool looks like pushing to NuGet has worked: https://www.nuget.org/packages/AngleSharp/1.0.0-alpha-1

FlorianRappl:
Awesome!

I'll now try to release a 0.14.1 or so.

FlorianRappl:
(I will also switch all projects to use GitHub actions; essentially copying your change to the GitBase repo (https://github.com/AngleSharp/AngleSharp.GitBase) and pulling from there. Hope this is fine for you?)

Thanks again for the superb contribution - much appreciated 🍻 !

campersau:
Sure, didn't know that existed :)
Let me know if i can help.

FlorianRappl:
Sure - I'd make a PR and add you as a reviewer there!

PR #946, opened 2021-04-22
Reduce memory allocations on hot path

lahma:
# Types of Changes

## Prerequisites

- [ ] I have read the **CONTRIBUTING** document - didn't find it?

## Contribution Type

- [x] All new and existing tests passed - except for the SubmittingToGoogleWithCookiesShouldWork , seems to be broken

## Description

While profiling Orchard Core CMS I saw considerable amount of allocations made by AngleSharp. With these changes I'm trying to alleviate the situation:

* Only allocate `HtmlParseEvent` if someone is listening
* Add separate GetEnumerator implementation that returns concrete type, no need to box via interface (CLR knows how to bind to concrete `GetEnumerator` method)

/cc @sebastienros

### DIFF
--- a/src/AngleSharp/Dom/EventTarget.cs
+++ b/src/AngleSharp/Dom/EventTarget.cs
@@ -148,6 +148,8 @@
             return false;
         }
 
+        internal Boolean HasEventListeners => _listeners != null && _listeners.Count > 0;
+
         /// <summary>
         /// Dispatch an event to this Node.
         /// </summary>
--- a/src/AngleSharp/Dom/Internal/NodeList.cs
+++ b/src/AngleSharp/Dom/Internal/NodeList.cs
@@ -78,7 +78,9 @@
 
         #region IEnumerable Implementation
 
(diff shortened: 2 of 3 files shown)


FlorianRappl (approved):
Great change!

While this (and I assume the other(s) / if there are more to come) PR is great by itself I would appreciate at least one perf. test to show the before / after. Once with a smaller document (shouldn't have negative side effects) and once with a larger (should be noticeable there).

Again, great work! Thanks 🍻

lahma:
I understand, I'm not familiar enough with the code base, but the request is well warranted. I usually use BenchmarkDotNet for comparisons.

 I'll try to improve if/when there's other places to improve. Thanks for the quick response!

lahma:
Just adding a note that these two PRs targeted the "greatest offenders" I was while running profiling for a simple case. Here's OC's session filtered for AngleSharp allocations and the four top most should be now covered with better patterns.

image

FlorianRappl:
Awesome thanks for the info! You rock @lahma ! :beers:

PR #947, opened 2021-04-22
Text source optimization

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

- [x] All new and existing tests passed - except for the Google one

## Description

TextSource always allocates `_buffer` and `_chars` even though they are only needed when `baseStream` is present. With use case like Orchard the string-based constructor is only called so there's no need to do the extra allocations.

/cc @sebastienros


FlorianRappl (approved):
This is great. Thanks!

PR #949, opened 2021-04-23
Make Configuration.Default reuse same configuration instance

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

The Configuration.Default getter always allocates a new instance of configuration. As configuration is meant to be immutable (which is very cool) there should be no need to create new one each time. Getter-like service can also be surprising to end users when it allocates every time.

### DIFF
--- a/src/AngleSharp/Configuration.cs
+++ b/src/AngleSharp/Configuration.cs
@@ -58,7 +58,7 @@ namespace AngleSharp
                 Creator<ICssSelectorParser>(ctx => new CssSelectorParser(ctx)),
                 Creator<IHtmlParser>(ctx => new HtmlParser(ctx)),
                 Creator<INavigationHandler>(ctx => new DefaultNavigationHandler(ctx)),
-            }; ;
+            };
         }
 
         #endregion
@@ -69,7 +69,7 @@ namespace AngleSharp
         /// Gets the default configuration to use. The default configuration
         /// can be overriden by calling the SetDefault method.
         /// </summary>
-        public static IConfiguration Default => new Configuration();
+        public static IConfiguration Default { get; } = new Configuration();
 
         #endregion
 


lahma:
There seems to be something wrong with the linux build 🤷🏻‍♂️

FlorianRappl:
> There seems to be something wrong with the linux build 🤷🏻‍♂️

Common NuGet issue... I think re-queuing fixes it. Let's see...

FlorianRappl (approved):
LGTM!

campersau:
The tests are running in parallel and the services of the configuration are now the same which means accessing a service like `IDocumentFactory` returns the same instance which is not threadsafe and thus throws as you can see in the linux build.

```
Error: LoadCustomDocumentWithoutUnregisteredHandler: System.ArgumentException : An item with the same key has already been added. Key: text/markdown
  Failed LoadCustomDocumentWithoutUnregisteredHandler [138 ms]
  Error Message:
   System.ArgumentException : An item with the same key has already been added. Key: text/markdown
  Stack Trace:
     at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at AngleSharp.Dom.DefaultDocumentFactory.Register(String contentType, Creator creator) in /home/runner/work/AngleSharp/AngleSharp/src/AngleSharp/Dom/DefaultDocumentFactory.cs:line 42
   at AngleSharp.Core.Tests.Library.ContextLoadingTests.LoadCustomDocumentWithoutUnregisteredHandler() in /home/runner/work/AngleSharp/AngleSharp/src/AngleSharp.Core.Tests/Library/ContextLoading.cs:line 403
```

The configuration itself is immutable but not the services it holds.

lahma:
@campersau thanks for the insights. This is a bit problematic as the configuration building can be quite resource intensive (always generates new instances, uses distinct etc). I'm wondering if there's some way to restructure the logic or making things more thread safe.

lahma:
It seems that this cannot be easily done as public API allows service mutation for multiple services and they cannot be considered singletons. The caching is something that has to be done on library consuming side where it's known that setup won't be altered after configuration initialization. Ideally AngleSharp might at some point ensure that services can be singletons.

Thanks for helping me understand a bit more about the mechanics under the hood 👍🏻

FlorianRappl:
Yeah its a good point. I also thought its not immutable, but on the other hand all APIs (`With`, `Without`) use the `Configuration` just to create a new `Configuration`.

I guess we can / should make `Configuration` more immutable. After all, its just a carrier for the services - bringing them to a browsing context.

Thanks for bringing this up @campersau and thanks for the effort @lahma !

PR #951, opened 2021-04-23
Replace ForEachRange with one-time loops

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

`ForEachRange` was a bit of an anti-pattern as it led into doing loops against collections multiple times just to do separate things based on predicate. Switched to single loop that checks condition and executes the logic. 

I'll happily add performance results if you can give a description how you want them done.


lahma:
`RemoveChild` loves to allocate (because of the context-capturing closures, that's why imprerative local code is better)
image

FlorianRappl (approved):
LGTM! 🍻

PR #952, opened 2021-04-24
Enable nullable

iamcarbon:
# Types of Changes

This PR enables nullable annotations. No new APIs or functionality were introduced.

## Prerequisites

## Contribution Type


src/AngleSharp/BrowsingContext.cs
```diff
@@ -182,11 +182,11 @@ public IDocument Active
         /// <param name="name">The name of the child context, if any.</param>
         /// <param name="security">The security flags to apply.</param>
         /// <returns></returns>
-        public IBrowsingContext CreateChild(String name, Sandboxes security)
+        public IBrowsingContext CreateChild(String? name, Sandboxes security)
         {
             var context = new BrowsingContext(this, security);
 
-            if (!String.IsNullOrEmpty(name))
+            if (name is { Length: > 0 })
```
iamcarbon, line 189: Annoyingly, String.IsNullOrEmpty doesn't update the nullable status.

iamcarbon:
Ready for review / feedback.

FlorianRappl (commented):
This looks great! :beers:

I love it - only thing is right now the failing test. Not sure if that has to do with the PR or with the target. I'll look into that later.

iamcarbon:
Quick update. This single test failure is caused by a connection issue to http://www.empireaerials.net. 

I attempted to build a custom handler to decompress a local gzip resource -- but realized we are testing the AutomaticDecompression behavior of the DefaultHttpRequester. Sadly, this utilizes the legacy WebRequest and WebResponse API -- which doesn't appear to have an easy method of mocking.

My recommendation is that we disable this test for now and consider swapping out the HttpWebRequest / HttpWebResponse  backed API with HttpClient in a new PR. With an HttpClient backed implementation, we can fully emulate the http pipeline within the testing framework using a custom HttpMessageHandler.

FlorianRappl:
> My recommendation is that we disable this test for now and consider swapping out the HttpWebRequest / HttpWebResponse backed API with HttpClient in a new PR. With an HttpClient backed implementation, we can fully emulate the http pipeline within the testing framework using a custom HttpMessageHandler.

I think this would be rather something for the `HttpRequester` from `AngleSharp.Io`. For me its fine to disable it though.

FlorianRappl (approved):

PR #957, opened 2021-04-28
Add ConfigureAwaitChecker.Analyzer and fix issues

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Noticed that some `ConfigureAwait` calls were missing as it seems that they are set elsewhere. Added analyzer to report problems and made warnings as errors that they block compilation.

### DIFF
--- a/src/AngleSharp/Dom/Internal/Document.cs
+++ b/src/AngleSharp/Dom/Internal/Document.cs
@@ -1486,13 +1486,13 @@ namespace AngleSharp.Dom
             {
                 var url = new Url(e.CurrentLocation);
                 var request = DocumentRequest.Get(url, source: this, referer: DocumentUri);
-                await _context.OpenAsync(request, CancellationToken.None);
+                await _context.OpenAsync(request, CancellationToken.None).ConfigureAwait(false);
             }
             else
             {
                 var url = _location.Original;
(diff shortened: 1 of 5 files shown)


src/AngleSharp/Dom/Internal/Document.cs
```diff
@@ -1486,13 +1486,13 @@ private async void LocationChanged(Object sender, Location.ChangedEventArgs e)
             {
                 var url = new Url(e.CurrentLocation);
                 var request = DocumentRequest.Get(url, source: this, referer: DocumentUri);
-                await _context.OpenAsync(request, CancellationToken.None);
```
lahma, line 1489: context here also looks a bit frightening `private async void LocationChanged`, [async voids are bad and can cause all sorts of trouble](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void).

FlorianRappl: I know, but unfortunately (as it was the reason for allowing `void` in the first place) its the signature of this according to official spec.

FlorianRappl (approved):
LGTM - thanks for finding these and adding the analyzer!

PR #963, opened 2021-05-01
Reduce allocations + add new mime types

iamcarbon:
This PR:

- Reduces allocations in the string extension methods by using pooled buffers (via System.Buffers)
- Adds additional commonly used mime types and basic test coverage
- Replace IsOneOf(chars...) method calls with basic is + or pattern
- Makes Boundary readonly
- Eliminate various cases where we can leak a pooled StringBuilder instance
- Introduce a non-allocating method to return unused StringBuilder instances
- Fix a few cases where we are comparing strings using the current culture

Possible breaking changes:
- Removed public SplitWithoutTrimming(char[], ...) overload
- Removed public IsOneOf(char...) methods

## Prerequisites

## Contribution Type

### DIFF
--- a/src/AngleSharp/AngleSharp.Core.csproj
+++ b/src/AngleSharp/AngleSharp.Core.csproj
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
+    <PackageReference Include="System.Buffers" Version="4.5.1" />
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
--- a/src/AngleSharp/Text/CharExtensions.cs
+++ b/src/AngleSharp/Text/CharExtensions.cs
@@ -280,47 +280,5 @@ namespace AngleSharp.Text
         {
             return c == 0 || c > Symbols.MaximumCodepoint || (c > 0xD800 && c < 0xDFFF);
         }
-
-        /// <summary>
-        /// Determines if the given character is one of the two others.
-        /// </summary>
-        /// <param name="c">The character to test.</param>
-        /// <param name="a">The first option.</param>
-        /// <param name="b">The second option.</param>
-        /// <returns>The result of the test.</returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Boolean IsOneOf(this Char c, Char a, Char b)
-        {
-            return a == c || b == c;
-        }
-
-        /// <summary>
-        /// Determines if the given character is one of the three others.
-        /// </summary>
-        /// <param name="c">The character to test.</param>
-        /// <param name="o1">The first option.</param>
-        /// <param name="o2">The second option.</param>
-        /// <param name="o3">The third option.</param>
-        /// <returns>The result of the test.</returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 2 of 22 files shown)


iamcarbon:
Ready for review.

src/AngleSharp/Text/CharExtensions.cs
```diff
@@ -280,47 +280,5 @@ public static Boolean IsInvalid(this Int32 c)
...
-
-        /// <summary>
-        /// Determines if the given character is one of the two others.
-        /// </summary>
-        /// <param name="c">The character to test.</param>
-        /// <param name="a">The first option.</param>
-        /// <param name="b">The second option.</param>
-        /// <returns>The result of the test.</returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Boolean IsOneOf(this Char c, Char a, Char b)
```
FlorianRappl, line 292: Not sure if we are not using this in some other lib. Would need to check.

Would it hurt to leave these utilities in there?

src/AngleSharp/AngleSharp.Core.csproj
```diff
@@ -25,6 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
+    <PackageReference Include="System.Buffers" Version="4.5.1" />
```
FlorianRappl, line 28: Should this be in the Nuspec, too?

https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp.nuspec

src/AngleSharp/Xhtml/XhtmlMarkupFormatter.cs
```diff
@@ -170,7 +170,7 @@ public static String EscapeText(String content)
         /// </summary>
         /// <param name="name">The name to be properly represented.</param>
         /// <returns>The string representation.</returns>
-        public static String XmlNamespaceLocalName(String name) => name != NamespaceNames.XmlNsPrefix ? String.Concat(NamespaceNames.XmlNsPrefix, ":") : name;
+        public static String XmlNamespaceLocalName(String name) => !string.Equals(name, NamespaceNames.XmlNsPrefix, StringComparison.Ordinal) ? String.Concat(NamespaceNames.XmlNsPrefix, ":") : name;
```
FlorianRappl: We could use `Is` here (https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp/Text/StringExtensions.cs#L341).

FlorianRappl (commented):
Great changes again - thanks!

src/AngleSharp/Xhtml/XhtmlMarkupFormatter.cs
iamcarbon: Done.

src/AngleSharp/AngleSharp.Core.csproj
iamcarbon: Good catch. Added.

src/AngleSharp/Text/CharExtensions.cs
iamcarbon: Re-added with an obsolete notice.

iamcarbon:
@FlorianRappl Updated.

FlorianRappl (approved):
Awesome! Thanks :beers:

PR #965, opened 2021-05-04
Improve Code Quality 3

iamcarbon:
This PR:

- Compiles regular expressions
- Adds a timeout to regular expressions that may be from an untrusted source
- Replaces a few method calls with constant pattern matching
- Marks additional fields as readonly by removing private setters

## Prerequisites

## Contribution Type

### DIFF
--- a/src/AngleSharp/Html/InputTypes/ColorInputType.cs
+++ b/src/AngleSharp/Html/InputTypes/ColorInputType.cs
@@ -1,4 +1,4 @@
-﻿namespace AngleSharp.Html.InputTypes
+namespace AngleSharp.Html.InputTypes
 {
     using AngleSharp.Html.Dom;
     using System;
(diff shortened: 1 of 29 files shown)


src/AngleSharp/Html/InputTypes/ColorInputType.cs
```diff
@@ -8,7 +8,7 @@ class ColorInputType : BaseInputType
     {
         #region Fields
 
-        static readonly Regex color = new Regex("^\\#[0-9A-Fa-f]{6}$");
+        static readonly Regex hexColorPattern = new Regex("^\\#[0-9A-Fa-f]{6}$", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
FlorianRappl: Either ignore case and only have 0-9A-F or don't ignore case and keep the pattern. But having both is a bit inefficient.

FlorianRappl (commented):
Another great PR. :beers:

src/AngleSharp/Html/InputTypes/ColorInputType.cs
iamcarbon: Good catch. Updated.

iamcarbon:
Ready for review.

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #973, opened 2021-06-09
Fix attribute owner

denis-ivanov:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Pass attribute owner to the constructor+ tests.

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/DOM.cs
+++ b/src/AngleSharp.Core.Tests/Html/DOM.cs
@@ -515,7 +515,7 @@ namespace AngleSharp.Core.Tests.Html
             Assert.IsNotNull(docType);
             Assert.AreEqual(NodeType.DocumentType, docType.NodeType);
             Assert.AreEqual(@"html", docType.Name);
-            
+
             var html = doc.DocumentElement;
             Assert.AreEqual(2, html.ChildNodes.Length);
             Assert.AreEqual(2, html.Attributes.Count());
@@ -734,6 +734,16 @@ namespace AngleSharp.Core.Tests.Html
             var img = p.Images.Select(m => m).ToList();
             var html = p.ToHtml();
             Assert.AreEqual(1, img.Count);
+        }
+
+        [Test]
+        public void OwnerElement_ShouldNotBeNull()
+        {
+            var document = "<body><div class='one'></div></body>".ToHtmlDocument();
+            var div = document.QuerySelector("div");
+            var attr = div.Attributes["class"];
+
+            Assert.NotNull(attr.OwnerElement);
         }
     }
 }
--- a/src/AngleSharp.Core.Tests/Html/DOM.cs
+++ b/src/AngleSharp.Core.Tests/Html/DOM.cs
@@ -744,6 +744,7 @@ namespace AngleSharp.Core.Tests.Html
             var attr = div.Attributes["class"];
 
             Assert.NotNull(attr.OwnerElement);
+            Assert.AreSame(div, attr.OwnerElement);
         }
     }
 }
--- a/src/AngleSharp/Dom/Internal/Attr.cs
+++ b/src/AngleSharp/Dom/Internal/Attr.cs
@@ -15,7 +15,6 @@ namespace AngleSharp.Dom
         private readonly String _localName;
         private readonly String? _prefix;
         private readonly String? _namespace;
-        private readonly IElement? _ownerElement;
         private String _value;
 
         #endregion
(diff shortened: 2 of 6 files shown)


src/AngleSharp.Core.Tests/Html/DOM.cs
```diff
@@ -735,5 +735,15 @@ public void IteratingThroughDomShouldNeverFail_Issue763()
...
         }
+
+        [Test]
+        public void OwnerElement_ShouldNotBeNull()
+        {
+            var document = "<body><div class='one'></div></body>".ToHtmlDocument();
+            var div = document.QuerySelector("div");
+            var attr = div.Attributes["class"];
+
+            Assert.NotNull(attr.OwnerElement);
```
campersau, line 746: Would an additional `Assert.AreSame`
```suggestion
            Assert.NotNull(attr.OwnerElement);
            Assert.AreSame(div, attr.OwnerElement);
```
 make sense here?

src/AngleSharp/Dom/Internal/Attr.cs
```diff
@@ -15,6 +15,7 @@ public sealed class Attr : IAttr
         private readonly String _localName;
         private readonly String? _prefix;
         private readonly String? _namespace;
+        private readonly IElement? _ownerElement;
```
FlorianRappl: Well, this adds 1 field per attribute. As there are most likely more attributes than elements the memory consumption should not be disregarded here.

We already can infer the owner element via the container. Is there a reason to duplicate this information?

FlorianRappl (commented):
Well, in general is fine, but we should do something about the owner element field duplication. I'd rather like to see this only once in there. Maybe instead of passing the owner via the constructor we should pass the container? Or remove the owner from the container (but the container may need this info, too)?

src/AngleSharp/Html/Parser/HtmlDomBuilderExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static void SetAttributes(this Element element, List<HtmlAttributeToken>
             for (var i = 0; i < attributes.Count; i++)
             {
                 var attribute = attributes[i];
-                var item = new Attr(attribute.Name, attribute.Value);
+                var item = new Attr(attribute.Name, attribute.Value, element);
```
campersau:
```suggestion
                var item = new Attr(attribute.Name, attribute.Value);
                item.Container = container;
```

src/AngleSharp/Dom/Internal/Element.cs
```diff
@@ -630,7 +630,7 @@ protected void CloneElement(Element element, Document owner, Boolean deep)
 
             foreach (var attribute in _attributes)
             {
-                var attr = new Attr(attribute.Prefix, attribute.LocalName, attribute.Value, attribute.NamespaceUri);
+                var attr = new Attr(attribute.Prefix, attribute.LocalName, attribute.Value, attribute.NamespaceUri, this);
```
campersau:
```suggestion
                var attr = new Attr(attribute.Prefix, attribute.LocalName, attribute.Value, attribute.NamespaceUri);
                attr.Container = element._attributes;
```

campersau (commented):
Would it work if the container is set manually everywhere `FastAddItem` is used?

E.g. in `Element.cs`:
```c#
        public void AddAttribute(Attr attr)
        {
            attr.Container = _attributes;
            _attributes.FastAddItem(attr);
        }
```

FlorianRappl:
> Would it work if the container is set manually everywhere FastAddItem is used?

I think it would.

src/AngleSharp.Core.Tests/Html/DOM.cs
denis-ivanov: @campersau, done.

src/AngleSharp/Dom/Internal/Attr.cs
denis-ivanov: @FlorianRappl, it was a bad idea. I've removed this field.

src/AngleSharp/Html/Parser/HtmlDomBuilderExtensions.cs
denis-ivanov: Fixed

src/AngleSharp/Dom/Internal/Element.cs
denis-ivanov: Fixed

src/AngleSharp/Dom/Internal/Attr.cs
FlorianRappl: No worries - thanks for the awesome PR @denis-ivanov !

FlorianRappl (approved):
LGTM! Thanks :beers:

Issue #977, opened 2021-06-16
Add an interface for HtmlPictureElement


Turnerj:
# New Feature Proposal

## Description

Many of the other elements have an interface that corresponds with the internal implementation of their element but I noticed that picture doesn't.

## Background

I'm wanting to use it like `myElement is IHtmlPictureElement`.

## Specification

N/A

FlorianRappl:
Yep is fine for me! Thanks for bringing this up.

PR #980, opened 2021-06-17
Added additional element interfaces

Turnerj:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

I didn't realise when I did #977/#978 there were a number of other elements that didn't have corresponding interfaces. This PR doesn't currently fix all of them but adds a few more interfaces for common elements.

If you'd like, I can go through all the concrete types and add the same missing interfaces to them - just let me know.

(I didn't raise a new issue for this as I felt this kinda continued from the other issue and otherwise felt a little repetitive - if you'd rather I still do it, happy to do so)

### DIFF
--- a/src/AngleSharp/Html/Dom/IHtmlBoldElement.cs
+++ b/src/AngleSharp/Html/Dom/IHtmlBoldElement.cs
@@ -1,0 +1,13 @@
+namespace AngleSharp.Html.Dom
+{
+    using AngleSharp.Attributes;
+    using System;
+
+    /// <summary>
+    /// Represents the HTML bold element.
+    /// </summary>
+    [DomName("HTMLBoldElement")]
+    public interface IHtmlBoldElement : IHtmlElement
+    {
+    }
+}
--- a/src/AngleSharp/Html/Dom/IHtmlCodeElement.cs
+++ b/src/AngleSharp/Html/Dom/IHtmlCodeElement.cs
@@ -1,0 +1,13 @@
+namespace AngleSharp.Html.Dom
+{
+    using AngleSharp.Attributes;
+    using System;
+
+    /// <summary>
+    /// Represents the HTML code element.
+    /// </summary>
+    [DomName("HTMLCodeElement")]
+    public interface IHtmlCodeElement : IHtmlElement
+    {
+    }
+}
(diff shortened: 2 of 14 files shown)


src/AngleSharp/Html/Dom/IHtmlBoldElement.cs
```diff
@@ -0,0 +1,13 @@
+namespace AngleSharp.Html.Dom
+{
+    using AngleSharp.Attributes;
+    using System;
+
+    /// <summary>
+    /// Represents the HTML bold element.
+    /// </summary>
+    [DomName("HTMLBoldElement")]
```
FlorianRappl, line 9: There is no such interface according to W3C.

FlorianRappl (commented):
None of these interfaces exist; I am not sure if we need the empty .NET interfaces, too (I know why you introduced them, but each interface also brings maintenance and since they are empty I don't see much value).

If you'd persist we can leave the interfaces, but the `DomName` attributes need to be dropped - otherwise we are not aligned with W3C.

src/AngleSharp/Html/Dom/IHtmlBoldElement.cs
FlorianRappl: > I didn't realise when I did #977/#978 there were a number of other elements that didn't have corresponding interfaces. 

Difference is / was: The interface you mentioned there has indeed been specified by the W3C since we introduced it. All the ones given here are not specified and will potentially never be specified.

Turnerj:
My main reason for the interfaces is so I can go `myElement is IHtmlWhateverTheElementIs myWhateverElement` and access whatever properties that may (or may not) be mapped. Because the implementations are all internal, I'd need corresponding interfaces to do this instead.

I've used `element.TagName == "A"` etc but it felt weird doing that for some elements but not for others. Plus if I want to access the helper properties that match to attributes, I'd still need to cast.

But you've hit the nail on the head with your other comment, explaining the purpose of the interfaces. See, I thought they were more-or-less arbitrary but you mentioned the point was to match the W3C interfaces. With that in mind, I don't think these should be added then.

What I'll probably do in my code is use the tag name comparison and avoid any interfaces, instead use the `HasAttribute`/`GetAttribute` methods directly to get what I want. That way my code can at least stay consistent within itself.

PR #981, opened 2021-06-17
[WIP] Introduce ValueStringBuilder to improve performance under load

iamcarbon:
Posting for early feedback.

This PR reduces the chance of the StringBuilderPool waiting on a lock or allocating when the project is used in a high-concurrency environment by utilizing a non-allocating ValueStringBuilder (via the .NET runtime).

Seeing a ~4% throughput improvement parsing 32 concurrent documents on an 8 core processor this change. I would also expect we can replace additional uses of StringBuilderPool.Obtain() to improve this further.

This does introduce a dependency on System.Memory, which does complicate things.

## Prerequisites

## Contribution Type

### DIFF
--- a/src/AngleSharp/AngleSharp.Core.csproj
+++ b/src/AngleSharp/AngleSharp.Core.csproj
@@ -25,7 +25,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
-    <PackageReference Include="System.Buffers" Version="4.5.1" />
+    <PackageReference Include="System.Memory" Version="4.5.4" />
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
--- a/src/AngleSharp/Css/CssUtilities.cs
+++ b/src/AngleSharp/Css/CssUtilities.cs
@@ -1,5 +1,5 @@
-using AngleSharp.Text;
 using System;
+using System.Text;
 
 namespace AngleSharp.Css
 {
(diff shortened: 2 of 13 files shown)


FlorianRappl:
I love the idea and I think the execution is also great.

My questions now are:

- Does that still work with all supported platforms (build seems to be fine etc.)?
- Are there any performance downsides (e.g., in small-workload scenarios)?
- Since the value string builder is custom (vs the out of the box stringbuilder) what's the anticipated cost of maintenance? Do you think that the standard SB is capable of catching up one day?

All in all it seems to me that there are very little downsides (if any) here. The complexity is higher, but if we are able to save perf. on *any* workload its a definite win.

iamcarbon:
/ Does that still work with all supported platforms (build seems to be fine etc.)?

- Yep, works on all our platform targets.

/ Are there any performance downsides (e.g., in small-workload scenarios)?

- I couldn't spot any regressions in single threaded scenarios, and as long as we're providing reasonable estimates on the default buffer size, should see some notable improvements. We're also able to provide a stack allocated buffer in cases where we're confident in our estimate to improve things further. 

/ Since the value string builder is custom (vs the out of the box stringbuilder) what's the anticipated cost of maintenance? Do you think that the standard SB is capable of catching up one day?

- We cloned directly from the .NET source library and can easily sync up with the changes. ValueStringBuilder is a ref struct over a 3 fields (providing direct access to a continuous buffer) -- I think this is as simple as it gets. StringBuilder has 5 fields and utilizes chunks to grow -- which results in some indirections in data access. Its design likely limits it's ability to be improved much in the future.

FlorianRappl:
Should we merge this or are you planning further enhancements here @iamcarbon ? Thanks! :beers:

iamcarbon:
@FlorianRappl I'm hoping to spend a bit more time benchmarking this and tuning the default buffer sizes before we commit to a merge.

iamcarbon:
Going to revisit thing again with .NET6.0

FlorianRappl:
Thanks for the update!

PR #1002, opened 2021-10-07
Allow DocumentUri to be changed

meesg:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

After loading a document from static HTML with `HtmlParser` there doesn't seem to be an easy way to change the documentUri to location where the html was retrieved from. This becomes a problem when we try to create a form request through `HtmlFormElement.GetSubmission()` , because the documentUri is 'about:blank'. Dependent on if the action property is present in the form, either the DocumentUri gets used directly our indirectly through the BaseUrl of the `HtmlFormElement` in `HtmlFormElement.SubmitForm(...)`. This issue can be resolved by a little bit of a dirty hack, namely always making sure the action property is always an absolute link, however I feel like this isn't a proper solution. Thoughts?

### DIFF
--- a/src/AngleSharp/Dom/IDocument.cs
+++ b/src/AngleSharp/Dom/IDocument.cs
@@ -48,7 +48,7 @@ namespace AngleSharp.Dom
         /// Gets the URI of the current document.
         /// </summary>
         [DomName("documentURI")]
-        String DocumentUri { get; }
+        String DocumentUri { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


FlorianRappl:
It would have been good to either open an issue first or ask on the gitter chat.

You should use the `BrowsingContext` if you want to have the address set.

meesg:
I will close the pull request and discuss it further in the gitter chat

PR #1006, opened 2021-10-21
Add IHtmlEncoder #1005

### LINKED ISSUE #1005: Can I customize HtmlEncode?
[This is a feature request]

I want to customize [HtmlEncode()](https://github.com/AngleSharp/AngleSharp/blob/0bee862e8aa61fe3ace84068113d2e97a77beb32/src/AngleSharp/Text/StringExtensions.cs#L628).

In my case, the encoding of the HTML document is `shift_jis` and I am trying to send the character `~` (U+301C) which cannot be converted to `shift_jis`.
Currently, the string sent is `?` in the current situation. I would like to replace `~`(U+301C) with `&#12316;` in HtmlEncode().

Original string:
```
〜
```

ShiftJIS encoded string(current):
```
?
```

ShiftJIS encoded string(using HTML character entity reference):
```
&#12316;
```

If we can inject `IHtmlEncoderService`, it would be very helpful.

tsu1980:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

This PR makes HtmlEncode customizable.

I'm not sure if this is an appropriate name for the interface. If you have a better name, please let me know.

Fixes #1005

### DIFF
--- a/src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
+++ b/src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
@@ -9,10 +9,10 @@ namespace AngleSharp.Html.Forms
 
     static class FormDataSetExtensions
     {
-        public static Stream CreateBody(this FormDataSet formDataSet, String enctype, String? charset, IHtmlEncoder htmlEncoder)
+        public static Stream CreateBody(this FormDataSet formDataSet, String enctype, String? charset, IHtmlEncoder? htmlEncoder)
         {
             var encoding = TextEncoding.Resolve(charset);
-            return formDataSet.CreateBody(enctype, encoding, htmlEncoder);
+            return formDataSet.CreateBody(enctype, encoding, htmlEncoder ?? new DefaultHtmlEncoder());
         }
 
         public static Stream CreateBody(this FormDataSet formDataSet, String enctype, Encoding encoding, IHtmlEncoder htmlEncoder)
--- a/src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
+++ b/src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
@@ -1,5 +1,6 @@
 namespace AngleSharp.Html.Forms
 {
+    using AngleSharp.Html.Forms.Submitters;
     using AngleSharp.Io;
     using AngleSharp.Text;
     using System;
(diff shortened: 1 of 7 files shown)


src/AngleSharp/Html/Forms/Submitters/IHtmlEncoder.cs
```diff
@@ -0,0 +1,20 @@
+namespace AngleSharp.Html.Forms.Submitters
+{
+    using System;
+    using System.Text;
+
+    /// <summary>
+    /// Represents the HTML encoder.
+    /// </summary>
+    public interface IHtmlEncoder
```
FlorianRappl, line 9: Should this also have a `Decode` methode? For symmetric purposes (i.e., to cover not only the en- but also decoding with the same basic implementation)?

src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
```diff
@@ -8,21 +9,21 @@ namespace AngleSharp.Html.Forms
 
     static class FormDataSetExtensions
     {
-        public static Stream CreateBody(this FormDataSet formDataSet, String enctype, String? charset)
+        public static Stream CreateBody(this FormDataSet formDataSet, String enctype, String? charset, IHtmlEncoder htmlEncoder)
```
FlorianRappl: Maybe these extensions here use default values for the last argument?

i.e.:

`IHtmlEncoder htmlEncoder = new DefaultHtmlEncoder()`

FlorianRappl (commented):
I think this looks great. Maybe we can have both combinations?

Also I am a bit surprised that I did not find it in the tokenizer / parser. Maybe I forgot how it works, but I thought it's also used there.

src/AngleSharp/Html/Forms/Submitters/IHtmlEncoder.cs
tsu1980: Hmm, but I think there is no place that HTML decoding in AngleSharp. If so, do you still need the Decode method? Do you think it will be necessary in the future?

src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
tsu1980: Does this mean that the following code is better?

```
public static Stream CreateBody(this FormDataSet formDataSet, String enctype, String? charset, IHtmlEncoder? htmlEncoder)
{
    var encoding = TextEncoding.Resolve(charset);
    return formDataSet.CreateBody(enctype, encoding, htmlEncoder ?? new DefaultHtmlEncoder());
}
```

tsu1980:
Yeah I also thought IHtmlEncoder's namespace is a bit long. But the only one that used HtmlEncode was `MultipartFormDataSetVisitor`.

src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
FlorianRappl: I'd say yes it would be preferred. So if one uses this method and does not care about the encoder then its the same behavior as right now.

src/AngleSharp/Html/Forms/Submitters/IHtmlEncoder.cs
FlorianRappl: Ah I thought we have such a method already. If not then forget my comment - just thought it would be nice for symmetric purposes (but if we don't have one then let's not implement it for no reason).

src/AngleSharp/Html/Forms/FormDataSetExtensions.cs
tsu1980: Done.

tsu1980:
Hmm, Why WindowIntervalCanBeCancelled failed?

FlorianRappl (approved):

PR #1011, opened 2021-11-21
Add .NET 6.0 target framework

silkfire:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

I believe that as a modern library, AngleSharp should target the current and latest LTS release of .NET.

This PR adds **.NET 6.0** as as a target framework in addition to the currently existing ones.
Some adjustments have been made in order to fix any resulting nullable reference errors.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
+++ b/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
@@ -431,7 +431,7 @@ namespace AngleSharp.Core.Tests.Library
             var mcp = new MemoryCookieProvider();
             var config = Configuration.Default.With(mcp).With(requester).WithDefaultLoader();
             var context = BrowsingContext.New(config);
-            var receivedCookieHeader = "THECOOKIE=value1";
+            var receivedCookieHeader = "THECOOKIE=value1; Path=/path1";
             var url = new Url("http://example.com/path1");
             //request 1: /path1, set a cookie THECOOKIE=value1
             requester.BuildResponse(req => VirtualResponse.Create(r => r
@@ -441,11 +441,13 @@ namespace AngleSharp.Core.Tests.Library
             context.OpenAsync("http://example.com/path1");
             //request 2: /path1/somefile.jsp redirects to /path2/file2.jsp
             requester.BuildResponses(new Func<Request, IResponse>[] {
-                req => VirtualResponse.Create(r => r
+                req => {
+                    return VirtualResponse.Create(r => r
                     .Address("http://example.com/path1/somefile.jsp")
                     .Content("")
                     .Status(System.Net.HttpStatusCode.Redirect)
-                    .Header(HeaderNames.Location, "http://example.com/path2/file2.jsp")),
+                    .Header(HeaderNames.Location, "http://example.com/path2/file2.jsp"));
+                },
                 req => {
                     receivedCookieHeader = req.Headers.GetOrDefault(HeaderNames.Cookie, String.Empty);
                     return VirtualResponse.Create(r => r
--- a/src/AngleSharp/Annotations/NullableAnnotations.cs
+++ b/src/AngleSharp/Annotations/NullableAnnotations.cs
@@ -1,6 +1,5 @@
 #if !NET5_0_OR_GREATER
 
-// ReSharper disable once CheckNamespace
 namespace System.Diagnostics.CodeAnalysis
 {
     internal sealed class AllowNullAttribute : Attribute { }
--- a/src/AngleSharp/Annotations/NullableAnnotations.cs
+++ b/src/AngleSharp/Annotations/NullableAnnotations.cs
@@ -1,3 +1,6 @@
+#if !NET5_0_OR_GREATER
+
+// ReSharper disable once CheckNamespace
 namespace System.Diagnostics.CodeAnalysis
 {
     internal sealed class AllowNullAttribute : Attribute { }
@@ -44,3 +47,5 @@ namespace System.Diagnostics.CodeAnalysis
         public string ParameterName { get; }
     }
 }
+
+#endif
--- a/src/AngleSharp/Common/ObjectExtensions.cs
+++ b/src/AngleSharp/Common/ObjectExtensions.cs
@@ -27,7 +27,7 @@ namespace AngleSharp.Common
                 foreach (var property in properties)
                 {
                     var value = property.GetValue(values, null) ?? String.Empty;
-                    symbols.Add(property.Name, value.ToString());
+                    symbols.Add(property.Name, value.ToString() ?? String.Empty);
                 }
             }
 
@@ -121,7 +121,7 @@ namespace AngleSharp.Common
         /// <param name="values">The dictionary for the lookup.</param>
         /// <param name="key">The key to look for.</param>
         /// <returns>An object instance or null.</returns>
-        public static Object TryGet(this IDictionary<String, Object> values, String key)
+        public static Object? TryGet(this IDictionary<String, Object> values, String key)
         {
             values.TryGetValue(key, out var value);
             return value;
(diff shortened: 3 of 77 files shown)


src/AngleSharp.Core.Tests/Library/CookieHandling.cs
```diff
@@ -431,7 +431,7 @@ public void ShouldChangeSelectedCookiesOnRedirect_Issue548()
             var mcp = new MemoryCookieProvider();
             var config = Configuration.Default.With(mcp).With(requester).WithDefaultLoader();
             var context = BrowsingContext.New(config);
-            var receivedCookieHeader = "THECOOKIE=value1";
+            var receivedCookieHeader = "THECOOKIE=value1; Path=/path1";
```
FlorianRappl, line 434: Reason for changing the test?

src/AngleSharp/Annotations/NullableAnnotations.cs
```diff
@@ -1,3 +1,6 @@
+#if !NET5_0_OR_GREATER
+
+// ReSharper disable once CheckNamespace
```
FlorianRappl: These are not necessary. The project does not use ReSharper.

src/AngleSharp/Common/ObjectExtensions.cs
```diff
@@ -165,9 +165,23 @@ public static Double Constraint(this Double value, Double min, Double max)
         public static String GetMessage<T>(this T code)
             where T : struct
         {
-            var field = typeof(T).GetField(code.ToString());
```
FlorianRappl, line 168: Not sure why this code was changed - please revert.

src/AngleSharp/Css/Priority.cs
```diff
@@ -181,7 +181,11 @@ public Priority(Byte inlines, Byte ids, Byte classes, Byte tags)
         /// </summary>
         /// <param name="obj">The object to test with.</param>
         /// <returns>True if the two objects are equal, otherwise false.</returns>
-        public override Boolean Equals(Object obj) => (obj is Priority other) ? Equals(other) : false;
+#if NET5_0_OR_GREATER
+        public override Boolean Equals(Object? obj) => obj is Priority other && Equals(other);
```
FlorianRappl, line 184: This is a change I do not understand. The old signature should still be valid and accepted.

src/AngleSharp/Dom/Url.cs
```diff
@@ -443,6 +447,16 @@ public override Boolean Equals(Object obj)
         /// <returns>
         /// True if the given url is equal to the current url, otherwise false.
         /// </returns>
+#if NET5_0_OR_GREATER
+        public Boolean Equals(Url? other)
```
FlorianRappl, line 446: I did not write this for the others, but my first comment on the `Equals` signature applies to all. Avoid code duplication here and only have a single signature.

src/AngleSharp/Html/Dom/Internal/HtmlTableElement.cs
```diff
@@ -29,24 +29,38 @@ public HtmlTableElement(Document owner, String? prefix = null)
...
-        public IHtmlTableCaptionElement Caption
+        public IHtmlTableCaptionElement? Caption
         {
             get => ChildNodes.OfType<IHtmlTableCaptionElement>().FirstOrDefault(m => m.LocalName.Is(TagNames.Caption));
-            set { DeleteCaption(); InsertChild(0, value); }
+            set
+            {
+                DeleteCaption();
+
+                if (value != null) InsertChild(0, value);
```
FlorianRappl: Use braces - we exclusively use block statements.

FlorianRappl (commented):
Great - thanks!

src/AngleSharp.Core.Tests/Library/CookieHandling.cs
silkfire: Starting from .NET 5, the `CookieContainer` is more strict on how it assigns the subpath of the cookie. It doesn't parse it from the URL argument, instead it has to be supplied explicitly according to the RFC specification. Without this, the unit test will fail because it will assign it to the `/` path instead of `/path1`.

src/AngleSharp/Annotations/NullableAnnotations.cs
silkfire: Feel free to remove them, I'm getting quite alot of similar warnings throughout the solution.

src/AngleSharp/Common/ObjectExtensions.cs
silkfire: I was getting nullability reference errors here, so had to adjust the logic.

src/AngleSharp/Css/Priority.cs
silkfire: The signature since .NET 5 includes nullable reference operator: https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-6.0

src/AngleSharp/Dom/Url.cs
silkfire: See above. Without fixing the nullable reference errors I wasn't able to compile the solution.

src/AngleSharp/Html/Dom/Internal/HtmlTableElement.cs
silkfire: Not sure I understand what you mean?

silkfire:
The CI fails because the SDK being used is .NET 5 which is not able to compile a .NET 6 project.

src/AngleSharp/Html/Dom/Internal/HtmlTableElement.cs
FlorianRappl:
```cs
if (value != null)
{
  InsertChild(0, value);
}
```

src/AngleSharp/Annotations/NullableAnnotations.cs
FlorianRappl: Yes please remove them (it's your PR).

src/AngleSharp/Html/Dom/Internal/HtmlTableElement.cs
silkfire: Apparently there were quite a lot of files that did not follow this convention; I took the liberty and updated them all.

src/AngleSharp/Dom/Url.cs
FlorianRappl: Sure - but again, we should be able to solve this without any duplication at all. In my opinion one signature is all that's required.

silkfire: I'm fine with that, I think the previous versions of .NET should accept NR.

silkfire:
On a slightly related note, you should perhaps consider replacing the deprecated `WebRequest.Create` and use `HttpClient` instead. For now I've silenced the errors so I could compile the solution.

FlorianRappl:
> On a slightly related note, you should perhaps consider replacing the deprecated WebRequest.Create and use HttpClient instead. For now I've silenced the errors so I could compile the solution.

Use `AngleSharp.Io` for that.

silkfire:
@FlorianRappl What's remaining till the PR can be approved?

FlorianRappl:
@silkfire thanks for the ping. I guess only the one active comment above and CI/CD.

I've just updated the runner to include .NET6 - just pull from `devel` / update the PR and let's try running it again.

silkfire:
@FlorianRappl Made a new push but I think you have to approve the CI/CD workflow to run.

silkfire:
@FlorianRappl Are we ready to merge?

FlorianRappl:
Hi @silkfire yes pretty much. As written the comment in `src/AngleSharp/Common/ObjectExtensions.cs` still needs to be resolved.

silkfire:
As I mentioned in the comment, I rewrote the logic to get rid of the nullability errors. The functionality remains the same. `typeof(T).GetField(code.ToString())` returns a `FieldInfo?` which may or may not be `null`. The previous code does not check for `null` here, hence the error.

FlorianRappl:
Might this one work? It's just one more character but it does the job.

```cs
var field = typeof(T).GetField(code.ToString());
var description = field?.GetCustomAttribute<DomDescriptionAttribute>()?.Description;
return description ?? "An unknown error occurred.";
```

I guess the improvement to take is that instead of a struct I'd say let's have this an `enum` (I think it can be constrained to an enum).

silkfire:
What enum are you referring to?

FlorianRappl:
It is used with multiple enums; what I was after is that I think you can now improve the constraint instead of `where T : struct` have `where T : enum`.

silkfire:
@FlorianRappl Reverted the method, added a null conditional on the field and changed the generic constraint.

Should be good to go :)

FlorianRappl (approved):
LGTM! Awesome work :beers:

PR #1050, opened 2022-06-29
None-breaking refactorings/improvements

egil:
- refactor: detect if items passed to GetItemByIndex implement IReadOnlyList<T> for faster retrieval of item by index

# Types of Changes

## Prerequisites

## Contribution Type

## Description

These are small improvements to certain types, methods and classes, that aim to make AngleSharp easier to work with and more efficient.

### DIFF
--- a/src/AngleSharp/Common/ObjectExtensions.cs
+++ b/src/AngleSharp/Common/ObjectExtensions.cs
@@ -1,10 +1,10 @@
 namespace AngleSharp.Common
 {
-    using AngleSharp.Attributes;
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Reflection;
+    using AngleSharp.Attributes;
 
     /// <summary>
     /// Some methods for working with bare objects.
@@ -44,6 +44,11 @@ namespace AngleSharp.Common
         /// <returns>The item at the specified index.</returns>
         public static T GetItemByIndex<T>(this IEnumerable<T> items, Int32 index)
         {
+            if (items is IReadOnlyList<T> list)
+            {
+                return list[index];
+            }
+
             if (index >= 0)
             {
                 var i = 0;
@@ -139,8 +144,8 @@ namespace AngleSharp.Common
         /// <returns>The value or the provided fallback.</returns>
         [return: NotNullIfNotNull("defaultValue")]
         public static U? GetOrDefault<T, U>(this IDictionary<T, U> values, T key, U? defaultValue)
-            where T: notnull
-            where U: notnull
+            where T : notnull
+            where U : notnull
         {
             return values.TryGetValue(key, out var value) ? value : defaultValue;
         }


egil:
Leaving this as draft for now since I might stumble onto more things like this as I explore the code base.

Feel free to decline all or some of the changes.

src/AngleSharp/Common/ObjectExtensions.cs
```diff
@@ -44,6 +44,13 @@ public static class ObjectExtensions
         /// <returns>The item at the specified index.</returns>
         public static T GetItemByIndex<T>(this IEnumerable<T> items, Int32 index)
         {
+            if (items is IReadOnlyList<T> list)
+            {
+                return list.Count > index
+                    ? list[index]
```
FlorianRappl: What if index is negative?

src/AngleSharp/Dom/Internal/HtmlCollection.cs
```diff
@@ -45,6 +45,8 @@ public HtmlCollection(INode parent, Boolean deep = true, Func<T, bool>? predicat
 
         public Int32 Length => _elements.Count();
 
+        public Int32 Count => Length;
```
FlorianRappl: Personally, I think we should not introduce this. If we introduce it - then make it only "visible" when the type is the IReadOnlyList. Otherwise, having two things for the same seems confusing to me.

FlorianRappl (commented):
Thanks!

src/AngleSharp/Common/ObjectExtensions.cs
egil: Then an "out of range exception" is thrown. I guess we could skip the is too large check to, as IReadOnlyList will do this for us too.

egil: Yeah, so use explicitly implement IReadOnlyList. I'll adjust.

FlorianRappl: Yes - sure. My question was rather pointed in the direction: Why invent a new method when there is already a method used below? The iterator should also work on the read only list, right?

On the other hand optimizing the access here seems fine, too. Just returning `return list[index]` would then be the best as suggested.

src/AngleSharp/Dom/Internal/HtmlCollection.cs
egil: I have changed this to be explicitly implemented. If users access the collections as a IReadOnlyList they get the count method, otherwise they dont see it.

src/AngleSharp/Common/ObjectExtensions.cs
egil: Fixed

src/AngleSharp/Dom/INodeList.cs
```diff
@@ -17,7 +17,7 @@ public interface INodeList : IEnumerable<INode>, IMarkupFormattable
         /// <returns>The element at the given index.</returns>
         [DomName("item")]
         [DomAccessor(Accessors.Getter)]
-        INode this[Int32 index] { get; }
+        new INode this[Int32 index] { get; }
```
FlorianRappl: This is still a bit of an annoying part.

The alternative would be that only the `INodeList` implementing classes would implement `IReadOnlyList`. This way, it also would not be breaking.

Right now anyone out there aleady implementing `INodeList` would need to implement `IReadOnlyList`-  mnaking it breaking imho.

egil: Ahh yeah, you are right of course. This is a breaking change.

Could make NodeList implement IReadOnlyList, but then it's invisible to the user, and almost no value is gained. My idea with IReadOnlyList was to make it easier to use it since that interface is more generally applicable than INodeList.

I'll revert.

FlorianRappl: Well you can still cast - and that's the value that is gained imho.

But yes, it's not implicit as now and therefore not as useful - fully agreed.

egil:
I promised non-breaking changes, so Ive reverted the changes that was not.

FlorianRappl:
Thanks! This is now part of `devel`!

PR #1066, opened 2022-10-18
Fixing performance issue with deeply nested structures 

heinrich-ulbricht:
Fixing performance issue with deeply nested structures (https://github.com/AngleSharp/AngleSharp.Xml/issues/18)

# Types of Changes

## Prerequisites

## Contribution Type

## Description

This fixes the parent hierarchy being traversed over and over again. Testing with the structure from https://github.com/AngleSharp/AngleSharp.Xml/issues/18 the time goes down from 35 seconds to ~500 ms.

This might not be the preferred approach to fixing this issue. It's something that works for my specific case and could be fix. Although at the price of introducing a new property to `IElement`. Which will require other implementors of `IElement` to be extended as well, although the functionality stays the same (is this a breaking change?).


FlorianRappl:
Thanks! This is now part of `devel`!

Issue #1052, opened 2022-07-01
TagClosedWrong error when having nested equal formatting elements


gamezjulian:
# Bug Report

## Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version of AngleSharp? No, 0.16.1. But it is also replicable in the latest version.
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository? (there are multiple AngleSharp libraries, e.g., `AngleSharp.Css` for CSS support)
- [x] Did you perform a search in the issues?

## Description

Working on a functionallity to sanitize HTML I found out that with some specifc html structure the parser fails with a TagClosedWrong error. However, the html is well formed. The issue seems to be related to the number of nested formatting elements 
withing the provided html document. If the same html element is repeated 3 times (one inside the other), some of them are removed causing that, in later validations, these nodes are detected as not properly closed elements. 

## Steps to Reproduce

1. Make sure you are running a code that has a method attached to the Error handler, like this:
```c
var defaultHtmlParser = new HtmlParser(default, BrowsingContext.New(Configuration.Default.WithCss(new CssParserOptions
            {
                IsIncludingUnknownDeclarations = true,
                IsIncludingUnknownRules = true,
                IsToleratingInvalidSelectors = true,
            })));

            defaultHtmlParser.Error += (sender, e) =>
            {
                var htmlError = e as HtmlErrorEvent;

                if (htmlError != null)
                {
                    this.logger.LogInformation("An error occurred while parsing: {htmlError.Message}", htmlError.Message);
                }
                else
                {
                    this.logger.LogInformation("An error of type {e.Type} occurred while parsing", e.Type);
                }
            };
```
2. Run the ParseDocument method with the following html template
```html
<!doctype html>

<head></head>

<body>
    <div class="jive-rendered-content">
        <p style="margin: 0;">
            <strong style="color: rgba(0, 0, 0, 1); font-size: 14pt; font-family: Calibri;"> <strong
                    style="color: rgba(0, 0, 0, 1); font-size: 14pt; font-family: Calibri;">
                    <strong style="color: rgba(0, 0, 0, 1); font-size: 14pt; font-family: Calibri;">
                        <strong style="color: rgba(0, 0, 0, 1); font-size: 14pt;">
[...]

PR #1074, opened 2022-11-20
Update to net7.0

driekus77:
Also tried to fix failing unit test (https://github.com/AngleSharp/AngleSharp/issues/1052). Became a bit of a hack / exclusion specially for `<strong>` tag(s). (Hence the Work in Progress.)

All UnitTests pass now. But benchmark tests are having some difficulties with HttpClient.

# Types of Changes

## Prerequisites

## Contribution Type

- [x] Bug fix for https://github.com/AngleSharp/AngleSharp/issues/1052 (non-breaking change which fixes an issue, please reference the issue id)

## Description

Update to net7.0 with a possible fix for https://github.com/AngleSharp/AngleSharp/issues/1052.

### DIFF
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -15,16 +15,14 @@
   </ItemGroup>
 
   <ItemGroup>
-      <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0" PrivateAssets="all" />
-      <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-      <PackageReference Include="NUnit" Version="3.13.3" />
-      <PackageReference Include="NUnit3TestAdapter" Version="4.3.1" />
-      <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0" PrivateAssets="all" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.3.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
 
-      <PackageReference Include="Microsoft.Net.Http" Version="2.2.29" Condition="'$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'net461'" />
-
-      <PackageReference Include="System.Net.Primitives" Version="4.3.1" />
-    </ItemGroup>
+    <PackageReference Include="Microsoft.Net.Http" Version="2.2.29" Condition="'$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'net461'" />
+  </ItemGroup>
   
   <ItemGroup>
     <Compile Remove="Xml\DTDTree.cs" />
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net472;net6.0;net7.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
     <IsPackable>false</IsPackable>
@@ -15,14 +15,16 @@
     </ProjectReference>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0" PrivateAssets="all" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="NUnit" Version="3.13.2" />
-    <PackageReference Include="NUnit3TestAdapter" Version="4.2.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
-  </ItemGroup>
+    <ItemGroup>
+      <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0" PrivateAssets="all" />
+      <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+      <PackageReference Include="NUnit" Version="3.13.2" />
+      <PackageReference Include="NUnit3TestAdapter" Version="4.2.0" />
+      <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
 
+      <PackageReference Include="Microsoft.Net.Http" Version="2.2.29" Condition="'$(TargetFramework)' == 'net472'"/>
+    </ItemGroup>
+  
   <ItemGroup>
     <Compile Remove="Xml\DTDTree.cs" />
     <Compile Remove="Xml\XmlDTD.cs" />
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net6.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net6.0;net7.0</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;net472;net6.0;net7.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
@@ -23,7 +23,7 @@
 
     <PackageReference Include="Microsoft.Net.Http" Version="2.2.29" Condition="'$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'net461'" />
   </ItemGroup>
-  
+
   <ItemGroup>
     <Compile Remove="Xml\DTDTree.cs" />
     <Compile Remove="Xml\XmlDTD.cs" />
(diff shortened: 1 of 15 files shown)


src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
```diff
@@ -15,14 +15,16 @@
...
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="GitHubActionsTestLogger" Version="2.0.0" PrivateAssets="all" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="NUnit" Version="3.13.2" />
-    <PackageReference Include="NUnit3TestAdapter" Version="4.2.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
-  </ItemGroup>
+    <ItemGroup>
```
FlorianRappl: Here we get inconsistent spaces - can you make this consistent please?

src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
```diff
@@ -3186,7 +3190,12 @@ private void InBodyEndTagAnythingElse(HtmlTagToken tag)
                 }
                 else if ((node.Flags & NodeFlags.Special) == NodeFlags.Special)
                 {
-                    RaiseErrorOccurred(HtmlParseError.TagClosedWrong, tag);
+                    // Make exception for strong tag(s) to get TagClosedWrongWithNestedEqualFormattingElements_Issue1052 passed.
```
FlorianRappl: That is not the solution here - the problem is not a strong tag, but the general formatting algorithm. For details #1052.

src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
```diff
@@ -3029,7 +3029,11 @@ private void HeisenbergAlgorithm(HtmlTagToken tag)
 
                 if (openIndex != _openElements.Count - 1)
                 {
-                    RaiseErrorOccurred(HtmlParseError.TagClosedWrong, tag);
+                    // Make exception for strong tag(s) to get TagClosedWrongWithNestedEqualFormattingElements_Issue1052 passed.
+                    if (!(tag.Name.Is("strong") && tag.Type == HtmlTokenType.EndTag))
```
FlorianRappl: That is also not the solution here; see #1052.

(Yes we have one tag where `strong` is used, but it could be any of the formatting elements. Also a tag closed wrong may appear due to other reasons, too - which might be all valid and should not do this.)

FlorianRappl (commented):
Looks good in general, but the CI/CD should also include net7.0.

For this, I think, you need to include the target here: https://github.com/AngleSharp/AngleSharp/blob/devel/build.cake#L8

FlorianRappl:
This is great - thanks for the contribution.

The issue is a long outstanding issue that I worked on a bit (maybe 1-2 hours), but finding a solution was not possible in that time frame. The issue itself is clear, the question is how to solve it (the issue is in the formatting algorithm and especially in the part where it takes elements off from the formatting stack; it must close them, but it should remain consistent)

FlorianRappl:
(For the PR don't feel blocked by the broken unit test - it is broken (because it is a real issue), but we should not make a fix that makes this one case "unbroken", but (1) introduces new issues and (2) does not solve the issue in general; only in the one provided case).

src/AngleSharp/AngleSharp.Core.csproj
```diff
@@ -17,6 +17,7 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <NoWarn>$(NoWarn);SYSLIB0014</NoWarn>
+    <TargetFrameworks>netstandard2.0;net472;net6.0;net7.0</TargetFrameworks>
```
campersau: Maybe update line 6 instead?

FlorianRappl: Great point - thanks for spotting.

driekus77: Yes found it already thanks. Busy getting the build fixed.... For some reasons the multiple unit tests only are only running the net6.0 and net7.0....

driekus77:
> (For the PR don't feel blocked by the broken unit test - it is broken (because it is a real issue), but we should not make a fix that makes this one case "unbroken", but (1) introduces new issues and (2) does not solve the issue in general; only in the one provided case).

You are totally right. Thats why this is a WIP / Draft PR. I identified the 2 spots where the issue originates from. Any tips for real / proper fix are welcome. :-)

src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
driekus77: Sorry. Fixed it.

src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
driekus77: Yep already reverted the "fix"(hack).

src/AngleSharp/Html/Parser/HtmlDomBuilder.cs
driekus77: Already reverted the fix (hack).

driekus77:
The Windows build is now oke but fails on the 1 or 2 Unittests that fail:
Failed!  - Failed:     1, Passed:  3508, Skipped:     2, Total:  3511, Duration: 46 s - AngleSharp.Core.Tests.dll (net7.0)
An error occurred when executing task 'Run-Unit-Tests'.
Error: One or more errors occurred.
.NET Core CLI: Process returned an error (exit code 1).

The Linux build fails in finding nuget.exe (as it seems like).

On my Windows 11 machine the Unittests results are:
image

driekus77:
The "new" failing Unittests are having problems with the CookieContainer in combination with an empty "" Url/Uri.
This error only occurs for net461 and net472. net6.0 and net7.0 work fine.

FlorianRappl:
> Any tips for real / proper fix are welcome. :-)

Yeah I wish I could give some here. I'll reserve some time this week to work on that issue.

> The "new" failing Unittests are having problems with the CookieContainer in combination with an empty "" Url/Uri. This error only occurs for net461 and net472. net6.0 and net7.0 work fine.

This is a bit strange (as those targets did not really change) - but I'll look into it.

driekus77:
> > Any tips for real / proper fix are welcome. :-)
> 
> Yeah I wish I could give some here. I'll reserve some time this week to work on that issue.
Thanks!
 
> > The "new" failing Unittests are having problems with the CookieContainer in combination with an empty "" Url/Uri. This error only occurs for net461 and net472. net6.0 and net7.0 work fine.
> 
> This is a bit strange (as those targets did not really change) - but I'll look into it.

The "old" TargetFramework for the unittests project explains why these errors didn't come up before.
```xml
<TargetFrameworks>net6.0</TargetFrameworks> 
```

Now multitargeting the UnitTests to:
```xml
    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net6.0</TargetFrameworks>
    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;net472;net6.0;net7.0</TargetFrameworks>
```

Kind regards, Henry

FlorianRappl:
Yeah the (.NET) cookie container changed a bit - but I think this is not AngleSharp's problem (if one wants to have a proper cookie container implementation one should use the custom container from `AngleSharp.Io`, which follows the reference implementation from Netscape / Firefox - which is pretty much the same as in all current browsers).

So maybe we should exclude some tests for certain frameworks or adjust their expectations to be consistent. I'll think about it. Thanks for following up on that @driekus77 !

driekus77:
I'm busy trying to fix the build on Linux by updating to newest Cake.Tool version (3.0.0).
On Windows this succeeds, again with failing unit tests. 

Just realized that **Mono support** is **gone** in this Cake version. 

How desperate does AngelSharp needs to build with Mono?

Is the Cake.Tool update a possibility for you guys?

driekus77:
Had both Windows and Linux build green thanks to ContinueOnError in "Run-Unit-Tests"  Cake Task.
Now removed this ContinueOnError and leave the build waiting on fixes for the Unit Tests.

For my part this PR is ready for review.

FlorianRappl:
> How desperate does AngelSharp needs to build with Mono?

I'd say that is no longer required - it was back in the old days when no other option existed on nix systems.

> Is the Cake.Tool update a possibility for you guys?

Definitely - in the mid/long run I'd envision we migrate away from Cake though to [nuke](https://nuke.build/).

FlorianRappl:
> For my part this PR is ready for review.

Awesome - you are a true hero! :rocket:

driekus77:
> > How desperate does AngelSharp needs to build with Mono?
> 
> I'd say that is no longer required - it was back in the old days when no other option existed on nix systems.
> 
> > Is the Cake.Tool update a possibility for you guys?
> 
> Definitely - in the mid/long run I'd envision we migrate away from Cake though to [nuke](https://nuke.build/).

Don't know nuke but I can give it a try if you want? Looks promising from there website!

FlorianRappl:
> Don't know nuke but I can give it a try if you want? Looks promising from there website!

That would be superb - also regarding this PR. I will be able to make a final review / merge later today or tomorrow. Just at the moment caught up in project work.

src/AngleSharp.nuspec
```diff
@@ -26,6 +26,14 @@
         <dependency id="System.Buffers" version="4.5.1" />
         <dependency id="System.Text.Encoding.CodePages" version="6.0.0" />
       </group>
+      <group targetFramework="net60">
+        <dependency id="System.Buffers" version="4.5.1" />
+        <dependency id="System.Text.Encoding.CodePages" version="7.0.0" />
```
FlorianRappl: I think this should be 6.0.0 or?

Alternatively, if their matrix (https://www.nuget.org/packages/System.Text.Encoding.CodePages#supportedframeworks-body-tab) is correct, then we should also change the version to 7.0.0 for the netstandard2.0 target.

tools/anglesharp.cake
```diff
@@ -30,6 +30,7 @@ if (!isRunningOnWindows)
 {
     frameworks.Remove("net461");
     frameworks.Remove("net472");
+    frameworks.Remove("net7.0");
```
FlorianRappl: Well, imho this should work on nix systems, too. Any reason to remove? The other two frameworks have been removed as they are not available as `dotnet` (i.e., .NET Core) frameworks.

FlorianRappl (commented):
I think this looks great - I think we can merge this. :rocket:

src/AngleSharp.nuspec
driekus77: Fixed it.

tools/anglesharp.cake
driekus77: Removed the line so net7.0 wil build on *nix.

FlorianRappl: Hm .NET 7 should also be available for Linux and Mac, right (https://dotnet.microsoft.com/en-us/download/dotnet/7.0)?

driekus77: Yes thats why I removed the net7.0 line from frameworks dict especially for *nix.

FlorianRappl (approved):
Great work - thanks!

PR #1080, opened 2023-01-02
Count specificity properly for various pseudo classes with exceptional rules

patrikwlund:
# Types of Changes

## Prerequisites

## Contribution Type

NOTE: It's breaking in the sense that selector priorities change to fit the spec. If that's not considered breaking, then please consider it a bug fix.

## Description

Specificity calculations are incomplete when using certain pseudo classes.
- `:has()`
- `:matches()`
- `:not()`
- `:nth-child()`
- `:nth-last-child()`

They all incorrectly count as 1 CLASS in weight when they should actually contribute the weight of their most specific inner selector to the total specificity. Additionally, while the `nth-child` classes should count themselves as 1 CLASS in addition to the contained selector, the others should only count their contained selector.

Example: `foo:matches(.bar, #bar)`
Expected: `1 0 1` (one ID, one TAG)
AngleSharp: `0 1 1` (one CLASS, one TAG)

See the spec about specificity here: [[spec]](https://w3c.github.io/csswg-drafts/selectors/#specificity-rules)

Related to this, the following were missing entirely. I was unsure if it was okay to add them in the same PR, but it's easy enough to just cut them out if desired:
- `:is()` (same as `:matches()` [[spec]](https://w3c.github.io/csswg-drafts/selectors/#matches-pseudo) )
- `:where()` (same as `:matches()` but with 0 contributed specificity)

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -655,13 +655,14 @@ namespace AngleSharp.Core.Tests.Css
             Assert.AreEqual("first", result[0].Id);
         }
 
-        [Test]
-        public void MatchesWithTwoElements()
+        [TestCase("is")]
+        [TestCase("matches")]
+        public void MatchesWithTwoElements(string variant)
         {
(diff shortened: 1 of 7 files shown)


FlorianRappl (approved):
This looks great to me - thanks for your contribution!

PR #1085, opened 2023-01-18
Fix StackOverflowException in StyleExtensions.GetStyleSheets()

### LINKED ISSUE #1084: StackOverflowException when accessing the StyleSheets property
# Description
If you have a sufficiently deep DOM, accessing the `StyleSheets` property on a parsed document causes a StackOverflowException to be thrown by AngleSharp. 

After some investigation I found that this is because the method `StyleExtensions.GetStyleSheets()` uses the program stack to iterate the DOM in a recursive way.

## Background
Our application running in IIS somtimes see StackOverflowExceptions when sanitizing HTML documents. We use [HtmlSanitizer](https://github.com/mganss/HtmlSanitizer) which uses AngleSharp to parse the document. This [issue](https://github.com/mganss/HtmlSanitizer/issues/417) is the same as we are having and there was a similar issue in HtmlSanitizer, which is now fixed. However, we still get StackOverflowException from AngleSharp. 

When we get a StackOverflowException, the entire app crash and needs to be restarted by IIS.

When an application is running in IIS it has a relatively small stack size compared to if you are running a console application (256KB for 32bit app and 512KB for 64bit app). Hence the demo application below tries to replicate the "IIS environment".

## Steps to reproduce:

Using this sample application with a relatively low stack size similar to what an application running in IIS gets by default. Create a new solution with this code, add AngleSharp (v1.0.1) and AngleSharp.Css (v.0.17.0) Nuget packages. Build and run the application and observe it throws StackOverflowException in `StyleExtensions.GetStyleSheets()`.

```c#
using AngleSharp;
using AngleSharp.Html.Parser;

var thread = new Thread(() =>
{
    var beginTags = String.Join("", Enumerable.Repeat("<div>", 10000));
    var endTags = String.Join("", Enumerable.Repeat("</div>", 10000));
    var html = $"<html><head><style></style></head><body>{beginTags}<style></style>{endTags}</body></html>";

    var context = BrowsingContext.New(Configuration.Default.WithCss());
    var document = context.GetService<IHtmlParser>()!.ParseDocument(html);

    var sheets = document.StyleSheets;
    Console.WriteLine(sheets.Length);
}, 512 * 1024);

thread.Start();
thread.Join();
```

## Possible solution
While we could try to mess with stack sizes in our application, create new threads and similar - a better solution would be if AngleSharp did not throw an exception in the first place. 

A way to achieve this is to use a stack on the heap instead of the program stack when iterating the DOM. I am currently working on a pull request with this change.

bjornri:
## Contribution Type

## Description

Refer to issue https://github.com/AngleSharp/AngleSharp/issues/1084. Reimplemented `StyleExtensions.GetStyleSheets()` to use a stack on the heap instead of the program stack to avoid stack overflow exceptions. I have kept the original depth-first ordering of the return value.

I did make a unit test, but because AngleSharp.Css is not available here I believe I need to mock the styling service and return a dummy object to be able to assert something. To keep the execution time low I created a thread with a small stack size. Using the main thread gave an execution time of 3 seconds or so on "my machine" because we need a very deep DOM. Using a 64KB stack made the unit test run in just 85ms (also on "my machine").

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/StyleExtensions.cs
+++ b/src/AngleSharp.Core.Tests/Css/StyleExtensions.cs
@@ -1,0 +1,90 @@
+namespace AngleSharp.Core.Tests.Css
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AngleSharp.Css;
+    using AngleSharp.Css.Dom;
+    using AngleSharp.Html.Parser;
+    using AngleSharp.Io;
+    using Dom;
+    using Mocks;
+    using NUnit.Framework;
+    using Text;
+
+    public class StylesheetExtensions
+    {
+        [Test]
+        public void TestStyleSheetsDoesNotThrowStackOverflowException()
+        {
+            var thread = new Thread( () =>
+            {
+                var beginTags = String.Join("", Enumerable.Repeat("<div>", 1000));
+                var endTags = String.Join("", Enumerable.Repeat("</div>", 1000));
+                var html = $"<html><head><style></style></head><body>{beginTags}<span><style></style></span>{endTags}</body></html>";
+
+                var stylingService = new MockStylingService();
+                var context = BrowsingContext.New(Configuration.Default.WithOnly<IStylingService>(stylingService));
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1089, opened 2023-02-21
move to Nullable package

SimonCropp:


src/AngleSharp/AngleSharp.Core.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="5.0.0" PrivateAssets="All" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
+    <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="all" />
```
FlorianRappl, line 24: I guess we also need to have this then as a dependency in the created nuget package?

SimonCropp: nope. it is a dev only dependency

FlorianRappl: Are you sure - because the metadata is still there and we now reference attributes that are in this package?

SimonCropp: not sure what u mean by "metadata is still there". the resulting dll has no ref 
image

FlorianRappl: Metadata is what is needed for reflection (i.e., any type information). It contains among other things the class and method definitions (incl. attributes).

The picture above is misleading in multiple areas - it's the v4.6.1 which should anyway not contain any references (even if the metadata would rely on nullable it should not in the v4.6.1 as nullable is not supported in there).

SimonCropp: the compiler does not use reflection for the attributes. and nullable works in net461

SimonCropp: fwiw i use this approach in over 70 packages over many years

FlorianRappl: > and nullable works in net461

That's right - the package is actually only required for .NET below .NET 5 (it's the other way round).

Anyway, the thing that I missed (my bad) is that we set `PrivateAssets` to `true` - with that the respective code is already part of the generated code and no reference it required. So all good!

FlorianRappl (approved):

PR #1090, opened 2023-02-21
remove some redundant casts

SimonCropp:


src/AngleSharp.Core.Tests/Auxiliary/ConfigurationsExtensions.cs
```diff
@@ -12,7 +12,7 @@ public class ConfigurationsExtensionsTests
         public void WithoutTServiceRemovesItems()
         {
             var config = new Configuration(new object[0])
-                .With((IRequester)new DefaultHttpRequester())
```
FlorianRappl: Danger: This is not redundant.

FlorianRappl (approved):
LGTM!

PR #1092, opened 2023-02-21
remove System.Buffers

SimonCropp:
is it really required?


FlorianRappl:
Previously it was. Should be confirmed that it works without it.

FlorianRappl:
I think this is required. See #963.

SimonCropp:
weird that it still builds?

FlorianRappl:
Depends on the target. My guess is that since System.Buffers exists only up to version 4.5.1 with the targets we have it might not be needed (anymore) and that it was actually previously be over-specified (i.e., its part of all target frameworks and we do not need the explicit reference).

SimonCropp:
well it builds, and tests pass. so IMO we remove it

FlorianRappl (approved):

PR #1095, opened 2023-02-21
using declarations

SimonCropp:


src/AngleSharp.Core.Tests/Assets.cs
```diff
@@ -58,36 +58,28 @@ public static string GetManifestResourceString(string name, Encoding encoding =
         {
             encoding = encoding ?? Encoding.UTF8;
             var fullName = typeof(Assets).Namespace + "." + name;
-            using (var stream = typeof(Assets).Assembly.GetManifestResourceStream(fullName))
+            using var stream = typeof(Assets).Assembly.GetManifestResourceStream(fullName);
```
FlorianRappl, line 61: Well, if we simplify, then we can already do

```cs
using var stream = typeof(Assets).Assembly.GetManifestResourceStream(fullName) ?? throw new ArgumentException($"...");
```

(Same applies to the other places where a direct null assertion is followed)

FlorianRappl (commented):
LGTM in general - just further simplifications / enhancements possible.

FlorianRappl:
This is now anyway part of `devel`.

PR #1102, opened 2023-02-22
move to auto properties

SimonCropp:


src/AngleSharp.Core.Tests/Mocks/DelayRequester.cs
```diff
@@ -10,15 +10,14 @@
...
-        private Int32 _count;
 
         public DelayRequester(Int32 timeout)
         {
             _timeout = timeout;
-            _count = 0;
+            RequestCount = 0;
         }
 
-        public Int32 RequestCount => _count;
```
FlorianRappl, line 21: Keep such constructs - we don't want to call the property inside, but use the backing field directly.

src/AngleSharp/Common/BaseTokenizer.cs
```diff
@@ -70,22 +67,22 @@ protected set
         /// <summary>
         /// Gets the current source index.
         /// </summary>
-        public Int32 Position => _source.Index  - (_normalized ? 1 : 0);
+        public Int32 Position => _source.Index  - (IsNormalized ? 1 : 0);
 
         /// <summary>
         /// Gets the current character.
         /// </summary>
-        protected Char Current => _current;
```
FlorianRappl, line 78: No - keep it, we don't want to go over a property and have a method call here

(same for the others)

FlorianRappl (changes requested):
While some (~60%) of the changes are fine, others introduce unnecessary method calls and should be removed.

In general this seems like a refactoring without actual value added (like many of the other PRs). At this time it would be great to maybe outline what (and why) you want to have these changes?

PR #1103, opened 2023-02-22
IAttr.Value cant be null

SimonCropp:
according to the nullable annotations


src/AngleSharp/Html/HtmlMarkupFormatter.cs
```diff
@@ -90,14 +90,9 @@ protected virtual String Attribute(IAttr attr)
 
             WriteAttributeName(attr, temp);
 
-            if (attr.Value != null)
```
FlorianRappl, line 93: This certainly can be `null`.

Remember that this is an interface and no one is bound to these hints.

FlorianRappl (changes requested):

PR #1104, opened 2023-02-22
TextDataSetEntry._value cant be null

SimonCropp:


src/AngleSharp/Html/Forms/TextDataSetEntry.cs
```diff
@@ -18,7 +18,7 @@ public TextDataSetEntry(String name, String value, String type)
 
         public override Boolean Contains(String boundary, Encoding encoding)
         {
-            return _value != null && _value.Contains(boundary);
+            return _value.Contains(boundary);
```
FlorianRappl, line 21: It can easily be `null` as there is no check - the only usage is via `encoding.WebName` and unless its proven that every possible encoding has a non-`null` `WebName` property it's better to be defensive here.

I think the "cant be null" part can only be used in cases where the only usages are internal (fulfilled in this case) with values we control (not fulfilled in this case).

FlorianRappl (changes requested):

Issue #740, opened 2019-01-07
Missing System.Text.Encoding.CodePages reference


kpkozak:
### Description

Missing reference to System.Text.Encoding.CodePages in latest (0.10.0) version

### Steps to Reproduce

1. Reference AngleSharp NuGet package in your project
2. Use AngleSharp
```
class Program
    {
        static void Main(string[] args)
        {
            var parser = new HtmlParser();
            var result = parser.ParseDocument("<html></html>");
        }
    }
```
3. Result:  Exception
```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'AngleSharp.Configuration' threw an exception. ---> System.IO.FileNotFoundException: Could not lo
ad file or assembly 'System.Text.Encoding.CodePages, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
```

**Expected behavior:** 
AngleSharp does not throw initialization exception

**Actual behavior:** 
Exception thrown. 

**Environment details:** .netcore console app

### Possible Solution
Manually adding reference to mentioned package fixes the issue

FlorianRappl:
Thanks for the report!

I think this is just missing from the NuGet - right? So the workaround is to explicitly reference it at the moment. Could you please confirm this?

kpkozak:
@FlorianRappl yes, that's true. If referencing this explicitly, manually in the project, everything works (I had exactly this on my mind writing "Possible Solution" section) ;)

FlorianRappl:
I'll release a hotfix soon! Thanks for the confirmation!

jgrandydev:
I added to my project the NuGet package `System.Text.Encoding.CodePages` 

Project references now includes `System.Text.Encoding.CodePages` ( Version 4.1.1.0 , Runtime Version v4.0.30319 )

But this did not resolve this issue.

FlorianRappl:
@jgrandydev not sure what you refer to - the latest version (0.11.0) ships with `System.Text.Encoding.CodePages` as a dependency on all platforms (so there is nothing you need to add).

If you face a problem then please describe it with all necessary details, otherwise I won't be able to give you any support.

jeremymeng:
You probably meant to @jgrandydev

FlorianRappl:
Right, thanks for the hint @jeremymeng !

jgrandydev:
[...]

PR #1124, opened 2023-06-26
remove System.Text.Encoding.CodePages in net7

SimonCropp:
doesnt seem to be required based on this
```
  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net461' or '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'net6.0' ">
    <PackageReference Include="System.Text.Encoding.CodePages" Version="6.0.0" />
  </ItemGroup>
```

### DIFF
--- a/src/AngleSharp.nuspec
+++ b/src/AngleSharp.nuspec
@@ -27,9 +27,6 @@
       <group targetFramework="net60">
         <dependency id="System.Text.Encoding.CodePages" version="6.0.0" />
       </group>
-      <group targetFramework="net70">
-        <dependency id="System.Text.Encoding.CodePages" version="7.0.0" />
-      </group>
     </dependencies>
   </metadata>
 </package>


FlorianRappl:
No I rather think it's the other way round. The given line should be extended for .NET 7.

It is never directly required but indirectly it may be required - after all the package brings codepages. You'll only realize that you miss a codepage if a page you want to look at uses a codepage that is not by default included.

Presumably the best way would be to just drop the references from the project and only include them as NuGet references. This way users will always get the codepages.

An alternative is to document that codepages might be needed and that installing the `System.Text.Encoding.CodePages` is required in those cases. Then we could just drop it altogether.

SimonCropp:
guess i was confused as to how that packages is used. thanks for the clarification

FlorianRappl:
Yeah no problem - would still be great to get some insights from someone with experience what approach we should take with this kind of package...

On the one hand we originally received this as a bug report (see #740), on the other hand removing this and leaving it up to the end user would also be fine (after all, that "workaround" already worked in the reported issue).

What's your opinion here @SimonCropp ?

PR #1125, opened 2023-06-29
typo: Descendents => Descendants

SimonCropp:
equivalent method in .net https://learn.microsoft.com/en-us/dotnet/api/system.xml.linq.xcontainer.descendants?view=net-7.0

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/MutationObserver.cs
+++ b/src/AngleSharp.Core.Tests/Library/MutationObserver.cs
@@ -153,7 +153,7 @@ namespace AngleSharp.Core.Tests.Library
         }
 
         [Test]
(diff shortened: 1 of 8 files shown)


FlorianRappl (commented):
This would be a breaking change. Since it's just a typo I'd not merge this as-is right now.

What we can do instead is to mark it as deprecated and have the renamed one (i.e. as an alias) on the side. Otherwise I'd stick with it in the method and have it only corrected in the comments and tests.

Thoughts?

SimonCropp:
> What we can do instead is to mark it as deprecated and have the renamed one (i.e. as an alias) on the side

happy with that. want me to do that as a PR?

FlorianRappl:
Yeah that would be awesome (either change this one or in a new one; I mean many/most changes would be the same)!

🍻

PR #1130, opened 2023-07-26
Remove `delegate HtmlElement Creator` from IElementFactory implementations

maartenba:
# Types of Changes

## Prerequisites

## Contribution Type

- [x] Minor performance improvement

## Description

@matkoch and myself have been profiling the project after OSS Power-Ups, and have not uncovered any glaring issues. There are some micro-optimizations possible here and there that may shave off a bit of memory in high-scale scenarios (such as a string interning mechanism for the `Url` schema field), but we failed to find any glaring big issues.

One performance improvement that shaved off a few % was in optimizing out the dictionary + delegates in `IElementFactory` implementations. Now, there's a `// REVIEW` comment in there that matters. The dictionary-based approach would match tags in a case-insensitive way, which makes sense for some tag names in the library. We have one case (`"foreignobject"` in the `SvgElementFactory` class), where the approach might depend on your preferred code style. Added this case as a second commit for easier review.


maartenba:
Improvement: 265ms to 220ms for `CreateHtmlElement` when running https://gist.github.com/jirisykora83/d1316a636d5f6c154872cc746c7654a1 500 times (still a micro-optimization imo)

src/AngleSharp/Dom/TagNames.cs
```diff
@@ -13,7 +13,7 @@ public static class TagNames
         /// <summary>
         /// Gets the DOCTYPE constant.
         /// </summary>
-        public static readonly String Doctype = "DOCTYPE";
+        public const String Doctype = "DOCTYPE";
```
FlorianRappl: Yeah unfortunately this is a breaking change...

src/AngleSharp/Html/HtmlElementFactory.cs
```diff
@@ -163,12 +20,147 @@ sealed class HtmlElementFactory : IElementFactory<Document, HtmlElement>
         /// <returns>The specialized HTMLElement instance.</returns>
         public HtmlElement Create(Document document, String localName, String? prefix = null, NodeFlags flags = NodeFlags.None)
         {
-            if (creators.TryGetValue(localName, out var creator))
+            // REVIEW: is ToLowerInvariant() the right approach here? are TagNames always lowercase?
```
FlorianRappl: Yes they are.

FlorianRappl:
So `TagNames` is public and therefore we cannot change from static readonly (real references) to const (compile-time replacement).

Regarding the dictionary to pattern matching change; this is a great approach and works with me.

src/AngleSharp/Dom/TagNames.cs
maartenba: Then a simple if/else would do the trick. Would you like me to update the PR in that direction? (and also the lowercase/mixed case tag names)

FlorianRappl: Yeah would be great (do you think if / else if is as fast as a dictionary, which is O(1)?)

maartenba: Fixed the breaking change.

src/AngleSharp/Html/HtmlElementFactory.cs
maartenba: Fixed.

maartenba:
Regarding https://github.com/AngleSharp/AngleSharp/pull/1130#discussion_r1275522677 - only one way to find out - benchmarks!

Here's a branch with a benchmark (don't merge it, adds the old implementations into `AngleSharp.Core`) - https://github.com/maartenba/AngleSharp/tree/devel-benchmarks

The results on a `ShortRunJob` with .NET 4.8.1:
```
|          Method |     Mean |    Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|---------------- |---------:|---------:|----------:|---------:|---------:|--------:|----------:|
|     UsingIfElse| 9.050 ms | 6.235 ms | 0.3418 ms | 421.8750 | 203.1250 | 62.5000 |   2.73 MB |
| UsingDictionary  | 8.494 ms | 2.134 ms | 0.1170 ms | 421.8750 | 203.1250 | 62.5000 |   2.77 MB |
```

Results on a full run with .NET 4.8.1:
```
|          Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
|     UsingIfElse| 8.867 ms | 0.1753 ms | 0.2881 ms | 421.8750 | 171.8750 | 62.5000 |   2.73 MB |
| UsingDictionary  | 8.742 ms | 0.1720 ms | 0.3553 ms | 421.8750 | 171.8750 | 62.5000 |   2.77 MB |
```

The results on a `ShortRunJob` with .NET 7.0:
```
|          Method |     Mean |    Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|---------------- |---------:|---------:|----------:|---------:|---------:|--------:|----------:|
|     UsingIfElse| 4.825 ms | 2.635 ms | 0.1444 ms | 312.5000 | 242.1875 | 70.3125 |    2.6 MB |
| UsingDictionary  | 5.662 ms | 3.545 ms | 0.1943 ms | 312.5000 | 242.1875 | 70.3125 |   2.63 MB |
```

Results on a full run with .NET 7.0:
```
|          Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
|     UsingIfElse| 4.905 ms | 0.0809 ms | 0.1352 ms | 312.5000 | 242.1875 | 70.3125 |    2.6 MB |
| UsingDictionary  | 4.639 ms | 0.0747 ms | 0.0699 ms | 312.5000 | 242.1875 | 70.3125 |   2.63 MB |
```

As mentioned earlier, this is going to be a micro-optimization that may affect high-scale scenarios, but overall it may be unnecessary to make this change in AngleSharp. Seeing the benchmarks with this final set of changes, I'm not sure it makes sense to include.

FlorianRappl:
Yeah thanks for investigating that in all detail; makes sense and I share your conclusion.

Going back to the pattern matching proposal; we could do both - having the `const`s internal (and only used *inside* AngleSharp) and keeping the external `static readonly` ones. The latter would, of course, use the internal ones in their initialization:

```cs
internal const String _Div = "div";
public static readonly String Div = _Div;
```

maartenba:
That's a nice way indeed. For `foreignObject` the casing remains, were you ok with the I ital approach there that did a case insensitive comparison there?

FlorianRappl:
> For `foreignObject` the casing remains, were you ok with the I ital approach there that did a case insensitive comparison there?

Yes, totally - much appreciated!

maartenba:
Updated. Also realized the names of benchmarks in previous run are inverted (so updated them, results still very close).

`ShortRunJob` benchmark on .NET 4.8.1 using pattern matching:
```
|               Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|--------------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
| UsingPatternMatching | 9.754 ms |  1.455 ms | 0.0797 ms | 421.8750 | 203.1250 | 62.5000 |   2.78 MB |
|      UsingDictionary | 9.969 ms | 14.889 ms | 0.8161 ms | 421.8750 | 203.1250 | 62.5000 |   2.77 MB |
```

`ShortRunJob` benchmark on .NET 7.0 using pattern matching:
```
|               Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|--------------------- |---------:|----------:|----------:|---------:|---------:|--------:|----------:|
| UsingPatternMatching | 9.754 ms |  1.455 ms | 0.0797 ms | 421.8750 | 203.1250 | 62.5000 |   2.78 MB |
|      UsingDictionary | 9.969 ms | 14.889 ms | 0.8161 ms | 421.8750 | 203.1250 | 62.5000 |   2.77 MB |
```

Full benchmark on .NET 4.8.1 using pattern matching:
```
|               Method |     Mean |    Error |   StdDev |     Gen0 |     Gen1 |    Gen2 | Allocated |
|--------------------- |---------:|---------:|---------:|---------:|---------:|--------:|----------:|
| UsingPatternMatching | 11.64 ms | 0.345 ms | 1.001 ms | 421.8750 | 171.8750 | 62.5000 |   2.78 MB |
|      UsingDictionary | 11.34 ms | 0.309 ms | 0.912 ms | 421.8750 | 171.8750 | 62.5000 |   2.77 MB |
```

Full benchmark on .NET 7.0 using pattern matching:
```
|               Method |     Mean |     Error |    StdDev |   Median |     Gen0 |     Gen1 |    Gen2 | Allocated |
|--------------------- |---------:|----------:|----------:|---------:|---------:|---------:|--------:|----------:|
| UsingPatternMatching | 5.672 ms | 0.1946 ms | 0.5616 ms | 5.611 ms | 312.5000 | 242.1875 | 70.3125 |    2.6 MB |
|      UsingDictionary | 5.435 ms | 0.1322 ms | 0.3858 ms | 5.248 ms | 312.5000 | 242.1875 | 70.3125 |   2.63 MB |
```

I'd say, dictionaries and delegates are fine, perf-wise :-)

FlorianRappl:
Hm the result is really close; and interestingly perf. in .NET 7 is even better for dictionaries (of course, both ways win with the whole optimizations the .NET team did).

So conclusion would be to keep the dictionary? What's your recommendation?

maartenba:
I'd keep the dictionary indeed. It's as readable as it comes (so dev experience is better), for a non-significant tradeoff.

lahma:
You could maybe go faster than basic dictionary when you know the input. Basically the gist is to check against input length and known paths based in where the first character is different than the others. We did a lookup in [Esprima.NET](https://github.com/sebastienros/esprima-dotnet) to try to optimize the chosen path (source-generated).

https://github.com/sebastienros/esprima-dotnet/blob/main/test/Esprima.Tests.SourceGenerators/Snapshots/StringMatcherGeneratorTests.ScannerStringMatchingGeneration%23StringMatchers.g.verified.cs

PR #1135, opened 2023-08-21
Improve ChildSelector performance

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

After the last optimization pass the child selector performance didn't look all right to me be, seemed like an outlier compared to other selectors. I went through the code and tried to improve things. Used the struct via generic constraint interface trick to get rid of interface dispatch still allowing both `INodeList` and `NodeList` to be selector's input. The gist is to have a second `DoMatch` method that is being called that ensures less interface dispatch - struct will be inlined in terms of calls as its compile time constant.

Please note, I would highly recommend removing the `INodeList` interface and just substituting it with concrete sealed type `NodeList` (public API checked). `NodeList` is the only implementation after all. The interface dispatch has real cost here for such integral type. Of course it would be major breaking change. The interface also returns `INode` instead of `Node` unlike what `NodeList` does, all access to node properties will cause interface dispatch again.

## Benchmarks

```

BenchmarkDotNet v0.13.7, Windows 11 (10.0.23526.1000)
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK 8.0.100-preview.7.23376.3
  [Host]   : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2
  ShortRun : .NET 6.0.21 (6.0.2123.36311), X64 RyuJIT AVX2

Job=ShortRun  IterationCount=3  LaunchCount=1  
WarmupCount=3  

```

## AngleSharp.Benchmarks.SelectorBenchmark

| **Diff**|Selector|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |p:first-child|113.7 μs|2.92 μs|17.98 KB|
| **New** | **p:first-child** | **111.6 μs (-2%)** | **6.88 μs** | **17.98 KB (0%)** |
| Old |p:last-child|116.8 μs|4.33 μs|17.46 KB|
| **New** | **p:last-child** | **115.4 μs (-1%)** | **4.57 μs** | **17.46 KB (0%)** |
| Old |p:nth-child(2n)|673.9 μs|7.17 μs|21.17 KB|
| **New** | **p:nth-child(2n)** | **326.1 μs (-52%)** | **18.40 μs** | **21.18 KB (0%)** |
| Old |p:nth-child(2n+1)|679.3 μs|5.51 μs|21.2 KB|
| **New** | **p:nth-child(2n+1)** | **324.8 μs (-52%)** | **27.78 μs** | **21.21 KB (0%)** |
| Old |p:nth-child(even)|683.4 μs|125.11 μs|21.15 KB|
| **New** | **p:nth-child(even)** | **316.8 μs (-54%)** | **9.39 μs** | **21.15 KB (0%)** |
| Old |p:nth-child(n)|676.9 μs|81.74 μs|25.17 KB|
| **New** | **p:nth-child(n)** | **327.5 μs (-52%)** | **7.90 μs** | **25.17 KB (0%)** |
| Old |p:nth-child(odd)|683.7 μs|34.28 μs|21.15 KB|
| **New** | **p:nth-child(odd)** | **320.6 μs (-53%)** | **18.94 μs** | **21.16 KB (0%)** |
| Old |p:only-child|216.1 μs|6.90 μs|16.95 KB|
| **New** | **p:only-child** | **217.5 μs (+1%)** | **13.76 μs** | **16.95 KB (0%)** |

### DIFF
--- a/src/AngleSharp/Dom/Internal/NodeList.cs
+++ b/src/AngleSharp/Dom/Internal/NodeList.cs
@@ -13,7 +13,7 @@ namespace AngleSharp.Dom
     {
         #region Fields
 
-        private readonly List<Node> _entries;
+        internal readonly List<Node> _entries;
 
         /// <summary>
         /// Gets an empty node-list. Shouldn't be modified.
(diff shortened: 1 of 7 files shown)


src/AngleSharp/Dom/Internal/NodeList.cs
```diff
@@ -97,4 +97,57 @@ public void ToHtml(TextWriter writer, IMarkupFormatter formatter)
 
         #endregion
     }
+
+    /// <summary>
+    /// Helper interface which can remove interface dispatch overhead when used in combination with generic methods.
+    /// </summary>
+    internal interface INodeListAccessor
```
campersau, line 104: Why do you need this interface when you could just use `T : INodeList` like in the other PR?
Does the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` for the `InterfaceNodeListAccessor` improve things?

lahma: With just `T : INodeList` I cannot bypass the interface's indexer as compiler isn't always smart enough to trust that `NodeList`'s indexer can be accessed directly. Profiler traces (and debugger) showed hitting the explicit interface implementation which redirects to concrete one (+ extra access + `INode` as return value which is again interface). This is why I opted to use this more verbose version which goes directly to concrete type.

In other PR it was compile-time assured that type is `NodeList`, here I'm casting it.

Like mentioned in the description, life would be easier (performance-wise) without the interface.

FlorianRappl (approved):
LGTM - great one. Thanks for your insights and contributions! :beers:

PR #1148, opened 2023-11-15
Fix Range bug Issue

### LINKED ISSUE #1147: Range's functions StartWith and EndWith don't set the End offset
### Prerequisites

### Description

I happened to use `ClearContent` function on AngleSharp.Dom.IRange 

To use that I created a Range object from an existing document and then set both start & end nodes.
After that I intended to clear the content within the selection of the range but it didn't.
Later when I debug my code and I found that after it called `StartWith()` it did set the `Start` offset property in the range object.
But when it called `EndWith()` function afterwards, it didn't set `End` property, i.e. didn't change at all (Usually in VS debug mode we can see whether a variable did change within Locals panel).

So I trace the code in the github and saw at src/AngleSharp.Core.Tests/Library/Range.cs:
```csharp
...
            if (!_start.IsExplicit)
            {
                _start = bp;
                _end = bp;
            }
            else if (bp < _start || Root != refNode.GetRoot())
            {
                _end = bp;
            }
...
```

The expression `bp < _start` compares two objects with the class Boundary, the operator function I saw:
```csharp
        private readonly struct Boundary : IEquatable<Boundary>
        {
...
            public static Boolean operator >(Boundary a, Boundary b)
            {
                return false;
            }

            public static Boolean operator <(Boundary a, Boundary b)
            {
                return false;
            }
...
```

It by default returns false value. 
Which I believe it is the reason why my ClearContent didn't work due to End property not being set.

### Steps to Reproduce

I have this function to remove between strings
```csharp
        private static void RemoveBetweenStrings(string start, string end, INode nodeStart, INode nodeEnd, IDocument doc)
        {
            var parentStart = nodeStart.Parent;
            var parentEnd = nodeEnd.Parent;
            var range = doc.CreateRange();
            var offsetStart = nodeStart.NodeValue.IndexOf(start); // I checked this value, non-zero, non-negative
            var offsetEnd = nodeEnd.NodeValue.IndexOf(end) + end.Length; // this too

            range.StartWith(nodeStart, offsetStart); // successfully set both Start and End property
            range.EndWith(nodeEnd, offsetEnd); // didn't set/modify the End property

            range.ClearContent(); // didn't clear the content in-between start & node range

            Action<INode> removeIfEmpty = (node) => {
                var text = node.Text();
                if (doc.Contains(node) && string.IsNullOrWhiteSpace(text))
                    node.Parent.RemoveChild(node);
            };
            removeIfEmpty(parentStart);
            removeIfEmpty(parentEnd);
        }
```

so I can test it using parser and AngleSharp.XPath:
```csharp
string start = "{!START}";
string end = "{!END}";
var doc = parser.ParseDocument(@"<html>
<head></head>
<body>
<p> no deletion before start {!START} this should be cleared </p>
<p> this should be deleted too {!END} This is not to be deleted</p>
<p> this should not be deleted either</p>
</body></html>");
var startNode = doc.Body.SelectSingleNode($"//*//text()[contains(., '{start}')]");
var endNode = doc.Body.SelectSingleNode($"//*//text()[contains(., '{end}')]");
RemoveBetweenStrings(start, end, startNode, endNode, doc); //  this is an expected bug, not being able to delete
```

### Expected Behavior

I think it's because of
```csharp
public static Boolean operator <(Boundary a, Boundary b)
{
          return false;
}
```

the operator should return the actual position comparison value between both nodes within the html dom tree.

### Actual Behavior

This piece of code didn't set the end property after setting the start property
```csharp
range.EndWith(nodeEnd, offsetEnd); // didn't set/modify the End property
```

I saw the algorithm to set whether the boundary point is equal, before, or after in this ref:
https://dom.spec.whatwg.org/#boundary-points

zukarusan:
# Types of Changes

## Prerequisites

## Contribution Type

## Description
Ref: [#1147](https://github.com/AngleSharp/AngleSharp/issues/1147)
Fix the issue of not setting up `Start` and `End` property after one of both is explicitly given. The core fix is to modify `CompareTo` function according to the described algorithm from https://dom.spec.whatwg.org/#boundary-points Additionally, added more operator functions with overridden `Equals` & `GetHasCode` functions as well.

Also, the test case for this issue is added in Library/Range.cs.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -54,5 +54,23 @@ namespace AngleSharp.Core.Tests.Library
             Assert.AreEqual(common, range.CommonAncestor);
             Assert.IsFalse(range.IsCollapsed);
         }
+
+        [Test]
+        public void CanSelectRangeSomeRange_Issue1147()
+        {
+            var document = "<body></body>".ToHtmlDocument();
+            var text1 = document.Body.AppendChild(document.CreateTextNode("Text1"));
+            var text2 = document.Body.AppendChild(document.CreateTextNode("TextLonger2"));
+            var range = document.CreateRange();
+            range.StartWith(text1, "Text".Length);
+            range.EndWith(text2, "TextLonger".Length);
+
+            Assert.AreEqual("Text".Length, range.Start);
+            Assert.AreEqual(text1, range.Head);
+            Assert.AreEqual("TextLonger".Length, range.End);
+            Assert.AreEqual(text2, range.Tail);
+            Assert.AreEqual(document.Body, range.CommonAncestor);
+            Assert.IsFalse(range.IsCollapsed);
+        }
     }
 }
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -1,7 +1,6 @@
 namespace AngleSharp.Core.Tests.Library
 {
     using NUnit.Framework;
-    using static System.Net.Mime.MediaTypeNames;
 
     [TestFixture]
     public class RangeTests
@@ -45,7 +44,7 @@ namespace AngleSharp.Core.Tests.Library
             var text2 = common.LastChild;
             var range = document.CreateRange();
             range.StartBefore(text1);
-            range.EndAfter(text2);
+            range.EndBefore(text2);
 
             Assert.AreEqual(0, range.Start);
             Assert.AreEqual(text1.Parent, range.Head);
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
Great - yes this looks good to me. I will need to look into the W3C proposed algorithm, but from a first glance / the test it works wonderfully.

Much appreciated!

src/AngleSharp/Dom/INode.cs
```diff
@@ -243,5 +243,10 @@ public interface INode : IEventTarget, IMarkupFormattable
         /// Gets the associated node flags.
         /// </summary>
         NodeFlags Flags { get; }
+
+        /// <summary>
+        /// Gets the length of the node - see https://dom.spec.whatwg.org/#concept-node-length
+        /// </summary>
+        Int32 NodeLength { get; }
```
FlorianRappl: Please remove this. It would be a breaking change and usually such algorithms are rather implemented as helper functions.

In addition most things on the INode are only there if there is an official API for it - so unless there is a DomName its better to not put it on the INode.

FlorianRappl (changes requested):

src/AngleSharp/Dom/INode.cs
zukarusan: Hi @FlorianRappl 
Thank you for your kind review

I apologize for my bad habit of pushing into my local without realizing it's auto-creating pull requests.

It's now reset with another commit
https://github.com/AngleSharp/AngleSharp/pull/1148/commits/07cbdfffe5ee2cab4cdec45fd2b6f591176579fe

FlorianRappl (approved):
LGTM - this superb! Thanks for giving `IRange` the time & love it deserves. Much appreciated!

src/AngleSharp/Dom/INode.cs
FlorianRappl: All fine - thanks for putting in the effort. Went over the tests, yes, matches the desired behavior & spec. Well done!

PR #1172, opened 2024-02-23
Use a fully SemVer compatible versioning scheme, fixes #1171

### LINKED ISSUE #1171: SemVer scheme in AngleSharp -alpha versions broken
### Prerequisites

### Description

The AngleSharp prerelease packges use a naming scheme like [`1.1.0-alpha-379`](https://www.nuget.org/packages/AngleSharp/1.1.0-alpha-379) or [`1.0.0-alpha-117`](https://www.nuget.org/packages/AngleSharp.Css/1.0.0-alpha-117). The interesting part about this is that `alpha` and the following number are seperated by a hypen. The [SemVer spec](https://semver.org/#spec-item-11) `§ 11.4` reads (emphasis mine)

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
> 1. Identifiers consisting of only digits are compared numerically.
> 2. ***Identifiers with letters or hyphens are compared lexically in ASCII sort order.***
> 3. Numeric identifiers always have lower precedence than non-numeric identifiers.
> 4. A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.
> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

The effects of applying this rule are easiest to observe on the `AngelSharp.Css` NuGet page. I still decided to create the issue here (on the main AngleSharp repo), as the root cause is shared accross the AngleSharp projects.

image

NuGet.org shows the versions according to their precedence. Thus, the `-99` version is higher than the `-117` version. I think this is not the intended outcome. If the hypen was a dot instead, the ordering would be correct.

### Steps to Reproduce

```bash
dotnet new console
dotnet add AngleSharp.Css --prerelease
```

### Expected Behavior

A message like the following, signalling the newest version of an AngleSharp prerelease package was installed

> info : Die PackageReference für das Paket "AngleSharp.Css", Version 1.0.0-alpha.117, wurde der Datei "D:\git\anglesharptest\anglesharptest.csproj" hinzugefügt.

### Actual Behavior

A message like

> info : Die PackageReference für das Paket "AngleSharp.Css", Version 1.0.0-alpha-99, wurde der Datei "D:\git\anglesharptest\anglesharptest.csproj" hinzugefügt.

This is additionally problematic because NuGet lets you upgrade packages more easily than downgrading them. E.g. [HtmlSanitizer beta](https://www.nuget.org/packages/HtmlSanitizer/8.1.844-beta#dependencies-body-tab) references `AngleSharp.Css >= 1.0.0-alpha-99`. There is no way to upgrade to the newest `AngleSharp.Css` version by taking a direct dependency on  1.0.0-alpha-117, because nuget disallows the "downgrade":

```
error NU1109: Detected package downgrade: AngleSharp.Css from 1.0.0-alpha-99 to centrally defined 1.0.0-alpha-117. Update the centrally managed package version to a higher version.  [/home/vsts/work/1/s/MyProj.sln]
error NU1109:  MyProj -> HtmlSanitizer 8.1.844-beta -> AngleSharp.Css (>= 1.0.0-alpha-99)  [/home/vsts/work/1/s/MyProj.sln]
error NU1109:  MyProj -> AngleSharp.Css (>= 1.0.0-alpha-117) [/home/vsts/work/1/s/MyProj.sln]

```

For future prerelease versions, a naming scheme that takes a higher precedence then all existing prerelease packages could be chosen. An easy way would be to name them `beta` instead of `alpha`, followed by a dot and then a number. I understand that this might not be a viable way though, given the different meaning of `alpha` and `beta`. Another way might be to use `-alphax.N` (I didn't test this) or something alike, or `-prerelease.alpha.N`. I guess that also depends on personal taste.

Let me know if there's anything further I can contribute to resolve this issue.

georg-jung:
# Types of Changes

## Prerequisites

## Contribution Type

- [X] Bug fix (non-breaking change which fixes an issue, please reference the issue id) #1171

## Description

See #1171. Hopefully fixes #1171.

@FlorianRappl I'd be happy to contribute this - be aware though that this is the first time I'm looking at AngleSharp's sources and also the first time I edit a Nuke build. Thus, I wouldn't say with confidence that I got everything about this right.

## Notes 

* `_versionRegex` seems unused and SemVer-incompatible, thus I removed it.
* At a future time `-beta.` could be easily changed to `-alpha.` to be able to release alpha versions again, when targeting some future MAJOR.MINOR.PATCH.
* The CI action [runs fine](https://github.com/georg-jung/AngleSharp/actions/runs/8021008315), `.\build.ps1` and `.\build.ps1 -Target PrePublish` run fine locally too (but don't generate a prerelease package there).
* When merging these changes in my fork's devel branch to test a proper prerelease package is generated [that seems to happen](https://github.com/georg-jung/AngleSharp/actions/runs/8021250031/job/21912891256#step:4:94)
* The publishing itself etc. seem hard to test for me, given they directly publish to NuGet.org but don't emit artifacts on GitHub Actions

### DIFF
--- a/nuke/Build.cs
+++ b/nuke/Build.cs
@@ -90,11 +90,11 @@ class Build : NukeBuild
 
             if (ScheduledTargets.Contains(Default))
             {
-                Version = $"{Version}-ci-{buildNumber}";
+                Version = $"{Version}-ci.{buildNumber}";
             }
             else if (ScheduledTargets.Contains(PrePublish))
             {
-                Version = $"{Version}-alpha-{buildNumber}";
+                Version = $"{Version}-beta.{buildNumber}";
             }
         }
 
--- a/nuke/ReleaseNotesParser.cs
+++ b/nuke/ReleaseNotesParser.cs
@@ -17,16 +17,6 @@ using Microsoft.Build.Exceptions;
 /// </remarks>
 public sealed class ReleaseNotesParser
 {
-    private readonly Regex _versionRegex;
-
-    /// <summary>
-    /// Initializes a new instance of the <see cref="ReleaseNotesParser"/> class.
-    /// </summary>
-    public ReleaseNotesParser()
-    {
-        _versionRegex = new Regex(@"(?<Version>\d+(\s*\.\s*\d+){0,3})(?<Release>-[a-z][0-9a-z-]*)?");
-    }
-
     /// <summary>
     /// Parses all release notes.
     /// </summary>


FlorianRappl:
This is fine and the change seems good. Let's merge in and see what happens (but I am quite confident this will do).

Thanks for your work!

FlorianRappl (approved):

FlorianRappl:
Seems to work as it should:

https://github.com/AngleSharp/AngleSharp/releases/tag/1.1.0-beta.384
https://www.nuget.org/packages/AngleSharp/1.1.0-beta.384

georg-jung:
Great, thanks!

image

One more thing: If I understand correctly how AngleSharps versioning works, it'd now be good to add "something > 1.1.0" to `CHANGELOG.md`, possibly with just a "not finished yet" line. It's also important to manually add at least such an entry after every non-prerelease release for the next upcoming alpha/beta. In the SemVer sense, what we just published ([1.1.0-beta.384](https://github.com/AngleSharp/AngleSharp/releases/tag/1.1.0-beta.384)) would be a beta before (or read: "working towards") [1.1.0](https://www.nuget.org/packages/AngleSharp/1.1.0) - which was already released. Thus, this doesn't really make sense. Even though, it does solve the larger "you can not depend on the most recent version" issue I originally meant to file with #1171.

If one would now do a `dotnet add AngleSharp --prerelease`, I think he'd get the release version not the prerelease.

Edit: I think it still wouldn't be possible to depend on our now-released pre-release by directly taking a dependency on it, if a dependency depends on `1.1.0`. This now isn't an issue with the scheme anymore though, but rather requires "the discipline" to always add the CHANGELOG.md line right after publishing a non-prerelease release.

(I personally tend to use solutions like [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning) to remove the manual steps and also enable consistent versioning for local builds, but that's an opionionated thing I guess and also a bigger change.)

FlorianRappl:
The changelog is adjusted when there is something new. In this case you should have done this in your PR. I did not correct it as it also did not bring any new changes. So this will not be a release in any case.

To quantify a new version there needs to be something new. Without it reasoning about a new version is meaningless (this is not breaking, but also not a new feature and also no hotfix - it is "nothing" releasable).

> (I personally tend to use solutions like [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning) to remove the manual steps and also enable consistent versioning for local builds, but that's an opionionated thing I guess and also a bigger change.)

I dislike such solutions as they force everyone to adhere to a standard when making git commits. Also git commits are technical / with information about the code in nature, but a changelog might / should be written more towards end users.

> If one would now do a dotnet add AngleSharp --prerelease, I think he'd get the release version not the prerelease.

Which is correct. The most recent one is the release; the current prerelease does not bring any new features or anything at all - users should remain at the stable version.

georg-jung:
> I did not correct it as it also did not bring any new changes. So this will not be a release in any case.

> The most recent one is the release; the current prerelease does not bring any new features or anything at all - users should remain at the stable version.

I agree - just wanted to make sure you're aware there always needs to be an entry with a \<higher than the latest realesed version\> version in CHANGELOG.md before a prerelease (that is meant to be consumed by anybody) is pushed to nuget _or otherwise the SemVer of the prerelease will be broken again_.

> I dislike such solutions as they force everyone to adhere to a standard when making git commits. Also git commits are technical / with information about the code in nature, but a changelog might / should be written more towards end users.

NBGV isn't about generating the change log. It generates the package version number in an incrementing determinisitic way. Similar to using the latest version from CHANGELOG + Github Actions build number, as Angle Sharp does, but not bound to Github Actions. It doesn't require any committing standards too. It works actually quite similar to `read the version information from CHANGELOG.md + number of commits since last change to that version`. It thereby makes it easy to generate packages with valid SemVers.

But as I said, I guess that's an opinionated thing and I wasn't necessarily trying to convince you to do something else than you're doing currently - I think AngleSharp is a great library, thanks for the good work! I just wanted to note that there are options that require less manual attention to ensure SemVer compliance, in case you're interested. No offense or pushing intended :)

georg-jung:
Sorry, I guess I wrote the changelog comment in the wrong tone and it sounds a bit rude when I read it again. That wasn't my intention at all. Have a nice weekend and thanks again for creating AngleSharp!

FlorianRappl:
Sure all good - none taken :+1:.

My point was just that such libraries / tools (I was not really referring to the one you mentioned, just to such tools in general, which might not have been a good comparison given my complete lack of knowledge regarding the library you mentioned) don't really know what is actually going on in the code base (e.g., in .NET there can be many things that don't look to be breaking, but are actually breaking changes) - so I would not rely on such a library to correctly identify the next increment correctly, but of course I'm happy if you find it working for you.

Maybe I'll give it a shot and see if could work in this context.

Also thanks for the nice words (and, again, the contribution) 🚀 !

FlorianRappl:
Also @georg-jung  I've just now seen that you've sponsored my work - just wanted to say a HUGE THANK YOU! You've just made my day ❤️ ! Much appreciated 🍻 !

PR #1185, opened 2024-03-20
Fixed the loading of link element source

### LINKED ISSUE #1184: Link element source not loaded when element is appended to document
### Prerequisites

### Description

When setting up the BrowsingContext with the default resource loader and loading an HTML document it loads any `link` elements with stylesheets just fine. But when appending any new `link` elements to the document they do not get loaded as expected.

### Steps to Reproduce

```
var config = Configuration.Default
                .WithCss()
                .WithDefaultLoader(new LoaderOptions() { IsResourceLoadingEnabled = true });

var html = @"<!doctype html><body></body>";

var document = BrowsingContext.New(config).OpenAsync(m => m.Content(html));

var link = document.CreateElement("link");
link.SetAttribute(AttributeNames.Rel, "stylesheet");
link.SetAttribute(AttributeNames.Href, "/mock-stylesheet.css");

document.Body.AppendChild(link);
```

### Expected Behavior

The newly added link element's source should be loaded by the default resource loader (and in this case the CSS stylesheet should be parsed).

### Actual Behavior

The link element is added to the document, but no request is made for the link's href URL.

I have a PR ready for this with a fix.

tomvanenckevort:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

This is a fix for #1184 

The issue was that the `_relation` variable was only set during the `SetupElement` method. But when adding a new link element the `rel` attribute is often not set at this point. So if `rel` and `href` are set after creation it will never call `_relation?.LoadAsync()`, and therefore the resource loader won't kick in.

Changes:
- Updated `HtmlLinkElement` to make sure that it attempts to load the relation when either the `rel`, `href` attributes have changed, or when the parent has changed (i.e. when it's added to the document). I took inspiration from the `HtmlScriptElement` where something similar happens.
- Updated `DefaultAttributeObserver` to trigger the new `UpdateRel` method when the `rel` attribute changes.
- Added a new unit test to prove that newly added links now trigger the right resource loader.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/StyleExtensions.cs
+++ b/src/AngleSharp.Core.Tests/Css/StyleExtensions.cs
@@ -1,6 +1,8 @@
 namespace AngleSharp.Core.Tests.Css
 {
     using System;
+    using System.Collections;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Threading;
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
LGTM - thanks!

PR #1187, opened 2024-04-13
Remove unnecessary System.Text.Encoding.CodePages ref from NET 6+

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

NET 6+ doesn't need explicit reference to System.Text.Encoding.CodePages as it's part of runtime libraries. This trims down dependencies. I've changed minimum from NET 4.6.1 to 4.6.2 as it's the oldest supported FW platform and even Microsoft libaries aren't guaranteed to work with 4.6.1 (like Unsafe).

### DIFF
--- a/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
+++ b/src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net6.0;net7.0;net8.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;net472;net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net462;net472</TargetFrameworks>
     <IsPackable>false</IsPackable>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies> <!-- https://github.com/Tyrrrz/GitHubActionsTestLogger/issues/5 -->
   </PropertyGroup>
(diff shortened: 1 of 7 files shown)


FlorianRappl:
I am not sure this is part of the runtime / correctly handled if its removed. In the past we explicitly added it as not having the dependency resulted in some more exotic codepages not found; this is nothing that can be directly seen - you need to have a dependency on such a codepage to have the one missing.

I would be all for removing, but I first need confirmation that this fully and reliably works when we remove it. Unfortunately, .NET is not really well thought out in that area.

src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net6.0;net7.0;net8.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">net461;net472;net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net462;net472</TargetFrameworks>
```
FlorianRappl, line 4: Also if you remove frameworks please make sure to be consistent;

- https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp.nuspec
- https://github.com/AngleSharp/AngleSharp/blob/devel/README.md

These files (and maybe others) should be modified, too.

lahma: I've updated the places needed 4.6.1->4.6.2 update (hopefully)

lahma:
> I am not sure this is part of the runtime / correctly handled if its removed. In the past we explicitly added it as not having the dependency resulted in some more exotic codepages not found; this is nothing that can be directly seen - you need to have a dependency on such a codepage to have the one missing.
> 
> I would be all for removing, but I first need confirmation that this fully and reliably works when we remove it. Unfortunately, .NET is not really well thought out in that area.

Older platforms will get 6.0.0 (explicit reference), NET 6.0.x will get 6.0.x - NET 7 will get 7.0.x - NET 8 will get 8.0.x - as they are inbox.

FlorianRappl:
Well, that unfortunately does not answer my concern. As mentioned (and please use the search for this) we had the situation beforehand. There have been bug reports and other issues - these all went away when we have the dependency explicitly mentioned.

If its part of the framework / of the existing assemblies then I don't see any problem here; NuGet should not download the dependency.

Just to re-iterate: I would love to remove it, but I need confirmation that no bugs will occur due to that.

lahma:
Yes so far what I've read/understood from searching the issues it's always on full framework where you naturally need to have a package reference with a version minimum of what is supported. The explicit separate nuspec file can make things harder to understand compared to running `dotnet pack -c Release`  which will then validate that the code compiles with given references and produces a nuget package that is valid:

image

FlorianRappl:
The thing with `dotnet pack` is that it unfortunately does not cover the full range of spectrum that a nuspec does. It's another issue in .NET that I would think they should have fixed right away. Why have multiple ways if they are all providing different pros and cons? Just having a csproj and making the NuGet based on this (either via dotnet pack or by having the package created automatically at build) is just much better.

lahma:
I could also investigate if it would be possible to produce identical nuget package via csproj approach (removing the separate nuspec), I think this would make things more discoverable and easier to understand.

lahma:
> I could also investigate if it would be possible to produce identical nuget package via csproj approach (removing the separate nuspec), I think this would make things more discoverable and easier to understand.

Looking at the current nuspec files, I would argue that they could quite easily be replaced with csproj configuration, if wanted.

FlorianRappl (approved):

PR #1188, opened 2024-04-13
Upgrade to NUKE 8

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

NUKE 8 with code tweaks and also deterministic build to get rid of warning in NuGet package: https://nuget.info/packages/AngleSharp.Css/1.0.0-beta.139


build.ps1
```diff
@@ -18,11 +18,10 @@ $TempDirectory = "$PSScriptRoot\\.nuke\temp"
 
 $DotNetGlobalFile = "$PSScriptRoot\\global.json"
 $DotNetInstallUrl = "https://dot.net/v1/dotnet-install.ps1"
-$DotNetChannel = "Current"
+$DotNetChannel = "STS"
```
FlorianRappl, line 21: I think current is fine; otherwise I'd prefer LTS over STS, which would still work for the upgrade / .NET 8.

FlorianRappl (commented):
This looks great - I'd still prefer to use LTS but otherwise this could be merged.

build.ps1
lahma: This was produced by `nuke :update` so it's the update tool using the template defaults. The default name changed a while ago https://github.com/dotnet/core/pull/7517https://github.com/dotnet/core/pull/7517

FlorianRappl: Yeah I see - I mean my comment was rather that current/STS is fine, but in general I prefer LTS.

So all I wanted to remark is that since we touch it now we could also change it to LTS, too. But I am fine if we just make the 1:1 transition here without such modifications.

lahma:
This was a bit more tedious but turned out to be the file path resolving in tests which didn't behave correctly when deterministic build and paths were enabled. Worked around the problem by using more robust content copy to run directory and reading from thre.

FlorianRappl (approved):

PR #1190, opened 2024-04-14
Add back missing NET6+ targets to AngleSharp.nuspec

lahma:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

While there's still ongoing discussion about pros and cons for using csproj instead of nuspec file on #1189 I think it's better to fix what I broke as an intermediate solution. Bringing back explicit dependency targets without the actual dependencies.

### DIFF
--- a/src/AngleSharp.nuspec
+++ b/src/AngleSharp.nuspec
@@ -25,6 +25,12 @@
       <group targetFramework="net472">
         <dependency id="System.Text.Encoding.CodePages" version="8.0.0" />
       </group>
+      <group targetFramework="net60">
+      </group>
+      <group targetFramework="net70">
+      </group>
+      <group targetFramework="net80">
+      </group>
     </dependencies>
   </metadata>
 </package>


lahma:
Another food for thought, NET 7 will be obsolete next month so how long should that be supported as an explicit target.

FlorianRappl (approved):
Thanks for spotting this!

FlorianRappl:
> Another food for thought, NET 7 will be obsolete next month so how long should that be supported as an explicit target.

As long as we can (i.e., as long as it does not "hurt" / limit us) - but same as with .NET 4.6.1 if we ever encounter even the slightest problem we'd remove it.

lahma:
> > Another food for thought, NET 7 will be obsolete next month so how long should that be supported as an explicit target.
> 
> As long as we can (i.e., as long as it does not "hurt" / limit us) - but same as with .NET 4.6.1 if we ever encounter even the slightest problem we'd remove it.

Yep, I think the pain starts to hit if GH Actions runners start to drop obsolete targeting.

PR #1195, opened 2024-07-26
Added property attribute to meta tags for The Open Graph protocols support

hflexgrig:
# Types of Changes

## Prerequisites

## Contribution Type

- [x] I have updated the documentation accordingly

## Description

### DIFF
--- a/src/AngleSharp/Dom/ElementExtensions.cs
+++ b/src/AngleSharp/Dom/ElementExtensions.cs
@@ -1583,6 +1583,26 @@ namespace AngleSharp.Dom
         }
 
         /// <summary>
+        /// Returns the property attribute. Widely used for meta tags for Open Graph protocols
+        /// </summary>
+        /// <param name="element"></param>
+        /// <returns>The value of property attribute</returns>
+        public static String? GetProperty(this IElement element)
+        {
+            return element.GetAttribute("property");
+        }
+
+        /// <summary>
+        /// Sets the property value
+        /// </summary>
+        /// <param name="element"></param>
+        /// <param name="value"></param>
+        public static void SetProperty(this IElement element, String value)
+        {
+            element.SetAttribute("property", value);
+        }
+
+        /// <summary>
         /// Parses the HTML subtree of the given content in the context of
         /// the provided element.
         /// </summary>
--- a/src/AngleSharp/Dom/ElementExtensions.cs
+++ b/src/AngleSharp/Dom/ElementExtensions.cs
@@ -1587,7 +1587,7 @@ namespace AngleSharp.Dom
         /// </summary>
         /// <param name="element"></param>
         /// <returns>The value of property attribute</returns>
-        public static String? GetProperty(this IElement element)
+        public static String? GetProperty(this IHtmlMetaElement element)
         {
             return element.GetAttribute("property");
         }
@@ -1597,7 +1597,7 @@ namespace AngleSharp.Dom
         /// </summary>
         /// <param name="element"></param>
         /// <param name="value"></param>
-        public static void SetProperty(this IElement element, String value)
+        public static void SetProperty(this IHtmlMetaElement element, String value)
         {
             element.SetAttribute("property", value);
         }
--- a/src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
+++ b/src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
@@ -16,6 +16,12 @@ namespace AngleSharp.Html.Dom
         String? Name { get; set; }
 
         /// <summary>
+        /// Gets or sets the name of the meta element.
+        /// </summary>
+        [DomName("property")]
+        String? Property { get; set; }
+
+        /// <summary>
         /// Gets or sets the value of the equivalent in a meta element, which
         /// is effective if the server doesn't send a corresponding real header.
         /// </summary>
--- a/src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
+++ b/src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
@@ -16,12 +16,6 @@ namespace AngleSharp.Html.Dom
         String? Name { get; set; }
 
         /// <summary>
-        /// Gets or sets the name of the meta element.
-        /// </summary>
-        [DomName("property")]
-        String? Property { get; set; }
-
-        /// <summary>
         /// Gets or sets the value of the equivalent in a meta element, which
         /// is effective if the server doesn't send a corresponding real header.
         /// </summary>
--- a/src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
+++ b/src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
@@ -51,6 +51,12 @@ namespace AngleSharp.Html.Dom
             set => this.SetOwnAttribute(AttributeNames.Name, value);
         }
 
+        public String? Property
+        {
+            get => this.GetOwnAttribute(AttributeNames.Property);
+            set => this.SetOwnAttribute(AttributeNames.Property, value);
+        }
+
         #endregion
 
         #region Methods
--- a/src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
+++ b/src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
@@ -51,12 +51,6 @@ namespace AngleSharp.Html.Dom
             set => this.SetOwnAttribute(AttributeNames.Name, value);
         }
 
-        public String? Property
-        {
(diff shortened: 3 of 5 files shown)


src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
```diff
@@ -15,6 +15,12 @@ public interface IHtmlMetaElement : IHtmlElement
         [DomName("name")]
         String? Name { get; set; }
 
+        /// <summary>
+        /// Gets or sets the name of the meta element.
+        /// </summary>
+        [DomName("property")]
```
FlorianRappl: What is this? I've never seen this property in any spec and all browsers tell me that such a property does not exist.

src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
```diff
@@ -51,6 +51,12 @@ public HtmlMetaElement(Document owner, String? prefix = null)
             set => this.SetOwnAttribute(AttributeNames.Name, value);
         }
 
+        public String? Property
```
FlorianRappl: Why have this at all? Why not just use `GetAttribute("property")` from the outside?

FlorianRappl (commented):
I miss the value proposition for this one.

The introduced property `property` is not in any DOM spec that I'm aware of - no browser implements this.

As the property is only a wrapper for the attribute (this is valid, but then we'd need to remove the `DomNameAttribute`) I'd argue that it should be either an extension method or not exist at all.

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
hflexgrig: Open graph protocols for SEO. https://ogp.me/

src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
hflexgrig: For SEO purposes open graph protocols used widely along with Name attribute, thought will simplify job for others. https://github.com/hflexgrig/SeoFriendlySpa/blob/master/SeoFriendlySpa/Services/SeoService.cs 
For my project I had to use Attribute directly, but with Property would be easier.

hflexgrig: If you check any ECommerce site which needs to be SEO friendly, they're widely using them. 
For example <meta property="og:image" content="imagepath" />  used for showing image preview, when you share your url on social sites e.g. Facebook.

FlorianRappl: > If you check any ECommerce site which needs to be SEO friendly, they're widely using them.

Using what? My comment was not about the HTML (there you can use any *attribute*), but about the DOM spec (such a *property* does not exist - so no one can use it).

What og meta tags do I know - that was not my question.

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
FlorianRappl: Sorry you seem to confuse an **attribute** with a **property**.

What you refer to is an attribute that is in particular used for the og meta tags - but only because some meta tags (you can use any - `name` is a free field attribute - there is no constraint on the chosen value) want to denote further values in some dedicated attributes does not mean these attributes have an associated property.

FlorianRappl:
As mentioned you can (if you want to) bring in this *convenience* property. But then I'd also like to see other convenience properties on `meta`, too.

Furthermore, it needs to be compliant with the spec. So remove `DomName`. Such a thing does not exist.

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
hflexgrig: Removed DomName attribute on a latest commit

src/AngleSharp/Html/Dom/Internal/HtmlMetaElement.cs
hflexgrig: > What og meta tags do I know - that was not my question.

I'm not expert in this area, probably I'm confused. I came across AngleSharp just yesterday, when was trying to solve specific problem. And as I'm gonna use `property` tags often, decided to make it easier to use like with `name`.

FlorianRappl: Yeah but please see the README - the core design is to follow the official W3C spec (i.e., to work as in the browser). Therefore, you cannot just invent stuff. You need to follow what is there.

Why didn't you just place an extension method in your code? It would have solved it / make it more convenient. There is no need to change the library in this case.

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
```diff
@@ -15,6 +15,11 @@ public interface IHtmlMetaElement : IHtmlElement
         [DomName("name")]
         String? Name { get; set; }
 
+        /// <summary>
+        /// Gets or sets the name of the meta element.
+        /// </summary>
+        String? Property { get; set; }
```
FlorianRappl: I think this should be provided as an extension method instead.

There are multiple reasons for this:

1. Its breaking - anyone who implements the interface right now would also need to make this change.
2. The implementation would always be the same ("just grab the attribute")
3. It's not an official DOM property (hence the lack of `[DomName]`)

So I think going for an extension method instead would be sound (it's still more convenient, but transports the utility character better without requiring any implementation effort).

hflexgrig: Okay, lats then reject my PR, I'll go with extension method. Thank you for explaining it to me.

FlorianRappl: OK I meant an extension method in AngleSharp - but if you prefer an extension method in your code - feel free to do so.

hflexgrig: > OK I meant an extension method in AngleSharp - but if you prefer an extension method in your code - feel free to do so.

Oh, sorry got you wrong, I'll add extension method in AngleSharp then. Only could you please mention where exactly to add extension method ?

FlorianRappl: I am not sure if we have already a file for that. Either in the file `ElementExtensions` or create a new file like `HtmlMetaElementExtension` (I think the `ElementExtensions` should do).

src/AngleSharp/Dom/ElementExtensions.cs
```diff
@@ -1582,6 +1582,26 @@ public static String GetSelector(this IElement element)
             return path;
         }
 
+        /// <summary>
+        /// Returns the property attribute. Widely used for meta tags for Open Graph protocols
+        /// </summary>
+        /// <param name="element"></param>
+        /// <returns>The value of property attribute</returns>
+        public static String? GetProperty(this IElement element)
```
FlorianRappl: I would constrain this to `IHtmlMetaElement` as its usage is unclear otherwise (and if I'd see a general "Get Property" on a standard Element I'd ask "what property").

src/AngleSharp/Dom/ElementExtensions.cs
```diff
@@ -1582,6 +1582,26 @@ public static String GetSelector(this IElement element)
...
+        {
+            return element.GetAttribute("property");
+        }
+
+        /// <summary>
+        /// Sets the property value
+        /// </summary>
+        /// <param name="element"></param>
+        /// <param name="value"></param>
+        public static void SetProperty(this IElement element, String value)
```
FlorianRappl: I would constrain this to `IHtmlMetaElement` as its usage is unclear otherwise (and if I'd see a general "Get Property" on a standard Element I'd ask "what property").

FlorianRappl (commented):
I think this is good - maybe change the target to be an `IHtmlMetaElement`.

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
hflexgrig: Btw there's [discussions](https://github.com/dotnet/csharplang/discussions/5811) of introducing extension properties. Would be a nice solution on this case :)

FlorianRappl (approved):

src/AngleSharp/Html/Dom/IHtmlMetaElement.cs
FlorianRappl: I know - these discussions, however, are ongoing for the last 10 years (I'm a Microsoft MVP - originally coming from the languages area; so I was constantly in exchange with the product group). Generally, this will eventually be solved with extensions, which are much more generic and have a couple of advantages.

PR #1203, opened 2024-09-23
New setting for disabling Stringbuilder pooling.

### LINKED ISSUE #1201: StringBuilderPool performance improvements: Turn off locking
### Description

The locking mechanism that is used in the StringBuilderPool can cause a high amount of waits with systems that use a lot of threads to process multiple documents concurrently.  This has been a finding during some profiling on a heavily used process.  

Have a new Property/Setting on the StringBuilderPool class to optionally turn off pooling which will effectively skip the locking call (https://github.com/AngleSharp/AngleSharp/blob/6c4be442576f7524c8e3bb99540323537c7de2dd/src/AngleSharp/Text/StringBuilderPool.cs#L41).   
- Should stay off by default to preserve current behavior.
- If turned off, this could increase memory cost since pooling won't be used.

### Background

_No response_

### Specification

_No response_

micahbresette:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Added a new setting for disabling Stringbuilder pooling.  This can increase performance during heavy parallel processing as it will skip the lock.  However, by disabling pooling it can also increase overall memory consumption.

https://github.com/AngleSharp/AngleSharp/issues/1201

### DIFF
--- a/src/AngleSharp/Text/StringBuilderPool.cs
+++ b/src/AngleSharp/Text/StringBuilderPool.cs
@@ -14,7 +14,7 @@ namespace AngleSharp.Text
         private static Int32 _count = 4;
         private static Int32 _limit = 85000;
         private static bool _isPoolingDisabled = false;
-        private static Int32 _defaultStringBuilderSize = 1024;
+        private static readonly Int32 _defaultStringBuilderSize = 1024;
 
         /// <summary>
         /// Gets or sets whether string builder pooling is disabled.  When disabled, Obtain() will always return a new instance.
--- a/src/AngleSharp/Text/StringBuilderPool.cs
+++ b/src/AngleSharp/Text/StringBuilderPool.cs
@@ -51,7 +51,7 @@ namespace AngleSharp.Text
         /// <returns>A stringbuilder to use.</returns>
         public static StringBuilder Obtain()
         {
-            StringBuilder result;
+            StringBuilder result = null;
 
             if (_isPoolingDisabled)
             {
--- a/src/AngleSharp/Text/StringBuilderPool.cs
+++ b/src/AngleSharp/Text/StringBuilderPool.cs
@@ -51,7 +51,7 @@ namespace AngleSharp.Text
         /// <returns>A stringbuilder to use.</returns>
         public static StringBuilder Obtain()
         {
-            StringBuilder result = null;
+            StringBuilder result;
 
             if (_isPoolingDisabled)
             {
--- a/src/AngleSharp/Text/StringBuilderPool.cs
+++ b/src/AngleSharp/Text/StringBuilderPool.cs
@@ -13,15 +13,15 @@ namespace AngleSharp.Text
         private static readonly Object _lock = new();
         private static Int32 _count = 4;
         private static Int32 _limit = 85000;
-        private static bool _isPoolingDisabled = false;
-        private static readonly Int32 _defaultStringBuilderSize = 1024;
+        private static Boolean _isPoolingDisabled = false;
+        private const Int32 _defaultStringBuilderSize = 1024;
 
         /// <summary>
         /// Gets or sets whether string builder pooling is disabled.  When disabled, Obtain() will always return a new instance.
         /// Disabling will increase memory usage and GC pressure, but may improve performance in scenarios with high
         /// parallel processing.
         /// </summary>
-        public static bool DisablePooling
+        public static Boolean IsPoolingDisabled
         {
             get => _isPoolingDisabled;
             set => _isPoolingDisabled = value;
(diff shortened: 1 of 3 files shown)


src/AngleSharp/Text/StringBuilderPool.cs
```diff
@@ -13,6 +13,19 @@ public static class StringBuilderPool
...
         private static Int32 _limit = 85000;
+        private static bool _isPoolingDisabled = false;
+        private static readonly Int32 _defaultStringBuilderSize = 1024;
+
+        /// <summary>
+        /// Gets or sets whether string builder pooling is disabled.  When disabled, Obtain() will always return a new instance.
+        /// Disabling will increase memory usage and GC pressure, but may improve performance in scenarios with high
+        /// parallel processing.
+        /// </summary>
+        public static bool DisablePooling
```
FlorianRappl: Why not call it `IsPoolingDisabled`? Usually the `Is...` convention is used for boolean properties.

Plus we use the real type name and not the keyword / alias in the code base. (`bool` -> `Boolean`)

src/AngleSharp/Text/StringBuilderPool.cs
```diff
@@ -38,14 +51,21 @@ public static Int32 SizeLimit
         /// <returns>A stringbuilder to use.</returns>
         public static StringBuilder Obtain()
```
FlorianRappl, line 52: Please put the original method lock body in an inlined method (e.g., `ObtainUnlocked`) -> then you can either call it from a `lock` or without the `lock` (depending on the condition).

FlorianRappl: There is also a `_lock` in `ReturnToPool`, which should be taken (same strategy there). Otherwise, it's not consistent.

src/AngleSharp/Text/StringBuilderPool.cs
```diff
@@ -13,6 +13,19 @@ public static class StringBuilderPool
         private static readonly Object _lock = new();
         private static Int32 _count = 4;
         private static Int32 _limit = 85000;
+        private static bool _isPoolingDisabled = false;
+        private static readonly Int32 _defaultStringBuilderSize = 1024;
```
FlorianRappl: Why is this a static readonly? Make it a const (it's not exposed and cannot be changed anyway -> a classic const).

src/AngleSharp/Text/StringBuilderPool.cs
micahbresette: Thanks for pointing out the ReturnToPool, I was thinking about that one the earlier today.  As far as wrapping the original implementation in a separate method I don't think it will add any benefit, as we end up making it more complex by adding the additional checks and enabling the ability to mess up the logic by calling the internal method incorrectly (not in a lock, but saying get locked version).   There is not much code duplication (minus the creation of the actual string builder).  Maybe I am missing something here though.

src/AngleSharp/Text/StringBuilderPool.cs
micahbresette: No reason it cannot be a const

src/AngleSharp/Text/StringBuilderPool.cs
micahbresette: Updated

src/AngleSharp/Text/StringBuilderPool.cs
micahbresette: I will mark as resolved for now.  @FlorianRappl please let me know your thoughts on the comment above and if its a blocker.

FlorianRappl: Yeah we don't want to have such duplicates in there. These blocks should be 100% the same - so duplicating them does not feel right.

micahbresette: So just to make sure I am understanding your ask correctly here.  You want an anonymous method that encapsulates the call to: 

```csharp
return new StringBuilder(_defaultStringBuilderSize);
```

micahbresette: Done, let me know if you have any other suggestions @FlorianRappl

FlorianRappl: Well, good that you ask - actually, you are right. I was under the wrong impression that both blocks have the same code. But they surely do not (we don't need to check if there is a builder pooled, as we assume that this was set from the beginning). So yes, the logic in both blocks (lock and non-lock) is different - especially in the return to pool.

Yeah let's keep it as is. Thanks for asking!

FlorianRappl (changes requested):
Don't delete files - especially files that are used as test case inputs.

micahbresette:
> Don't delete files - especially files that are used as test case inputs.

@FlorianRappl That file is flagged as virus via windows defender.  My system deleted it.  

Added it back to the sln

FlorianRappl (approved):
LGTM - thanks!

PR #1206, opened 2024-10-15
Add support for targetting iframes

### LINKED ISSUE #1205: Anchor element targeting iframe does not update iframe on click
### Prerequisites

### Description

When attempting to navigate an anchor element using the `DoClick` method (this might also apply to form submit buttons as well), which has a target referring to a named iframe, the document is not loaded in the iframe, but instead as a child context of the context of the document containing the link, opening the page in effectively a new window.

### Steps to Reproduce

```csharp
using AngleSharp;
using AngleSharp.Html.Dom;
using System.Diagnostics;


var context = BrowsingContext.New(Configuration.Default.WithDefaultLoader(new() { IsResourceLoadingEnabled= true }));

await context.OpenAsync(req => req.Content("""
    <html>
    <body>
        <a id="frame" target="target" href="https://github.com/AngleSharp/AngleSharp/">AngleSharp</a>
        <iframe id="target" name="target">
        </iframe>
    </body>
    </html>
"""));


var linkElement = context.Active.GetElementById("frame") as IHtmlAnchorElement;

linkElement.DoClick();

Thread.Sleep(1000);

var frameElement = context.Active.GetElementById("target") as IHtmlInlineFrameElement;


Debug.Assert(frameElement.ContentDocument?.Title.StartsWith("GitHub - AngleSharp") ?? false, "frame document title does not start with AngleSharp");

var targetContext = context.FindChild("target");
Debug.Assert(targetContext.Active?.Title.StartsWith("GitHub - AngleSharp") ?? false, "child context title does not start with AngleSharp");
```

### Expected Behavior

The page with the url located in the anchor element should be loaded as the content document of the iframe.

### Actual Behavior

The page with the url located in the anchor element is loaded in a child context which is not the content document of the iframe.

One possible work around would be to, instead of calling `DoClick`, to instead check if the anchor tag has the target attribute set, and check if the document, or any documents of parent contexts, up to the root context, has any frames with the name identical to the target, and manually navigate that context to the url specific by the anchor tag, while also checking that no other subscribed event handler is cancelling the click.

A possible solution would be, when the frame is creating a new child context, if the frame has a name, it uses that name as the name of the child context, such that resolving targets pick the frame's context, rather than creating a new window.  For this to work correctly, however, it may be necessary to check the parent browsing context for the associated frame browsing context, as in testing with FireFox, an anchor tag in one frame is able to target another frame in the parent page by name.  I don't know how far up or down is necessary, as I've not tested beyond this.

TheJayMann:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Currently, a named iframe creates an unnamed child browsing context, and a link or form which targets the iframe will instead create a separate named child browsing context not associated with the iframe.  Also, the `_top` target should target the "window" browsing context containing the current context, and "_parent" should not target the parent context if the current browsing context is a "window" context.

This pull request is a breaking change as it will change how child browsing context are created, as well as changing most current behavior when clicking a link or button which has a target defined.

Fixes #1205

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,3 +20,4 @@ csharp_prefer_braces = true:suggestion
 csharp_using_directive_placement = inside_namespace:suggestion
 csharp_style_var_for_built_in_types=true:suggestion
 csharp_style_var_when_type_is_apparent=true:suggestion
+csharp_new_line_before_open_brace = all
--- a/src/AngleSharp/BrowsingContext.cs
+++ b/src/AngleSharp/BrowsingContext.cs
@@ -218,6 +218,18 @@ namespace AngleSharp
                 reference.TryGetTarget(out context);
             }
 
+            if (context is null && Active is Document active)
+            {
+                foreach (var childContext in active.GetAttachedReferences<IBrowsingContext>())
+                {
+                    context = childContext.FindChild(name);
+                    if (context is not null)
+                    {
+                        break;
+                    }
+                }
+            }
+
             return context;
         }
 
--- a/src/AngleSharp/BrowsingContext.cs
+++ b/src/AngleSharp/BrowsingContext.cs
@@ -105,7 +105,7 @@ namespace AngleSharp
         /// a window. Useful for properly determining the proper target
         /// for `_top` and `_parent`.
         ///</summary>
-        public Boolean IsFrame => _isFrameContext;
+        internal Boolean IsFrame => _isFrameContext;
 
         /// <summary>
         /// Gets the session history of the given browsing context, if any.
@@ -190,9 +190,20 @@ namespace AngleSharp
         /// </summary>
         /// <param name="name">The name of the child context, if any.</param>
         /// <param name="security">The security flags to apply.</param>
+        /// <returns></returns>
+        public IBrowsingContext CreateChild(String? name, Sandboxes security)
+        {
+            return CreateChild(name, security, false);
+        }
+
+        /// <summary>
+        /// Creates a new named browsing context as child of the given parent.
+        /// </summary>
+        /// <param name="name">The name of the child context, if any.</param>
+        /// <param name="security">The security flags to apply.</param>
         /// <param name="isFrameContext">Whether the child context is for a frame.</param>
         /// <returns></returns>
-        public IBrowsingContext CreateChild(String? name, Sandboxes security, Boolean isFrameContext = false)
+        internal IBrowsingContext CreateChild(String? name, Sandboxes security, Boolean isFrameContext)
         {
             var context = new BrowsingContext(this, security, isFrameContext);
 
--- a/src/AngleSharp/BrowsingContext.cs
+++ b/src/AngleSharp/BrowsingContext.cs
@@ -211,6 +211,22 @@ namespace AngleSharp
         /// <returns>The found instance, if any.</returns>
         public IBrowsingContext? FindChild(String name)
         {
+            var excludedChild = default(IBrowsingContext);
+            var currentContext = this;
+            var foundChildContext = default(IBrowsingContext);
+            while (foundChildContext is null && currentContext is not null)
+            {
+                foundChildContext = currentContext.FindChildRecursive(name, excludedChild);
+                excludedChild = currentContext;
+                currentContext = currentContext.Parent as BrowsingContext;
+
+            }
+
+            return foundChildContext;
+        }
+
+        private IBrowsingContext? FindChildRecursive(String name, IBrowsingContext? excludedContext)
+        {
             var context = default(IBrowsingContext);
 
             if (!String.IsNullOrEmpty(name) && _children.TryGetValue(name, out var reference))
@@ -220,9 +236,13 @@ namespace AngleSharp
 
             if (context is null && Active is Document active)
             {
-                foreach (var childContext in active.GetAttachedReferences<IBrowsingContext>())
+                foreach (var childContext in active.GetAttachedReferences<BrowsingContext>())
                 {
-                    context = childContext.FindChild(name);
+                    if (childContext.Equals(excludedContext))
+                    {
+                        continue;
+                    }
+                    context = childContext.FindChildRecursive(name, null);
                     if (context is not null)
                     {
                         break;
--- a/src/AngleSharp/BrowsingContext.cs
+++ b/src/AngleSharp/BrowsingContext.cs
@@ -18,6 +18,7 @@ namespace AngleSharp
         private readonly Sandboxes _security;
         private readonly IBrowsingContext? _parent;
         private readonly IDocument? _creator;
+        private readonly Boolean _isFrameContext;
         private readonly IHistory? _history;
         private readonly Dictionary<String, WeakReference<IBrowsingContext>> _children;
 
@@ -54,11 +55,12 @@ namespace AngleSharp
             _history = GetService<IHistory>();
         }
 
-        internal BrowsingContext(IBrowsingContext parent, Sandboxes security)
+        internal BrowsingContext(IBrowsingContext parent, Sandboxes security, Boolean isFrameContext)
             : this(parent.OriginalServices, security)
         {
             _parent = parent;
             _creator = _parent.Active;
+            _isFrameContext = isFrameContext;
         }
 
         #endregion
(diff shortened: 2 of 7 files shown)


TheJayMann:
Currently I have only prepared tests for some cases that I've tested in LibreWolf (Firefox clone).  I personally will not have any time to invest in implementing a fix for this, at least for the next couple of weeks.  Should I find time later, and nobody else has picked this up, I can take over this fix again.

.editorconfig
```diff
@@ -20,3 +20,4 @@ csharp_prefer_braces = true:suggestion
 csharp_using_directive_placement = inside_namespace:suggestion
 csharp_style_var_for_built_in_types=true:suggestion
 csharp_style_var_when_type_is_apparent=true:suggestion
+csharp_new_line_before_open_brace = all
```
FlorianRappl, line 23: I like this one - thanks!

FlorianRappl (commented):
Sure - the test cases look good to me.

My suggestion would be to change the target to a new / custom branch - then we can already include it in the repository. As of right now we cannot take it into `devel` as its the preview branch (demanding finished features / working tests).

.editorconfig
TheJayMann: This was actually necessary for me to do, as my default editor settings are to mostly never include a new line before open brace, and it was messing up the existing code base.

TheJayMann:
Is the suggestion to rebase this branch to main?  If so, I should be able to easily do that.  Otherwise, I can revisit this later.

FlorianRappl:
> Is the suggestion to rebase this branch to main? If so, I should be able to easily do that. Otherwise, I can revisit this later.

No - to retarget to a custom branch (e.g., `feature/iframe-target`). `main` is reserved for stable releases only; `devel` is the feature aggregation branch for the next release.

FlorianRappl:
I changed the base - so we could park the feature / have it open for others to take over from the branch.

TheJayMann:
With the new update, I now have `_top` and `_parent` working properly.  However, it does not appear that an iframe's `CurrentDocument` is being updated when the associated `BrowsingContext` is navigated directly.  I assume this has to do with the `CurrentDocument` making use of a `FrameRequestProcessor`, and, perhaps, the built in url navigation is not making use of this.  Perhaps I can find a way to make this work properly later.

FlorianRappl:
Any news on this? We can take it for the 1.3.0 release.

TheJayMann:
It's definitely not done yet.  I've also not had too much free time to work on this.  I'm working on some tests for dealing with multiple iframes with the same name.  Also, this will likely require some changes to the implementation of child browsing contexts, as unnamed iframes need to be accessible in case it contains a nested named iframe.  Order also appears important, as, in the case of multiple iframes with the same name, an iframe encountered earlier takes priority over an iframe found later.

If you are looking to release 1.3 fairly soon, I'm not sure this will be ready to make that.

FlorianRappl:
Thanks for the fast response @TheJayMann 

> If you are looking to release 1.3 fairly soon, I'm not sure this will be ready to make that.

No rush here, but if there is something I can do in the meantime just write it here. I try to allocate some time for it / move it forward.

TheJayMann:
At this point, I believe I have an understanding of the algorithm used to determine the priority of targets.  When an element targets a link, it first looks at the current window for any iframes with the same name (i.e. looks at all current child windows of the current window), in order based on when the child window was added to the current window, not the order of the child windows within the current window (i.e. `insertBefore` does not give an iframe precedence).  If none is found, then it recursively does the same process, but for each child window (named or unnamed) in the current window.  If no window has been found, then it will perform the same search on the parent window.  If none is found after searching the top most window, it will then search other top most windows in the same context as the current top most windows, I assume in order the windows were created.  Then, if still none have been found, a new window is created.

TheJayMann:
At this point, I believe I have a mostly working implementation.  All tests pass except for those which need to start searching parent browsing contexts.  I need to think of a way to do this without causing reentrancy which would lead to either an infinite loop or a stack overflow.

Also, the method I used to get all of the child browsing contexts regardless of whether they are named or anonymous is by getting all attached references of type `IBrowsingContext` from the active document.  Not sure if that is okay or not, but it works, and I failed to find any other way that wouldn't require significant modification.

TheJayMann:
I have an implementation that works with all of the tests.  I do believe there are some edge cases not covered correctly, but I believe them to be rare, and I really don't want to make even more complicated tests for such cases yet.

I did notice that one completely unrelated test started failing recently, which had not failed prior, `SettingOneExpiredCookieAndAFutureCookieInRequestDoAppearInDocument`.  Looking at the test revealed why.  The cookie with an expiration date in the future just so happened to expire approximately four hours ago (Tue, 28 Jan 2025 13:37:00 GMT), thus, while the test expects it to not yet be expired, it in fact is expired.  Just so happened that this specific time passed as I was finishing the implementation.

FlorianRappl:
> I did notice that one completely unrelated test started failing recently, which had not failed prior, SettingOneExpiredCookieAndAFutureCookieInRequestDoAppearInDocument. Looking at the test revealed why. The cookie with an expiration date in the future just so happened to expire approximately four hours ago (Tue, 28 Jan 2025 13:37:00 GMT), thus, while the test expects it to not yet be expired, it in fact is expired. Just so happened that this specific time passed as I was finishing the implementation.

Yeah this is unfortunate - but unrelated and not a problem with the PR. Historically we use a third-party API for many of the HTTP functionality (together with a page that we own). Ideally, we'd only use a page that we own - such that sudden test failures due to changes in the third-party website behavior are not possible.

> I have an implementation that works with all of the tests. I do believe there are some edge cases not covered correctly, but I believe them to be rare, and I really don't want to make even more complicated tests for such cases yet.

Awesome! I'd say let's go ahead and merge it. Then we can refine it / release it.

src/AngleSharp/BrowsingContext.cs
```diff
@@ -181,10 +190,11 @@ public IEnumerable<T> GetServices<T>() where T : class
         /// </summary>
         /// <param name="name">The name of the child context, if any.</param>
         /// <param name="security">The security flags to apply.</param>
+        /// <param name="isFrameContext">Whether the child context is for a frame.</param>
         /// <returns></returns>
-        public IBrowsingContext CreateChild(String? name, Sandboxes security)
+        public IBrowsingContext CreateChild(String? name, Sandboxes security, Boolean isFrameContext = false)
```
FlorianRappl: We must add another method here (I know that in "practice" / i.e., when re-compiling the optional arg is not changing behavior, but when libs are just swapped that would be a problem). As its not a breaking change / breaking release we should remain ABI compatible.

Just add another method / overload and its fine!

src/AngleSharp/IBrowsingContext.cs
```diff
@@ -70,8 +77,9 @@ public interface IBrowsingContext : IEventTarget, IDisposable
         /// </summary>
         /// <param name="name">The name of the new context.</param>
         /// <param name="security">The sandboxing flag to use.</param>
+        /// <param name="isFrameContext">Whether the child context is for a frame.</param>
         /// <returns>The created browsing context.</returns>
-        IBrowsingContext CreateChild(String? name, Sandboxes security);
+        IBrowsingContext CreateChild(String? name, Sandboxes security, Boolean isFrameContext = false);
```
FlorianRappl: Yeah now it gets dicey. Should have looked at this before the other comment...

If we need to change public interfaces here we are breaking. Is it possible for this change to remain internal (i.e., have the overload with the boolean be `internal` and case to `BrowsingContext`)?

Otherwise we would need to have a 2.0 release for it.

FlorianRappl (commented):
In general looks great, but we will need to refine it a bit to be non-breaking.

src/AngleSharp/IBrowsingContext.cs
TheJayMann: These specific changes are to make `_top` and `_parent` work correctly, as it is important if the browser context is related to a frame or not, and this was the simplest way I could think to do it.  I see how changing an interface can cause problems.  I'd suggest perhaps a default implementation to work around this, but that won't work with `netstandard2.0`, `net462`, and `net472`.
Is there some other way to determine whether a browsing context was created by a frame as opposed to being created as a new window/tab?  If not, I can make these flags internal.  I do already have dependencies on internal code on both `BrowsingContext` and `Document`, so this is probably fine.

FlorianRappl: Well what we can do is to have an internal SubBrowsingContext class, which inherits from BrowsingContext and comes with the relevant references.

TheJayMann: I believe creating a new subclass to contain new functionality would require more effort than benefit can be gained.  After spending a couple hours getting helix and omnisharp to cooperate again, I made the changes necessary to revert the interface back to its original form, using the class directly for the new functionality.

FlorianRappl (approved):
LGTM - thanks!

PR #1211, opened 2024-12-21
Issue 1210 isstepmismatch validation

### LINKED ISSUE #1210: IsStepMismatch validation fails when using floating point numbers (e.g., Lat/Long)
### Prerequisites

### Description

Validity checks for step value mismatch returns false positive. 

For example, defining validation attributes on an HTML number input for geographic coordinates (latitude/longitude) to six places.

```html
<input type="number" max="-116.915989" min="-124.763068" step=".000001" value="-117" />
```
The value (-117) above should be and is valid in most browsers, however, it is reported as being invalid in `AngleSharp (v1.1.2)`.

Reviewing the source code, it appears the issue is in the `AngleSharp.Html.InputTypes.BaseInputType` class `IsStepMismatch()` method [permalink](https://github.com/AngleSharp/AngleSharp/blob/5dc976bb0c4b30799e28a0ea43eff180f4259964/src/AngleSharp/Html/InputTypes/BaseInputType.cs#L147).

```csharp
        /// <summary>
        /// Checks if the current value does not match the steps.
        /// </summary>
        protected Boolean IsStepMismatch()
        {
            var step = GetStep();
            var value = ConvertToNumber(_input.Value);
            var offset = GetStepBase();
            return step != 0.0 && (value - offset) % step != 0.0;    // <-- this expression does not account for rounding errors with floating point numbers
        }
```

### Steps to Reproduce

The following test demonstrates the failed assertion:

```csharp
[Test]
public void TestInputStepMismatchWithFloatingNumber()
{
    var context = BrowsingContext.New(Configuration.Default);
    var htmlParser = context.GetService<IHtmlParser>()!;
    var document = htmlParser.ParseDocument("");

    var element = document.CreateElement("input") as IHtmlInputElement;
    Assert.IsNotNull(element);
    element.Type = "number";
    element.RemoveAttribute("required");
    element.RemoveAttribute("pattern");
    element.RemoveAttribute("step");
    element.RemoveAttribute("max");
    element.RemoveAttribute("min");
    element.RemoveAttribute("maxlength");
    element.RemoveAttribute("value");
    element.RemoveAttribute("multiple");
    element.RemoveAttribute("checked");
    element.RemoveAttribute("selected");

    element.SetAttribute("min", "-124.763068");
    element.SetAttribute("step", ".000001");
    element.Value = "-117";

    Assert.AreEqual("number", element.Type);
    Assert.AreEqual(false, element.Validity.IsStepMismatch);
}
```

### Expected Behavior

The expected behavior is that `element.Validity.IsStepMismatch` is `false`

### Actual Behavior

The expected behavior is that `element.Validity.IsStepMismatch` is `true`, resulting in a false validation error.

A possible fix is to replace the return expression with an expression more friendly to floating points.

Instead of:
```csharp
return step != 0.0 && (value - offset) % step != 0.0;
```
use:
```csharp
return step != 0.0 && Math.Floor((value - offset) / step) * step) != 0.0;
```

rhuebner76:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Fix issue #1210 IsStepMismatch validation not valid with floating point numbers. Does not match expected documented behavior or that of browsers.

Modified `BasicTypeInput` `IsStepMismatch()` method to handle floating-point numbers more appropriately. Added two tests to prove assertions of both valid and invalid floating-point numbers are made as expected. In addition to automated testing, verified tests match validity state of same elements in chromium browser.

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/ValidityStepMismatch.cs
+++ b/src/AngleSharp.Core.Tests/Html/ValidityStepMismatch.cs
@@ -673,7 +673,7 @@ namespace AngleSharp.Core.Tests.Html
             element.RemoveAttribute("selected");
             
             element.SetAttribute("min", "-124.763068");
-            element.SetAttribute("step", ".000007");
+            element.SetAttribute("step", "0.000007");
             element.Value = "-117.000012"; // valid
 
             Assert.AreEqual("number", element.Type);
@@ -700,7 +700,7 @@ namespace AngleSharp.Core.Tests.Html
             element.RemoveAttribute("selected");
 
             element.SetAttribute("min", "124.763");
-            element.SetAttribute("step", ".002");
+            element.SetAttribute("step", "0.002");
             element.Value = "124.764";  // invalid
 
             Assert.AreEqual("number", element.Type);
--- a/src/AngleSharp.Core.Tests/Html/ValidityStepMismatch.cs
+++ b/src/AngleSharp.Core.Tests/Html/ValidityStepMismatch.cs
@@ -395,7 +395,7 @@ namespace AngleSharp.Core.Tests.Html
             element.RemoveAttribute("multiple");
             element.RemoveAttribute("checked");
             element.RemoveAttribute("selected");
-            element.SetAttribute("step", "");   // todo: default to 1
+            element.SetAttribute("step", "");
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM - thanks!

PR #1213, opened 2024-12-31
Fix 1212

### LINKED ISSUE #1212: Cookie 'Expires' Header Recognized as Local Time Instead of GMT in .NET 9
### Prerequisites

### Description

I am not fluent in English as I am Japanese, so I am using AI assistance to create this issue. Please forgive any mistakes in grammar or word choice. Feel free to ask for clarification if anything is unclear. Thank you for your understanding.
This problem occurs in .NET 9. NET 8, the problem does not occur in .NET 8, it is caused by changing the TargetFramework to net9.
When upgrading a project using AngleSharp from .NET 8 to .NET 9, an issue arises where cookies with an Expires attribute are misinterpreted. Specifically, the expiration date specified in GMT is incorrectly parsed as local time, leading to immediate expiration of cookies that should still be valid.

## Description

This issue occurs in the JST (GMT+9) time zone, based on the following example:

1. Access a URL that responds with the following `Set-Cookie` header:
`Set-Cookie: name=value; Expires=Mon, 30 Dec 2024 00:30:00 GMT`
Assume the current time is `Mon, 30 Dec 2024 00:30:00 GMT`, and the cookie is set to expire in 30 minutes.

2. Immediately send a request to another URL. Since the cookie has not yet expired, the Cookie header should be included in the request.

However, the `Expires` value (`Mon, 30 Dec 2024 00:30:00 GMT`) is mistakenly interpreted as local time (in this case, JST).

* The current time in JST is `Mon, 30 Dec 2024 09:00:00`
* Comparing these timestamps: `30 Dec 2024 00:30:00` < `30 Dec 2024 09:00:00`, the cookie is treated as expired.
* As a result, the `Cookie` header is not included in the request.

## Impact		
		
Due to this issue, in the JST time zone, any cookie with an expiration time of less than 9 hours is never used in requests.		
		
## Evidence		

Below is a confirmation of the issue based on packet capture:

### Cases of expiry
image

### Cases that do not expire
image

### Steps to Reproduce

```.cs
var angleConfig = Configuration.Default.WithDefaultLoader().WithDefaultCookies();
using var context = BrowsingContext.New(angleConfig);
_ = await context.OpenAsync($"http://192.168.163.1:2000/Home/Index2"); // URL to which the set-cookie is responded to.
_ = await context.OpenAsync($"http://192.168.163.1:2000/Home/Last"); // URL where the cookie set above is used for the request
```

### Expected Behavior

The `Expires` attribute in the `Set-Cookie` header should always be interpreted as a GMT timestamp, as specified by the HTTP standard.

* If the current time is before the `Expires` time, the cookie should be considered valid and included in the `Cookie` header of subsequent requests.
* This behavior should remain consistent across all time zones, including JST (GMT+9).

### Actual Behavior

The `Expires` attribute in the `Set-Cookie` header is incorrectly interpreted as local time (e.g., JST instead of GMT).

* In cases where the local time is already past the `Expires` time (even though the GMT time is still valid), the cookie is treated as expired.
* As a result, the `Cookie` header is not included in subsequent requests, even when the cookie is still valid according to GMT.

I plan to create a pull request to address this issue. However, this will be my first contribution to an open-source project, so it may take some time to complete. Thank you for your patience and understanding.

https://github.com/AngleSharp/AngleSharp/blob/5dc976bb0c4b30799e28a0ea43eff180f4259964/src/AngleSharp/Io/MemoryCookieProvider.cs#L84

In the string manipulation logic within the source code linked above, the "GMT" designation is being removed from the cookie string. If "GMT" is missing, it appears that in .NET 8, the timestamp was still interpreted as GMT. However, in .NET 9, the timestamp is interpreted as local time instead.

I believe this issue can be resolved by modifying the source code to retain the "GMT" designation, as shown below:

```.cs
var time = utc.ToUniversalTime().ToString("ddd, dd MMM yyyy HH:mm:ss 'GMT'", CultureInfo.InvariantCulture);
```

noobow34:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

I have made the fixes as described in #1212. During testing, I added .NET 9.0 to the TargetFramework, but this change is not included in the commit.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
+++ b/src/AngleSharp.Core.Tests/Library/CookieHandling.cs
@@ -81,6 +81,16 @@ namespace AngleSharp.Core.Tests.Library
         }
 
         [Test]
+        public async Task CookieExpiredInGmtInterpretedAsLocaltime()
+        {
+            //This test must be executed in an environment where the local time is ahead of GMT.
+            var expires = DateTime.Now.AddMinutes(10).ToUniversalTime();
+            var document = await LoadDocumentAloneWithCookie("");
+            document.Cookie = $"ppkcookie2=yet yet another test; expires={expires.ToString("ddd, dd MMM yyyy HH:mm:ss", CultureInfo.InvariantCulture)} GMT; path=/";
+            Assert.AreEqual("ppkcookie2=yet yet another test", document.Cookie);
+        }
+
+        [Test]
         public async Task SettingTwoSimpleCookiesInRequestAppearsInDocument()
         {
             var cookie = await LoadDocumentWithCookie("UserID=Foo, Auth=bar");
--- a/src/AngleSharp/Io/MemoryCookieProvider.cs
+++ b/src/AngleSharp/Io/MemoryCookieProvider.cs
@@ -81,7 +81,7 @@ namespace AngleSharp.Io
 
                     if (DateTime.TryParse(middle.Replace("UTC", "GMT"), out var utc))
                     {
-                        var time = utc.ToUniversalTime().ToString("ddd, dd MMM yyyy HH:mm:ss", CultureInfo.InvariantCulture);
+                        var time = utc.ToUniversalTime().ToString("ddd, dd MMM yyyy HH:mm:ss 'GMT'", CultureInfo.InvariantCulture);
                         cookie = $"{front}{time}{back}";
                     }
 


noobow34:
In Japan, the new year is just around the corner, but before 2024 ends, I was able to send my first pull request to an OSS project. Wishing you a Happy New Year!

FlorianRappl:
> In Japan, the new year is just around the corner, but before 2024 ends, I was able to send my first pull request to an OSS project. Wishing you a Happy New Year!

Happy new year and thanks for your contribution!

FlorianRappl (approved):
LGTM!

noobow34:
Thank you for the review. I’m very happy to have been able to contribute to OSS for the first time!

PR #1223, opened 2025-04-10
GetAttributeNode extension methods added to Element

### LINKED ISSUE #1221: getAttributeNode and getAttributeNodeNS are missing
### Prerequisites

### Description

IElement doesn't have the GetAttributeNode and GetAttributeNodeNS methods and don't expose their DOM counterparts (getAttributeNode/getAttributeNodeJS)

### Steps to Reproduce

```
var config = Configuration.Default.WithJs();
var document = await context.OpenAsync(req => req.Content("")).ConfigureAwait(false);
Console.WriteLine(document.ExecuteScript("document.createElement('div').getAttributeNode"));
```
and similar for getAttributeNode

### Expected Behavior

Console output: info about function

### Actual Behavior

Console output: undefined

Implement both methods

arekdygas:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

GetAttributeNode(name) and GetAttributeNode(namespace, localName) added to IElement. Resolves #1221.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/DOMActions.cs
+++ b/src/AngleSharp.Core.Tests/Library/DOMActions.cs
@@ -1422,5 +1422,30 @@ namespace AngleSharp.Core.Tests.Library
             Assert.IsNotNull(template.Content.FirstChild);
             Assert.AreEqual(0, template.ChildNodes.Length);
         }
+
+        [Test]
+        public void GetAttributeNode()
+        {
+            var ns = "http://anglesharp.com/ns";
+
+            var parser = new HtmlParser();
+            var document = parser.ParseDocument("<html><head></head><body></body></html>");
+
+            var div = document.CreateElement("div");
+            var attributeValue = "abc";
+            div.SetAttribute(ns, "test:name", attributeValue);
+
+            var attrByName = div.GetAttributeNode("name");
+            var attrByPrefixAndName = div.GetAttributeNode("test:name");
+            var attrByNamespaceAndName = div.GetAttributeNode(ns, "name");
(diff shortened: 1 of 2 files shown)


FlorianRappl:
In general this is great, but it would be breaking in a way that users that implemented from `IElement` cannot just update. They would need to change (there are new implementations required; so it's not ABI compatible as well).

A nice way out here would be to instead use something like `ElementExtensions.cs` (maybe a new file such as DomElementExtensions.cs with the class being decorated as `[DomExposed("Element")]`) and place the two methods as extension methods there.

For instance:

```cs
[DomName("getAttributeNode")]
public static IAttr? GetAttributeNode(this IElement element, String name)
{
    if (NamespaceUri.Is(NamespaceNames.HtmlUri))
    {
        name = name.HtmlLower();
    }

    return Attributes.GetNamedItem(name);
}
```

arekdygas:
Yes - you're right. I'll try to fix it tomorrow.

arekdygas:
Do you think new file would be better than using already existing `ElementExtensions.cs`? The latter is not decorated with `DomExposedAttribute` now, but adding it and implementing these two methods should work as expected, as only methods decorated with `DomNameAttribute` are used during extensions processing. New file would be a bit cleaner, but I'm not sure if it's a good idea to have two classes extending `Element`, with the only difference, that one contains only members that could be exposed to JS.

What do you think?

FlorianRappl:
I agree. A new file would be cleaner imho, too.

arekdygas:
Done. Let me know if it looks ok.

FlorianRappl (approved):
LGTM - thanks for the wonderful work!

PR #1224, opened 2025-04-17
INode.HasChildNodes is now exposed as a method to DOM

### LINKED ISSUE #1219: INode.HasChildNodes is a property, and according to the spec it should be a method
### Prerequisites

### Description

Being a getter only property is not a problem in AngleSharp itself, but when it is exposed to JS DOM, via DomNameAttribute, then it doesn't conform to the spec.

I decided to report here and not in AngleSharp.JS, as the code in current repo requires fix.

Tested on master versions of AngleSharp, AngleSharp.Js and Jint

### Steps to Reproduce

```
var config = Configuration.Default.WithJs();
var document = await context.OpenAsync(req => req.Content("")).ConfigureAwait(false);
Console.WriteLine(document.ExecuteScript("document.createElement('div').hasChildNodes()"));
```

### Expected Behavior

Console should output: False

### Actual Behavior

Exception is thrown:

Jint.Runtime.JavaScriptException: 'Property 'hasChildNodes' of object is not a function'

Workaround: using `document.createElement('div').hasChildNodes` (without parentheses)

Proposed solutions:
1. Change HasChildNodes to method.
Pros: JS signature is the same as the one in C#
Cons: breaking change
 
2. Introduce ExposeAsMethod property in DomNameAttribute and set it to true for that property.
Pros: no breaking change in C#; there's still breaking change in JS if anyone used mentioned workaround, but impact would be much lower
Cons: different C#/JS signatures

arekdygas:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Fixes #1219

`Accessors` enum has now a new member: `Method`. This was required to expose `HasChildNodes` to DOM as a method, instead as a getter-only property.

Current pull request is only a part of the fix. Seconds part: AngleSharp/AngleSharp.Js#106

### DIFF
--- a/src/AngleSharp/Attributes/Accessors.cs
+++ b/src/AngleSharp/Attributes/Accessors.cs
@@ -31,6 +31,10 @@
         /// <summary>
         /// Specifies that the method should be handled by an event remover.
         /// </summary>
-        Remover = 0x10
+        Remover = 0x10,
+        /// <summary>
+        /// Specifies that the getter only property should be handled by a method.
+        /// </summary>
+        Method = 0x20,
     }
 }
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM - thanks!

PR #1227, opened 2025-05-04
Have IHtmlCollection<T> implement IReadOnlyList<T>

### LINKED ISSUE #1226: Have IHtmlCollection implement IReadOnlyList
### Description

Please have the `IHtmlCollection<T>` implement `IReadOnlyList<T>`, which is the conventional .NET interface for a collection that enables accessing by index and provides a length property, features which `IHtmlCollection<T>` already provides.

### Background

_No response_

### Specification

_No response_

weitzhandler:
# Types of Changes
Makes the `IHTmlCollection<T>` interface implement `IReadOnlyList<T>`, which is the conventional interface for types that expose a count property and enables index access, features that `IHtmlCollection<T>` already provides.

Closes #1226

## Prerequisites

## Contribution Type

## Description

Upgrades `IHtmlCollection<T>` to implement `IReadOnlyList<T>` instead of just `IEnumerable<T>` since it already implements all its functionality.

### DIFF
--- a/src/AngleSharp/Dom/IHtmlCollection.cs
+++ b/src/AngleSharp/Dom/IHtmlCollection.cs
@@ -1,43 +1,56 @@
-namespace AngleSharp.Dom
+namespace AngleSharp.Dom;
+
+using AngleSharp.Attributes;
+using System;
+using System.Collections.Generic;
+
+/// <summary>
+/// HTMLCollection is an interface representing a generic collection
+/// (array) of elements (in document order) and offers methods and
+/// properties for selecting from the list.
+/// </summary>
+[DomName("HTMLCollection")]
+public interface IHtmlCollection<T> :
+#if NET8_0_OR_GREATER
+    IReadOnlyList<T>
+#else
+    IEnumerable<T>
+#endif
+
+    where T : IElement
 {
-    using AngleSharp.Attributes;
-    using System;
-    using System.Collections.Generic;
+    /// <summary>
+    /// Gets the number of items in the collection.
+    /// </summary>
+    [DomName("length")]
+    Int32 Length { get; }
 
     /// <summary>
-    /// HTMLCollection is an interface representing a generic collection
-    /// (array) of elements (in document order) and offers methods and
-    /// properties for selecting from the list.
+    /// Gets the specific node at the given zero-based index into the list.
     /// </summary>
-    [DomName("HTMLCollection")]
-    public interface IHtmlCollection<T> : IEnumerable<T>
-        where T : IElement
-    {
-        /// <summary>
-        /// Gets the number of items in the collection.
-        /// </summary>
-        [DomName("length")]
-        Int32 Length { get; }
+    /// <param name="index">The zero-based index.</param>
+    /// <returns>Returns the element at the specified index.</returns>
+    [DomName("item")]
+    [DomAccessor(Accessors.Getter)]
+#if NET8_0_OR_GREATER
+    abstract T IReadOnlyList<T>.this[Int32 index] { get; }
+#else
+    T this[Int32 index] { get; }
+#endif
 
-        /// <summary>
-        /// Gets the specific node at the given zero-based index into the list.
-        /// </summary>
-        /// <param name="index">The zero-based index.</param>
-        /// <returns>Returns the element at the specified index.</returns>
-        [DomName("item")]
-        [DomAccessor(Accessors.Getter)]
-        T this[Int32 index] { get; }
+    /// <summary>
+    /// Gets the specific node whose ID or, as a fallback, name matches the
+    /// string specified by name. Matching by name is only done as a last
+    /// resort, only in HTML, and only if the referenced element supports 
+    /// the name attribute.
+    /// </summary>
+    /// <param name="id">The id or name to match.</param>
+    /// <returns>Returns the element with the specified name.</returns>
+    [DomName("namedItem")]
+    [DomAccessor(Accessors.Getter)]
+    T? this[String id] { get; }
 
-        /// <summary>
-        /// Gets the specific node whose ID or, as a fallback, name matches the
-        /// string specified by name. Matching by name is only done as a last
-        /// resort, only in HTML, and only if the referenced element supports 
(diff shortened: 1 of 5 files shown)


FlorianRappl:
Sorry this would be breaking and is therefore not acceptable.

Maybe we can have a non-breaking way (however, I doubt that - as I guess you explicitly want the interface to implement it).

weitzhandler:
Why do you consider this a breaking change? It's not affecting any existing functionality.

FlorianRappl:
It changes a public interface. So if you upgrade your code will not compile (assuming you implement that interface somewhere). If you just swap you'll end up with an exception. 

So it requires changes which is by definition a breaking change.

weitzhandler:
We could maybe overcome this with default interface implementation can't we?

FlorianRappl:
In general default interface methods could help (up to some extend), but as they require runtime support (starting with .NET Core 3.0, i.e., .NET standard 2.1) it would be another breaking change (see: https://github.com/AngleSharp/AngleSharp/blob/devel/src/AngleSharp.nuspec - it's .NET Standard 2.0 that we support in the lib).

weitzhandler:
Would u consider target-specific implementation?

```c#
    public interface IHtmlCollection<T>
#if NET8_0_OR_GREATER
        : IReadOnlyList<T>
#endif
```

weitzhandler:
For example (all tests pass after changing only this file to the following):

```csharp
namespace AngleSharp.Dom;

using AngleSharp.Attributes;
using System;
using System.Collections.Generic;

/// <summary>
/// This type represents a set of space-separated tokens.
/// </summary>
[DomName("DOMTokenList")]
public interface ITokenList :
#if NET8_0_OR_GREATER
    IReadOnlyList<String>
#else
    IEnumerable<String>
#endif
{
    /// <summary>
    /// Gets the number of contained tokens.
    /// </summary>
    [DomName("length")]
    Int32 Length { get; }

    /// <summary>
    /// Gets an item in the list by its index.
    /// </summary>
    /// <param name="index">The index of the item.</param>
    /// <returns>The item at the specified index.</returns>
    [DomName("item")]
    [DomAccessor(Accessors.Getter)]
#if NET8_0_OR_GREATER
    new
#endif
    String this[Int32 index] { get; }

    /// <summary>
    /// Returns true if the underlying string contains a token, otherwise
    /// false.
    /// </summary>
    /// <param name="token">The token to search for.</param>
    /// <returns>The result of the search.</returns>
    [DomName("contains")]
    Boolean Contains(String token);

    /// <summary>
    /// Adds some tokens to the underlying string.
    /// </summary>
    /// <param name="tokens">A list of tokens to add.</param>
    [DomName("add")]
    void Add(params String[] tokens);

    /// <summary>
    /// Remove some tokens from the underlying string.
    /// </summary>
    /// <param name="tokens">A list of tokens to remove.</param>
    [DomName("remove")]
    void Remove(params String[] tokens);

    /// <summary>
    /// Removes the specified token from string and returns false.
    /// If token doesn't exist it's added and the function returns true.
    /// </summary>
    /// <param name="token">The token to toggle.</param>
    /// <param name="force"></param>
    /// <returns>
    /// True if the token has been added, otherwise false.
    /// </returns>
    [DomName("toggle")]
    Boolean Toggle(String token, Boolean force = false);


#if NET8_0_OR_GREATER
    Int32 IReadOnlyCollection<String>.Count => Length;
#endif
}
```

FlorianRappl:
I think it might be an option; but it definitely needs further testing (for instance the swap test; can the lib just be changed or is a re-compilation necessary; which also indicates a breaking change - this time not API, but ABI).

weitzhandler:
It worked for me with a manual swap test.

weitzhandler:
I ran a manual test I created a new project with a project reference to AngleSharp.Core then replaced only the AngleSharp.Core.dll file with the output of the compilation containing the IReadOnlyList change

FlorianRappl:
> I ran a manual test I created a new project with a project reference to AngleSharp.Core then replaced only the AngleSharp.Core.dll file with the output of the compilation containing the IReadOnlyList change

Did the project contain a reference / a class defined to, e.g., implement `ITokenList`? Because this is the scenario we are looking for.

weitzhandler:
Yes, it's a console app that adds a `List<String>` subclass that implements `ITokenList` and makes use of the implementation adding values and accessing it's `Length` and `Count` properties, all based on the current _devel_ branch that only implements `IEnumerable<String>`.

After replacing _AngleSharp.dll_ in the console folder with the new one compiled after introducing the interface implementation change it still works as intended.

FlorianRappl:
Then let's do it :)

weitzhandler:
FYI
There's one test failing, but it also fails on the _devel_ branch.

FlorianRappl (commented):
In general I like it and I think we can go for it. However, I don't like that we need to hide the indexer using `new`. That feels fishy. I know its rather a .NET problem, but my question is: Wouldn't we be able to use `IReadOnlyCollection` instead of `IReadOnlyList`, i.e., not using the indexer? What's your take - do you need the list / or would a collection be sufficient?

weitzhandler:
The reason we're 'overriding' it is to allow for the documentation and attributes.

FlorianRappl:
> The reason we're 'overriding' it is to allow for the documentation and attributes.

I know that. I have nothing against the reason; I have something against hiding and I'd like to avoid it (as written).

weitzhandler:
Fixed

FlorianRappl:
Both systems are failing; any reason why you still use the list and not a collection? I don't really get it. As it stands the previous iteration was better imho.

weitzhandler:
Bear with me just a bit more please flo 😁
Reason IReadOnlyList would be preferred is because IHtmlCollection exposes all features of it just in a different name. I mean, if it works why not include it.

BTW, I've learned a bit of the code in the CSS selector, a true masterpiece!

FlorianRappl (approved):
I like it with the abstract way. I think this is a great way to deal with it. Lets have this in the preview and let's recheck for breaking changes.

PR #1244, opened 2026-03-30
Optimize Url parsing to reduce allocations and improve throughput

### LINKED ISSUE #575: Url class performance
Hey @FlorianRappl 

I was profiling some of my code yesterday and found that `Url` is pretty slow - in 25,000 requests that take around 45sec, 8s was spent in `Url`. Unfortunately it didn't give me the stack trace of what was calling the constructor (I was only constructing the url myself once).

How come you're not using Uri?

jafin:
I suspect it has already been resolved/improved in in the past 7 years since issue raised, , took a go anyhow. 
Some marginal improvements, could not replicate the original posters 8 sec spent in 25k url parsing (#575)

## Prerequisites

## Contribution Type

## Description

- Add fast-path in NormalizeInput to skip StringBuilder for clean URLs
- Replace List<String> + String.Join in ParsePath with direct StringBuilder
- Replace C0ControlAndSpace LINQ initialization with char literal
- Use HashSet<String> for protocol lookups instead of linear array scan

Benchmark Results: Before vs After (25k urls)

| Method | Before (Mean) | After (Mean) | Speedup | Before (Alloc) | After (Alloc) | Memory Saved |
|---|---|---|---|---|---|---|
| Parse | 2.205 ms | 1.856 ms | 15.8% faster | 1.72 MB | 1.24 MB | 27.9% less |
| Href | 2.336 ms | 2.013 ms | 13.8% faster | 2.01 MB | 1.53 MB | 23.9% less |
| Components | 2.284 ms | 2.066 ms | 9.5% faster | 1.90 MB | 1.42 MB | 25.3% less |

Benchmarked ~15% faster parse, ~25% less memory allocated.

Resolves #575

### DIFF
--- a/src/AngleSharp.Benchmarks/UrlBenchmark.cs
+++ b/src/AngleSharp.Benchmarks/UrlBenchmark.cs
@@ -1,0 +1,169 @@
+using AngleSharp.Dom;
+using BenchmarkDotNet.Attributes;
+
+namespace AngleSharp.Benchmarks
+{
+    [MemoryDiagnoser, ShortRunJob]
+    public class UrlBenchmark
+    {
+        private const int Iterations = 25;
+
+        private static readonly string[] _urls =
+        {
+            "http://localhost:8080/?mytest",
+            "https://www.google.de/mypath",
+            "http://www.google.de",
+            "http://example.org/foo/bar",
+            "http://example-domain.com/image.jpg",
+            "https://github.com/FlorianRappl/AngleSharp",
+            "http://www.google.de/some-path?a=b#header",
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
I think this should be good enough.

I'd actually close the linked issue. Comparing `Uri` to `Url` is just flawed from the beginning and `Url` does a lot more and is spec compliant, while Uri is not.

Of course anyone who wants to tackle this / sees further room for improvement here (there certainly is) is welcome to make a PR. Just no need to keep an open issue for it.

Thanks for the splendid work! 🚀

Issue #802, opened 2019-06-11
Improve WHATWG Url Spec Conformance


hyspace:
### Description

This Proposal contains 2 parts:
1. Making `Url` to catch up with latest *WHATWG URL Living Standard*
2. Keep tracking failing [web-platform-tests]

### Background

After reading the source code, I found that the `Url` is not aligned with lastest *WHATWG URL Living Standard*
The difference is trivial, but it is blocking me from using `Url` directly in my project. 

The main difference is that `Url` is not returning `parse failure` for some invalid input. In current implementation, it is reflected by the wrong value of `IsInvalid` property after parsing.

For example, 
Port number validation is not returning failure when port number is larger than 65535.

```cs
(new Url("http://example.com:65536")).IsInvalid // expect true, but got false
```

I heard from @FlorianRappl in [another thread](https://github.com/AngleSharp/AngleSharp/issues/800#issuecomment-500326840) about this:

> I don't see any error here. Ports are currently 16-bit, however, who knows when (or if) this will change. It should be up to the specific requester to then block a request (let's say the port is 99999 - what should we do about it? Just drop the port? note that invalid URLs will result to a default URL -> it will have potential negative side-effects) if its doing something invalid. It's a potential enhancement, but I would regard it as highly optional and I'm fine with omitting it.

I agree it is highly optional for most of the use cases, but for me it is not. I'm trying to get all links from web pages using `AngleSharp.Html` and trying to log down all invalid links for security related research. I mean "invalid" by not able to be open in browser. If you try the example in browser's address bar, it will redirect you to search engine with the invalid URL as keyword. If you have such link in a HTML file like this:
```html
<a href='http://example.con:65536'>link</a>
```
you will see it links to empty page.

I don't see any negative side-effects of make this example URL returning `parse error` here, but I do see that I'm getting wrong result with current `Url` and I see "potential negative side-effects" of not following the standard. Different than your statement, I found current Browsers is following standard pretty good, and even they fail some test cases, I can find information in [Issues like this](https://github.com/web-platform-tests/wpt/issues/4934)

I understand that it may not worth it to implement some part of the standard for performance. But instead, developers need to understand which part we are currently not following the Standard and why, so they can make decision easier. Today, I have to find out which part we are not following standard by myself case by case.

My solution will be catch up with "latest living standard as much as possible", and provide the information about which part we are not following the standard by providing information of "failing test cases" in `web-platform-tests`, and have documentation of why we are not doing it.

[...]

PR #1247, opened 2026-04-02
Add WHATWG URL Standard tests

jafin:
## Prerequisites

## Contribution Type

## Description

Adds a harness that runs AngleSharp's `Url` parser against the official WHATWG/WPT `urltestdata.json` dataset.

**This is now test-only — the PR makes no changes to the parser.** The parsing fixes previously on this branch were reverted, and the userinfo-leak fix has since landed on `devel` independently.

### What's here

- `WptUrlTests` runs every entry of the WPT url dataset in a single NUnit test and reports one informational line:

  ```
  WPT url tests: 520 of 891 pass (58.4%), 371 fail, baseline is 520.
  ```

- The pass count is asserted against a `MinimumPassing` baseline, so a **regression fails the build** while improvements stay green and surface in the summary. Raise the constant to lock in progress.
- Per-entry mismatch detail is only emitted when the count regresses, capped at 25 entries, so a normal run doesn't flood the test output.
- Each entry is checked for `href`, `protocol`, `username`, `password`, `host`, `hostname`, `port`, `pathname`, `search` and `hash`, plus `origin` where the data provides one. Only 411 of the 891 entries carry an `origin` key, so — as in the WPT runners — it is asserted only when present; an opaque origin is the literal string `"null"` in the data, and a null `Url.Origin` is normalised to match.
- `Resources/urltestdata.json` is a pinned copy of the dataset (wpt commit `181476aa16e8b28a07698bef3a0275fa53dd22e5`, 2026-07-05) embedded in the test assembly. The tests need no network access, and the baseline can't drift when WPT edits the data upstream. Refreshing the file is a deliberate act that may require rebasing the baseline.

### Notes

- The baseline is 520 because that is the count on **net10.0**; **net462 and net472 pass 529** of the same entries. The gap is worth a separate look — most likely `IdnMapping`/IDN host handling differing between .NET Framework and modern .NET.
- Current failures cluster around:
  - opaque-path schemes — `foo:/bar.com/` collapses `PathName` to `/`
  - IPv6 compression — `http://[::127.0.0.1]` is not canonicalised to `http://[::7f00:1]`
  - out-of-range ports not being rejected — `http://f:999999/c` parses successfully
  - empty-password serialization — `https://test:@test`
  - origins that should be opaque — `gopher:` and `blob:` with a non-http inner URL (e.g. `blob:ws://example.org/`) report an origin where WHATWG expects `null`, because `ProtocolNames.IsOriginable` treats them as originable

<details>
<summary>Original description (superseded)</summary>

~~I flagged this as a breaking change, as I have modified the existing tests to align with the WPT tests (if I've interpreted the result correctly)~~

  - Add a WPT (Web Platform Tests) URL test harness that validates parsing against the official urltestdata.json dataset.  It should warn not error on future variation from the test data.
  - ~~Fix credential serialization to omit empty username/password from Href output~~
  - ~~Fix inherited userinfo leaking when a relative URL specifies a new authority — ParseAuthority and ParseFileHost now clear _username/_password on entry so base URL credentials don't persist~~

  - ~~Fix URL parsing for non-special schemes (blob:, wss:, etc.) to correctly parse authority (host, port, credentials) when :// is present, instead of treating everything as opaque scheme data~~
  - ~~Fix opaque scheme data (data:, mailto:, etc.) to be returned via PathName instead of Data~~
  - ~~Update existing tests to match corrected WHATWG spec behavior~~

</details>

Relates to #802


src/AngleSharp/Dom/Url.cs
```diff
@@ -121,8 +122,9 @@ public Url(Url address)
             _host = address._host;
             _username = address._username;
             _password = address._password;
-             _relative = address._relative;
-            _schemeData = address._schemeData;;
+            _relative = address._relative;
+            _hasPath = address._hasPath;
+            _schemeData = address._schemeData; ;
```
FlorianRappl: Nitpick; I know this is not new - but since we touch this part anyway; let's remove that single ";".

FlorianRappl:
imho not much is breaking in this (could be rather seen as a bug); however,

> Fix opaque scheme data (data:, mailto:, etc.) to be returned via PathName instead of Data

I think this one is indeed breaking and I'd like to keep the existing behavior / have it non-breaking.

jafin:
> imho not much is breaking in this (could be rather seen as a bug); however,
> 
> > Fix opaque scheme data (data:, mailto:, etc.) to be returned via PathName instead of Data
> 
> I think this one is indeed breaking and I'd like to keep the existing behavior / have it non-breaking.

Reverted.  This now becomes basically just a soft check against  WHATWG URL Standard tests

FlorianRappl:
We now have 3000 lines of debug output that don't make much sense.

I think we should remove that output to be present.

jafin:
> We now have 3000 lines of debug output that don't make much sense.
> 
> I think we should remove that output to be present.

Done, log noise removed, single line summary, an assert at the test passing level  (to catch any regression backwards)   test data now included so it is pinned to avoid unexpected test failures

FlorianRappl (approved):
LGTM!

PR #1248, opened 2026-04-20
apply StringSyntaxAttribute

SimonCropp:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

### DIFF
--- a/src/AngleSharp/Common/StringSyntaxAttribute.cs
+++ b/src/AngleSharp/Common/StringSyntaxAttribute.cs
@@ -1,30 +1,29 @@
 #if !NET7_0_OR_GREATER
-namespace System.Diagnostics.CodeAnalysis
+namespace System.Diagnostics.CodeAnalysis;
+
+[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Field | AttributeTargets.Property)]
+internal sealed class StringSyntaxAttribute : Attribute
 {
-    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
-    internal sealed class StringSyntaxAttribute : Attribute
+    public const String Regex = nameof(Regex);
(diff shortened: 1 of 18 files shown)


src/AngleSharp/Common/StringSyntaxAttribute.cs
```diff
@@ -0,0 +1,29 @@
+#if !NET7_0_OR_GREATER
+namespace System.Diagnostics.CodeAnalysis;
+
+[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Field | AttributeTargets.Property)]
+internal sealed class StringSyntaxAttribute : Attribute
+{
+    public const String Regex = nameof(Regex);
+    public const String Uri = nameof(Uri);
+    public const String Json = nameof(Json);
```
FlorianRappl, line 9: Where is this used?

src/AngleSharp/Common/StringSyntaxAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+    public const String Xml = nameof(Xml);
```
FlorianRappl, line 10: This seems unused, too.

src/AngleSharp/Common/StringSyntaxAttribute.cs
```diff
@@ -0,0 +1,29 @@
+#if !NET7_0_OR_GREATER
+namespace System.Diagnostics.CodeAnalysis;
+
+[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Field | AttributeTargets.Property)]
+internal sealed class StringSyntaxAttribute : Attribute
+{
+    public const String Regex = nameof(Regex);
+    public const String Uri = nameof(Uri);
```
FlorianRappl, line 8: I would not introduce this. .NET URI is a not a good / spec-compliant implementation. We should use "Url" imho (similar to "Html") for these cases.

PR #1251, opened 2026-06-09
Fix O(n^2) regression when tokenizing large plain-text runs

SimonCropp:
The ScanDataText bulk-append optimization (added in 1.5.0) scanned the remaining buffer with IndexOfAny on every character. When no DataText terminator ('<', '&', '\0', '\r', '\n') was present before EOF, IndexOfAny returned -1, which `found <= 0 ? 0 : found` collapsed to a zero-length run. The tokenizer then advanced a single character and re-scanned the entire remaining buffer again, making tag-less text O(n^2): a 2 MB plain-text body took ~70s to parse (vs a few ms in 1.4.0).

Treat the "no terminator" result as a run extending to the end of the buffer so the whole remainder is appended in one pass. Parsing is now linear (and faster than 1.4.0, since the run is appended via a single span).

Add a regression test covering both ReadOnlyMemoryTextSource and CharArrayTextSource that asserts correct token output and guards against the quadratic behaviour returning.

# Types of Changes

## Prerequisites

## Contribution Type

## Description

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/HtmlTokenization.cs
+++ b/src/AngleSharp.Core.Tests/Html/HtmlTokenization.cs
@@ -6,6 +6,7 @@ namespace AngleSharp.Core.Tests.Html
     using AngleSharp.Text;
     using NUnit.Framework;
     using System;
+    using System.Diagnostics;
     using System.IO;
     using System.Linq;
     using System.Text;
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
Thanks - LGTM!

SimonCropp:
@FlorianRappl thanks. no rush, but do u have an eta on deployment? give this one is a blocker for our code base, so we have pinned on the previous version

FlorianRappl:
The preview is out and I'll either place the 1.5.1 today or eo the week depending if anything else comes in. Right now I placed it as Saturday, but could be also earlier.

SimonCropp:
thank for all your work

PR #1256, opened 2026-07-04
Fix Range.ExtractContent to follow the DOM Standard

ivandrofly:
## Description

`Range.ExtractContent()` (`extractContents()`) deviated from the [DOM Standard algorithm](https://dom.spec.whatwg.org/#dom-range-extractcontents) in several ways and corrupted the document:

- **Contained children** were computed as *all descendants* of the common ancestor that *intersect* the range, instead of the *direct children* that are *contained* in it. Partially contained nodes and nested descendants were therefore moved into the fragment a second time, corrupting both the fragment and the source tree.
- The **first/last partially contained child** was searched among all descendants instead of the direct children of the common ancestor, so nested boundaries picked the wrong node.
- Partially contained elements were **cloned deeply** (spec requires a shallow clone) and their subrange fragment was appended to the outer fragment instead of the clone, flattening and duplicating content.
- The **new boundary** computation climbed one level too far and passed node/offset to the wrong `Boundary` arguments, throwing `NullReferenceException` when the common ancestor has no parent (detached subtrees) and collapsing the range to a wrong position otherwise. It now mirrors the correct logic already used by `ClearContent`.

### Example

```csharp
// <p>abc<b>def</b>ghi</p>, range from 'abc'@1 to 'ghi'@2
var fragment = range.ExtractContent();
// Before: fragment = bca<b></b>defigh, <p> left empty
// After (matches browsers): fragment = bc<b>def</b>gh, <p> = ai
```

Extracting inside a detached subtree (e.g. a `<div>` created via `CreateElement`) previously threw `NullReferenceException`.

`ClearContent`/`deleteContents()` was already correct and is untouched.

## Types of changes

## Checklist

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -193,5 +193,61 @@ namespace AngleSharp.Core.Tests.Library
             Assert.IsFalse(htmlRaw.Contains("This should be deleted too"));
             Assert.IsFalse(document.Contains(toDelete));
         }
+
+        [Test]
+        public void CanExtractContentWithPartiallyContainedTextNodes()
+        {
+            var document = "<p>abc<b>def</b>ghi</p>".ToHtmlDocument();
+            var p = document.QuerySelector("p");
+            var range = document.CreateRange();
+            range.StartWith(p.ChildNodes[0], 1);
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1257, opened 2026-07-04
Fix Range.CopyContent to follow the DOM Standard

ivandrofly:
## Description

`Range.CopyContent()` (`cloneContents()`) deviated from the [DOM Standard algorithm](https://dom.spec.whatwg.org/#dom-range-clonecontents) and duplicated content:

- **Contained children** were computed as *all descendants* of the common ancestor that *intersect* the range, instead of the *direct children* that are *contained* in it, so partially contained nodes and nested descendants were cloned into the fragment multiple times.
- The **first/last partially contained child** was searched among all descendants instead of the direct children of the common ancestor.
- Partially contained elements were **cloned deeply** (spec requires a shallow clone) and their subrange fragment was appended to the outer fragment instead of the clone, flattening and duplicating content.

### Example

```csharp
// <p>abc<b>def</b>ghi</p>, range from 'abc'@1 to 'ghi'@2
var fragment = range.CopyContent();
// Before: fragment = bcabc<b>def</b>defghigh
// After (matches browsers): fragment = bc<b>def</b>gh
```

Same root cause as the `ExtractContent` fix in #1256, applied to the clone-only path.

## Types of changes

## Checklist

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -163,6 +163,40 @@ namespace AngleSharp.Core.Tests.Library
         }
 
         [Test]
+        public void CanCopyContentWithPartiallyContainedTextNodes()
+        {
+            var document = "<p>abc<b>def</b>ghi</p>".ToHtmlDocument();
+            var p = document.QuerySelector("p");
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1258, opened 2026-07-04
Fix swapped END_TO_END and END_TO_START in Range.CompareBoundaryTo

ivandrofly:
## Description

Per the [DOM Standard](https://dom.spec.whatwg.org/#dom-range-compareboundarypoints), `compareBoundaryPoints(how, sourceRange)` maps:

- `END_TO_END` → compare **this end** to source end
- `END_TO_START` → compare **this start** to source end

`Range.CompareBoundaryTo` used the opposite boundary of this range for those two cases (this start for `EndToEnd`, this end for `EndToStart`). The bug only surfaces with **overlapping** ranges — for disjoint ranges both mappings happen to produce the same result, which is presumably why it went unnoticed.

```csharp
// outer = (p,0)..(p,3), inner = (p,1)..(p,2)
outer.CompareBoundaryTo(RangeType.EndToEnd, inner);
// Before: RangePosition.Before (compared outer start (p,0) to inner end (p,2))
// After:  RangePosition.After  (compares outer end (p,3) to inner end (p,2)) — matches browsers
```

## Types of changes

## Checklist

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -1,5 +1,6 @@
 namespace AngleSharp.Core.Tests.Library
 {
+    using AngleSharp.Dom;
     using NUnit.Framework;
 
     [TestFixture]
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1259, opened 2026-07-04
Fix Range.IsCollapsed to also compare offsets

ivandrofly:
## Description

Per the [DOM Standard](https://dom.spec.whatwg.org/#range-collapsed), a range is collapsed only if its start and end are the **same boundary point** (same node *and* same offset). `Range.IsCollapsed` only compared the nodes:

```csharp
public Boolean IsCollapsed => _start.Node == _end.Node;
```

so any range whose boundaries live in the same container — e.g. after `Select(node)` (`selectNode()`), which sets `(parent, i)` .. `(parent, i + 1)` — was wrongly reported as collapsed.

## Types of changes

## Checklist

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/Range.cs
+++ b/src/AngleSharp.Core.Tests/Library/Range.cs
@@ -36,6 +36,25 @@ namespace AngleSharp.Core.Tests.Library
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

Issue #1222, opened 2025-03-25
NodeIterator doesn't work properly if nodes are removed


arekdygas:
### Prerequisites

- [x] Can you reproduce the problem in a [MWE](https://en.wikipedia.org/wiki/Minimal_working_example)?
- [x] Are you running the latest version of AngleSharp?
- [x] Did you check the FAQs to see if that helps you?
- [x] Are you reporting to the correct repository? (there are multiple AngleSharp libraries, e.g., `AngleSharp.Css` for CSS support)
- [x] Did you perform a search in the issues?

### Description

If we're iterating over nodes using NodeIterator and the currently visited node or its parent node is removed, then nextNode() returns null.

### Steps to Reproduce

``` C#
var config = Configuration.Default;
var context = BrowsingContext.New(config);
var html = "<div id='outer'><div id='inner1'></div><div id='inner2'></div></div>";
var document = await context.OpenAsync(req => req.Content(html)).ConfigureAwait(false);
var iterator = document.CreateNodeIterator(document.GetElementById("outer"), AngleSharp.Dom.FilterSettings.Element);
var node = iterator.Next(); // node is outer div
node = iterator.Next(); // node is inner 1 div
node.Parent.RemoveChild(node); // inner 1 is removed
node = iterator.Next(); // node should be inner2
```

### Expected Behavior

node should be inner2 DIV

### Actual Behavior

node is null

### Possible Solution / Known Workarounds

According to the specs (https://dom.spec.whatwg.org/#ref-for-nodeiterator-pre-removing-steps%E2%91%A0 and https://dom.spec.whatwg.org/#nodeiterator-pre-removing-steps), when a node is removed from document, then node iterators created within the same document must run pre-remove steps. Implementing these steps should fix the bug.

FlorianRappl:
Fixed and pushed to the preview.

PR #1260, opened 2026-07-04
Fix NodeIterator pre-removing steps to use the first following node

ivandrofly:
## Description

Per the [DOM Standard pre-removing steps](https://dom.spec.whatwg.org/#nodeiterator-pre-removing-steps), when a node containing the iterator's reference is removed while the pointer is *before* the reference, the new reference must be the removed node's first **following** node (within the root, not a descendant of the removed node). If there is none, the pointer flips to *after* and the reference becomes the deepest last descendant of the previous sibling.

The implementation searched `Root.GetDescendants()` **from the beginning of the tree**, so it always found the first node of the root's subtree outside the removed branch — which can be *before* the removed node. This moved the reference backwards, made `NextNode()` re-return already visited nodes instead of `null`, and made `PreviousNode()` return the wrong node; the pointer-before/previous-sibling fallback was effectively unreachable.

The fix skips ahead to the removed node in tree order before searching, matching the spec (follow-up to the pre-remove handling introduced for #1222).

```csharp
// <div id=root><a></a><b><i></i></b></div>, iterator at <i> (pointer before), remove <b>
iterator.Next();
// Before: <a> (already visited!)   After: null — matches browsers
```

## Types of changes

## Checklist

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/NodeIterator.cs
+++ b/src/AngleSharp.Core.Tests/Library/NodeIterator.cs
@@ -237,5 +237,53 @@
             Assert.AreEqual(inner1, node2);
             Assert.AreEqual(inner2, node3);
         }
+
+        [Test]
+        public void NodeIteratorShouldMoveReferenceToFollowingNodeOnRemoval()
+        {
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1263, opened 2026-07-14
Add bounded UTF-8 stream parsing mode

dv00d00:
# Types of Changes

## Prerequisites

## Contribution Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] My change requires documentation
- [x] I have added tests

## Description

Adds an opt-in bounded streaming source to concrete `HtmlParser`:

```csharp
ParseDocumentAsync(
    Stream source,
    HtmlStreamSourceMode sourceMode,
    Encoding? encoding = null,
    CancellationToken cancel = default);
```

- Existing stream overload remains buffered and unchanged.
- `Streaming` uses pooled byte/character buffers and bounded tokenizer lookback.
- Null encoding enables BOM and `<meta charset>` detection within a restartable 1024-byte prelude.
- A supplied encoding is authoritative.
- Scripting falls back to the buffered source.

Public surface and naming suggestions are welcome.

## Limitations

- Streaming mode does not retain complete `Document.Source.Text` or support `PrefetchAllAsync`.
- Consumed input outside the lookback window is discarded after the encoding prelude.
- The existing tokenizer may refill synchronously when a token exceeds the prefetched window.

## Benchmark

BenchmarkDotNet 0.15.8, .NET 8.0.28, 126,853-byte HTML, 4 KB stream reads:

| Method | Mean | Allocated | Gen2 / 1k |
|---|---:|---:|---:|
| Buffered | 4.100 ms | 2.12 MB | 117.19 |
| Streaming, explicit UTF-8 | 3.687 ms | 1.36 MB | 0 |
| Streaming, automatic encoding | 3.520 ms | 1.36 MB | 0 |

Streaming is approximately 10-14% faster with 36% less managed allocation on this fixture. A full default run of automatic streaming measured **3.576 ± 0.066 ms** and **1.36 MB**.

.NET Framework 4.8.1 automatic-streaming ShortRun: **8.523 ms**, **2.01 MB**. Cross-runtime timings are not directly comparable.

The simulated network stream completes `ReadAsync` synchronously; this measures decoding and parsing overhead, not socket latency.

Environment: Windows 11, Intel Core i9-9900K.

## Validation

- net8: 3,698 passed, 2 skipped
- streaming tests: 49/49 on net8 and net472
- clean builds: net462, net472, netstandard2.0, net8.0, net10.0
- benchmark builds: net472 and net8.0
- `git diff --check` passes


FlorianRappl:
In general looks good to me - and that its an opt-in public API (not changing existing methods) is appreciated.

Maybe the only nitpick from top of my head is `Utf8Streaming` - I'd just call it `Utf8` (because `Stream` already appears in the enum itself, i.e., `HtmlStreamSourceMode.Utf8Streaming` seems - at least to me - a bit redundant vs. `HtmlStreamSourceMode.Utf8`, however, maybe my view here is too narrow).

dv00d00:
I am not sure about the api shape tbf, there is a chance to follow spec cheaply 
<meta /> with encoding should arrive fully in the first KB of html
experimenting atm. thanks for the nit, no hard position at all. all I care is ram :D

src/AngleSharp/Html/Parser/HtmlParser.cs
```diff
@@ -243,6 +243,25 @@ public Task<IHtmlDocument> ParseDocumentAsync(Stream source, CancellationToken c
...
+        /// <param name="source">The byte stream to parse.</param>
+        /// <param name="cancel">The cancellation token.</param>
+        /// <param name="sourceMode">The decoding and retention strategy.</param>
+        /// <exception cref="NotSupportedException">
+        /// Thrown when bounded UTF-8 streaming is combined with scripting source insertion.
+        /// </exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown for an unknown source mode.</exception>
+        public Task<IHtmlDocument> ParseDocumentAsync(
+            Stream source,
+            CancellationToken cancel,
```
campersau: Should `CancellationToken` be the last parameter? At least for public API?
https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1068

FlorianRappl (approved):

PR #1159, opened 2023-12-24
Extensible DOM construction + allocation rate improvements

dv00d00:
# Types of Changes

## Prerequisites

Please make sure you can check the following two boxes:

- [x] I have read the **CONTRIBUTING** document
- [x] My code follows the code style of this project

## Contribution Type

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue, please reference the issue id)
- [x] New feature (non-breaking change which adds functionality, make sure to open an associated issue first)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

## Description

This is the followup to the read only DOM proposal.

General ideas: 

1. Reduce allocations by postponing strings creation as late as possible and use stack/pooled arrays as much more.
2. Allow a new way to construct parallel DOM hierarchy. 
3. Reduce allocations during tokenization, this version of tokenizer produces close to zero allocations.
4. Reduce the number of tokens to consume and transform into DOM by utilizing new options and tokenizer filter middleware.

I moved read only DOM implementation into separate repository: [link](https://github.com/dv00d00/AngleSharp.ReadOnlyDom/tree/main/AngleSharp.ReadOnlyDom)

Latest benchmark results:

|               Method |              UrlTest |          Mean |         Error |        StdDev |      Gen0 |      Gen1 |     Gen2 |   Allocated |
|--------------------- |--------------------- |--------------:|--------------:|--------------:|----------:|----------:|---------:|------------:|
|      HTMLAgilityPack |                    ? |    262.566 us |     11.228 us |     0.6155 us |   78.1250 |    3.9063 |        - |   639.76 KB |
|           AngleSharp |                    ? |    572.632 us |     30.914 us |     1.6945 us |   59.5703 |   19.5313 |        - |   492.71 KB |
| AngleSharpPrefetched |                    ? |    535.683 us |    129.238 us |     7.0840 us |   41.9922 |    8.7891 |        - |   349.34 KB |
|                      |                      |               |               |               |           |           |          |             |
|      HTMLAgilityPack |                  163 |  1,407.258 us |    119.582 us |     6.5547 us |  361.3281 |  134.7656 |        - |  2956.94 KB |
[...]

PR #1265, opened 2026-07-16
Expose asynchronous HTML token source seam

### LINKED ISSUE #1264: Allow asynchronous alternative token sources to feed HTML tree construction
## Problem

The HTML tree constructor currently owns and synchronously pulls from `HtmlTokenizer`. Alternative tokenizers cannot reuse AngleSharp's mature tree-construction correction without first adapting their input back into the existing text-source/tokenizer path.

This blocks consumers that already have a segmented or asynchronous tokenizer—for example, a UTF-8 tokenizer reading network buffers—from feeding tokens into AngleSharp while preserving foster parenting, implied elements, formatting reconstruction, and the rest of the HTML tree builder.

## Proposed direction

Introduce a public token-source contract that:

- exposes the tokenizer modes/options controlled by tree construction;
- lets the builder synchronously take a token when one is available;
- lets asynchronous parsing wait when the source needs more input;
- preserves the existing `HtmlTokenizer` path as the default implementation;
- exposes the minimal `StructHtmlToken` factories required by external implementations.

Add an async `HtmlParser.ParseDocumentAsync(IHtmlTokenSource, CancellationToken)` entry point. The caller should retain ownership of the supplied source.

dv00d00:
# Types of Changes

## Prerequisites

## Contribution Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] My change requires documentation

## Description

Closes #1264.

Adds a minimal asynchronous token-source seam between tokenization and HTML tree construction:

- `IHtmlTokenSource` exposes five operations: configure, change tokenizer state, change character-data acceptance, try-read, and wait for input;
- `HtmlParser.ParseDocumentAsync(IHtmlTokenSource, CancellationToken)` feeds alternative tokenizers into the existing tree builder;
- the built-in `HtmlTokenizer` keeps its existing public shape and is connected through an internal adapter;
- the parser does not dispose caller-owned token sources;
- the required `StructHtmlToken` factories remain public.

The motivating consumer is a segmented UTF-8 tokenizer reusing AngleSharp tree correction without retokenizing decoded text or materializing an intermediate DOM.

Naming and public-surface suggestions are very welcome.

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/AsyncTokenSource.cs
+++ b/src/AngleSharp.Core.Tests/Html/AsyncTokenSource.cs
@@ -1,0 +1,91 @@
+namespace AngleSharp.Core.Tests.Html
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using AngleSharp.Html.Dom.Events;
+    using AngleSharp.Html.Parser;
+    using AngleSharp.Html.Parser.Tokens;
+    using AngleSharp.Html.Parser.Tokens.Struct;
+    using AngleSharp.Text;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class AsyncTokenSourceTests
+    {
+        [Test]
+        public async Task TreeBuilderWaitsWheneverSegmentedTokenSourceNeedsInput()
+        {
+            using var source = new SegmentedTokenSource(
+                StructHtmlToken.Open("pre"),
+                StructHtmlToken.Character("\nhello", default),
+                StructHtmlToken.Close("pre"),
+                StructHtmlToken.EndOfFile(default));
+
+            var parse = new HtmlParser().ParseDocumentAsync(source);
+            Assert.IsFalse(parse.IsCompleted);
+            var document = await parse;
+
+            Assert.AreEqual("hello", document.Body.TextContent);
+            Assert.AreEqual(3, source.WaitCount);
+        }
+
+        private sealed class SegmentedTokenSource : IHtmlTokenSource
+        {
+            private readonly Queue<StructHtmlToken> _tokens;
+            private Boolean _available = true;
+
+            public SegmentedTokenSource(params StructHtmlToken[] tokens) => _tokens = new(tokens);
+
+            public Int32 WaitCount { get; private set; }
+
+            public event EventHandler<HtmlErrorEvent> Error;
+
+            public HtmlParseMode State { get; set; }
+            public Boolean IsAcceptingCharacterData { get; set; }
+            public Boolean IsStrictMode { get; set; }
+            public Boolean IsSupportingProcessingInstructions { get; set; }
+            public Boolean IsNotConsumingCharacterReferences { get; set; }
+            public Boolean IsPreservingAttributeNames { get; set; }
+            public Boolean SkipRawText { get; set; }
+            public Boolean SkipScriptText { get; set; }
+            public Boolean SkipDataText { get; set; }
+            public Boolean SkipComments { get; set; }
+            public Boolean SkipPlaintext { get; set; }
+            public Boolean SkipRCDataText { get; set; }
+            public Boolean SkipCDATA { get; set; }
+            public Boolean SkipProcessingInstructions { get; set; }
+            public Boolean DisableElementPositionTracking { get; set; }
+            public ShouldEmitAttribute ShouldEmitAttribute { get; set; } =
+                static (ref StructHtmlToken _, ReadOnlyMemory<Char> _) => true;
+            public Action<HtmlToken, TextRange> OnToken { get; set; }
+
+            public Boolean TryGetStructToken(out StructHtmlToken token)
+            {
+                if (!_available)
+                {
+                    token = default;
+                    return false;
+                }
+                token = _tokens.Dequeue();
+                _available = _tokens.Count == 0;
+                return true;
+            }
+
+            public async Task WaitForInputAsync(CancellationToken cancellationToken)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                WaitCount++;
+                await Task.Yield();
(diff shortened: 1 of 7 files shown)


FlorianRappl:
Looking over it - I think I like this. It looks good. I think a few consistency tweaks here or there (e.g., maybe we should make all these constructor helpers public then in the `StructHtmlToken`) could be good. Will need to think about it for a while (and I think all these changes would justify a 1.6.0).

dv00d00:
Compiled public API diff (`3312d02af` → `10663da1c`): **13 additions, 0 removals** on `net8.0`, `net462`, and `net472`.

```diff
+ public interface IHtmlTokenSource
+ {
+     void Configure(
+         HtmlTokenizerOptions options,
+         Action<HtmlToken, TextRange>? onToken,
+         Action<HtmlParseError, TextPosition> reportError);
+     void SetState(HtmlParseMode state);
+     void SetAcceptingCharacterData(bool value);
+     bool TryGetStructToken(out StructHtmlToken token);
+     Task WaitForInputAsync(CancellationToken cancellationToken);
+ }
+
+ public Task<IHtmlDocument> HtmlParser.ParseDocumentAsync(
+     IHtmlTokenSource tokenSource,
+     CancellationToken cancel = default);
+
+ public static StructHtmlToken StructHtmlToken.Doctype(bool quirksForced, TextPosition position);
+ public static StructHtmlToken StructHtmlToken.Open(StringOrMemory name);
+ public static StructHtmlToken StructHtmlToken.Close(StringOrMemory s);
+ public static StructHtmlToken StructHtmlToken.Character(StringOrMemory name, TextPosition position);
+ public static StructHtmlToken StructHtmlToken.Comment(StringOrMemory name, TextPosition position);
+ public static StructHtmlToken StructHtmlToken.EndOfFile(TextPosition position);
```

dv00d00:
This pr is the least resistance change to satisfy prototype utf8 parser.
New apis are niche and should not interfere with regular use.

Not sure if the least resistance path is the right one tbf, was hesitant to land on bigger diff surface.

No rush with review, I need to exit flow state to think about this.

dv00d00:
One possible option is to adopt this [bad boy](https://github.com/dv00d00/AngleSharp.ReadOnlyDom/blob/main/src/AngleSharp.ReadOnlyDom.Streaming/Utf8HtmlTokenizer.cs) and his friends in core

wire it as an optin for mutable dom materialisation

then new public surface becomes:

```csharp
public sealed class Utf8HtmlTokenReader
{
    public bool TryRead(out StructHtmlToken token);
    public Task WaitForInputAsync(CancellationToken cancellationToken);
}
```

and pr grows ~2k loc

FlorianRappl:
Yeah the question is: Is this worth it? (Honest question, see below @dv00d00 ) 

I think it might be a good addition, but if the primary use case is *your* use case then it might not be ideal, because for any change you'd need you'd first need to make a PR here and wait for it to be approved... Hence you will most likely want to let this within your space. However, of course, if you think the target audience is broad it would be good to either have it in a dedicated lib or directly in here.

I'd leave it up to you. In general the change here (making it more easy to extend from outside) is appreciated though.

dmytro-kushnir-cority-meddbase:
Two libraries came out of the tokenizer work. Different dependency stories, so I'll separate them.

**1. Eager read-only DOM** — columnar, DOM-compatible, navigation-compatible, serializable, SIMD queries. Built on the construction extension points from #1159; **needs no new core API** and compiles against released 1.5.2 today. Trades mutability for layout: on the checked-in corpus, ≈0.6× time and 0.28× allocations vs the existing parser (574 ms → 331 ms, 232 MB → 65 MB, .NET 10 Server GC).

**2. Streaming query lib** — forward-only, constant memory, UTF-8. Same architectural class as [Cloudflare's lol-html](https://github.com/cloudflare/lol-html); I'm not aware of a .NET equivalent. Extracting 16 articles from a 125 KB page: 201 µs / 14.9 KB vs 1,807 µs / 938 KB for the scraper-tuned CSS path — 9× faster, 63× less allocation. Fully consuming a 1.96 MB page with a small result near EOF allocates **2.85 KB, zero Gen0** — allocation scales with results, not input.

This one does depend on the UTF-8 tokenizer surface from #1265's branch, so I won't undersell it: it needs core API that isn't released. Worth saying why it's small, though — the tokenizer's only real coupling to core is `HtmlEntityProvider`'s entity table. That's also the argument for it living here rather than outside: anywhere else and I'm forking spec data.

All numbers from `./scripts/bench.ps1 small` and `./scripts/bench.ps1 scraping` on .NET 10 with Server GC, reproducible from the repo.

Conformance: the tokenizer passes the official html5lib tokenizer vectors — 6,810 cases, every initial state, re-run at 1-byte and 7-byte segmentation, vendored from a pinned commit with upstream MIT preserved.

Arbitrary output shapes: [HTML → JSON scraping](https://github.com/dv00d00/AngleSharp.ReadOnlyDom/tree/main/samples/AngleSharp.ReadOnlyDom.Samples), rewriting streamed HTML, [rendering to markdown](https://github.com/dv00d00/AngleSharp.ReadOnlyDom/tree/main/samples/AngleSharp.ReadOnlyDom.MarkdownProxy).

I'd like to contribute both — not as core, as extensions. Separate packages under the org at `0.x`, net8.0+, no changes or removals to existing public API, or whatever shape is least burdensome on your end.

This has little to do with this draft, happy to continue conversation in any suitable form.

FlorianRappl:
Sounds good to me. Let's tackle this outside this issue - you have my mail for further communications / finding a good exchange channel?

dv00d00:
Closing as superseded by the standalone AngleSharp.ReadOnlyDom.Streaming lane. The UTF-8 tokenizer and async PipeReader execution no longer need an asynchronous token-source seam in AngleSharp Core, so keeping this draft open would expand the upstream API for a consumer that has moved out of Core. The branch is retained for history.

PR #1270, opened 2026-07-28
Improve selector matching performance (fixes #929)

### LINKED ISSUE #929: Most efficient way to get matching element?
I have an application that scrapes an entire website and it runs in about 12 hours. It uses `WebClient.DownloadString` to get the html, then uses `HtmlParser.ParseDocument` to parse it. Then, I do a lot of other parsing on top of that. This happens for about 293,000 pages so I'm trying to save any little bit of time that I can.

I've noticed that I've got a lot of places where I call `IHtmlDocument.GetElementsByTagName(TAG).FirstOrDefault(QUERY_SELECTOR)`. I believe I could collapse this into some sort of `IHtmlDocument.QuerySelector(QUERY_SELECTOR)` which theoretically would speed up the time by returning after the first match, but some preliminary testing has shown `QuerySelector` to be slow vs the old method. For instance:

`IElement element = doc.GetElementsByTagName("h2").FirstOrDefault(x => x.TextContent.Contains("Climbing Directory"));`

takes about 1 ms, where

`IElement element = doc.QuerySelector("h2:contains('Climbing Directory')");`

takes about 23 ms.

Any suggestions for improving my code?

lahma:
# Improve selector matching performance (fixes #929)

Issue #929 reported that `doc.QuerySelector("h2:contains('...')")` was dramatically slower
than the hand-rolled `GetElementsByTagName("h2").FirstOrDefault(x => x.TextContent.Contains(...))`.

Profiling the current `devel` (ultra/ETW sampling at 8190 Hz, 127 KB W3C Selectors spec page,
1781 elements) showed the gap is **not** in `:contains` — it is in the traversal machinery
behind `QuerySelector`, plus a handful of per-candidate interface casts and a quadratic
sibling walk. Where the time went on the unmodified code:

| cost | share | cause |
|---|---|---|
| `CastHelpers.StelemRef` / `StelemRef_Helper` | 33.8% of the `QuerySelector` path | array covariance checks pushing onto `Stack<INode>` |
| `NodeExtensions.GetDescendantsAndSelf.MoveNext` | 36.8% of the `QuerySelector` path | iterator state machine + `Func<>` filter per query |
| `CastHelpers.IsInstanceOfInterface` | 9.9% of the `QuerySelectorAll` sweep | `nodes[i] is IElement` in the nth-child family (96.8% from one method) |
| `Element.get_PreviousElementSibling` | 10.0% of the `QuerySelectorAll` sweep | re-reads `parent.ChildNodes` twice per loop iteration |
| `StelemRef` | 11.4% of the `QuerySelectorAll` sweep | `Stack<Element>` pushes |

## Changes

1. **`SelectorExtensions.MatchAny`/`MatchAll`** no longer route through the
   `GetDescendantsAndSelf` iterator with a `Stack<INode>` and a delegate filter.
   `QuerySelector` was measurably slower than `QuerySelectorAll` for strictly less work.
2. **`INodeListAccessor.TryGetElement`** tests `NodeType` instead of casting to `IElement`,
   used by `First`/`LastChildSelector` and `First`/`LastTypeSelector`.
3. **Sibling access**: `Previous`/`NextElementSibling` hoist the child list out of the loop
   and drop the interface casts; `Node.ParentElement` avoids the interface cast helper
   (it is on the ancestor walk of every descendant/child combinator candidate);
   `SiblingCombinator` scans the parent's child list once instead of re-locating the element
   for every sibling, which was quadratic.
4. **New `ElementTreeEnumerator`** — depth-first pre-order walk over parent links with a
   stack of child *indices* rather than nodes. No covariance checks, and the working set is
   bounded by tree depth instead of tree width. Used by `QuerySelectorAll`, `MatchAny` and
   `MatchAll`.
5. **New `TextContainsMatcher`** — `:contains` streams descendant text nodes through
   Knuth-Morris-Pratt (prefix table built once, at selector parse time) instead of
   materializing `TextContent` for every candidate. Matches spanning a node boundary are
   still found. `Element.TextContent` itself also drops its nested LINQ iterators.

## Results

BenchmarkDotNet 0.15.8, AMD Ryzen 9 5950X, Windows 11. Default job (not ShortRun), machine
idle for both runs. StdDev under 1% throughout.

### Issue #929 — the reported scenario

Hand-rolled `GetElementsByTagName` + `TextContent.Contains` is the baseline (ratio 1.00).

**net10.0**

| | before | after |
|---|---|---|
| found — `h2:contains(...)` | 9.997 μs (0.29×) | **2.873 μs (0.10×)** |
| not found — `h2:contains(...)` | 63.182 μs (**1.93× slower**) | **26.902 μs (0.78×)** |
| not found — allocated | 23.14 KB (3.13×) | **176 B (0.09×)** |

**net8.0**

| | before | after |
|---|---|---|
| found — `h2:contains(...)` | 12.85 μs (0.34×) | **3.277 μs (0.09×)** |
| not found — `h2:contains(...)` | 100.43 μs (**2.49× slower**) | **29.451 μs (0.77×)** |
| not found — allocated | 23.3 KB (3.09×) | **184 B (0.10×)** |

The full-scan case went from ~2× slower than the hand-rolled loop to faster than it, on both
runtimes. The selector form now wins in both directions.

### `QuerySelector` (net10.0)

| selector | before | after | speedup |
|---|---:|---:|---:|
| `div` | 3,998 ns | 145 ns | 27.6× |
| `p:nth-child(2n+1)` | 4,418 ns | 160 ns | 27.6× |
| `div p` | 4,085 ns | 179 ns | 22.8× |
| `#title` | 3,571 ns | 185 ns | 19.3× |
| `div ~ p` | 6,658 ns | 818 ns | 8.1× |
| `ul.toc li.tocline2` | 7,175 ns | 1,377 ns | 5.2× |
| `p:contains(selectors)` | 6,093 ns | 1,273 ns | 4.8× |
| `.note` | 23,932 ns | 8,330 ns | 2.9× |
| `div.no-such-class` (full scan) | 61,804 ns | 27,255 ns | 2.3× |
| `div + p` | 33,785 ns | 16,903 ns | 2.0× |
| `p:only-child` | 111,092 ns | 80,451 ns | 1.4× |

Allocation is now a flat **176 B** per call. It was a flat **16,808 B** — paid even when the
match was the first element visited, because the node stack grew to the width of the document
regardless of when the search stopped.

### `QuerySelectorAll` (net10.0)

| selector | before | after | speedup |
|---|---:|---:|---:|
| `div ~ p` | 339,976 ns | 75,601 ns | 4.5× |
| `p:nth-child(2n+1)` | 158,150 ns | 94,504 ns | 1.7× |
| `p:contains(selectors)` | 116,171 ns | 75,399 ns | 1.5× |
| `ul.toc li.tocline2` | 40,464 ns | 30,192 ns | 1.3× |
| `#title` | 31,450 ns | 24,819 ns | 1.3× |
| `div` | 32,681 ns | 26,300 ns | 1.2× |
| `div + p` | 80,632 ns | 68,402 ns | 1.2× |
| `div p` | 45,044 ns | 40,038 ns | 1.1× |
| `p:only-child` | 110,207 ns | 109,585 ns | 1.0× |

`p:contains(selectors)` allocation: **235.75 KB → 1,256 B (192× less)**.

### Broad sweep — existing `SelectorBenchmark`, all 42 selectors

| | net10.0 | net8.0 |
|---|---|---|
| regressions over 3% | **0** | **0** |
| smallest gain | 1.13× | 1.14× |
| largest gain | 4.52× (`div ~ p`) | 4.04× (`div ~ p`) |

## Benchmark additions

`SelectorBenchmark` only ever covered `QuerySelectorAll`, so the slowest path had no coverage.
Added:

- `QuerySelectorBenchmark` — 11 selectors against **both** `QuerySelector` and
  `QuerySelectorAll`, default job for low noise; includes a deliberate no-match selector to
  force a complete traversal.
- `Issue929Benchmark` — the head-to-head from the issue, in found and not-found variants.
- `net10.0` added to the benchmark project's target frameworks.

## Tests

All existing tests pass on net10.0, net8.0, net472 and net462. Seven tests added, six covering
the `:contains` streaming matcher where the regression risk is real — matches spanning child
elements (`Climb<b>ing</b> Directory`), deep nesting, and correct restart on partial matches
(`aab` within `aaab`) both inside a single text node and across a node boundary — plus one
asserting `QuerySelector` returns the same element as `QuerySelectorAll[0]`, since those are
now two separate traversal implementations.

### DIFF
--- a/src/AngleSharp.Benchmarks/AngleSharp.Benchmarks.csproj
+++ b/src/AngleSharp.Benchmarks/AngleSharp.Benchmarks.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net472;net8.0</TargetFrameworks>
+    <TargetFrameworks>net472;net8.0;net10.0</TargetFrameworks>
     <SignAssembly>false</SignAssembly>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/src/AngleSharp.Benchmarks/Issue929Benchmark.cs
+++ b/src/AngleSharp.Benchmarks/Issue929Benchmark.cs
@@ -1,0 +1,44 @@
+using System.IO;
+using System.Linq;
+using AngleSharp.Dom;
+using AngleSharp.Html.Parser;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+
+namespace AngleSharp.Benchmarks
+{
+    /// <summary>
+    /// Reproduces https://github.com/AngleSharp/AngleSharp/issues/929 - looking up a single
+    /// element by tag name plus text content. The hand-rolled variant is the baseline the
+    /// issue reports as being dramatically faster than the equivalent CSS selector.
+    /// </summary>
+    [MemoryDiagnoser, GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory), CategoriesColumn]
+    public class Issue929Benchmark
+    {
+        private const string Present = "Selectors";
+        private const string Absent = "NoSuchTextAnywhereInThisDocument";
+
+        private static readonly HtmlParser parser = new HtmlParser();
+        private IDocument document;
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            document = parser.ParseDocument(File.ReadAllText("page.html"));
+        }
+
+        [BenchmarkCategory("Found"), Benchmark(Baseline = true)]
+        public IElement HandRolled() =>
+            document.GetElementsByTagName("h2").FirstOrDefault(x => x.TextContent.Contains(Present));
+
+        [BenchmarkCategory("Found"), Benchmark]
+        public IElement Selector() => document.QuerySelector("h2:contains(" + Present + ")");
+
+        [BenchmarkCategory("NotFound"), Benchmark(Baseline = true)]
+        public IElement HandRolledMiss() =>
+            document.GetElementsByTagName("h2").FirstOrDefault(x => x.TextContent.Contains(Absent));
+
+        [BenchmarkCategory("NotFound"), Benchmark]
+        public IElement SelectorMiss() => document.QuerySelector("h2:contains(" + Absent + ")");
+    }
+}
--- a/src/AngleSharp.Benchmarks/QuerySelectorBenchmark.cs
+++ b/src/AngleSharp.Benchmarks/QuerySelectorBenchmark.cs
@@ -1,0 +1,58 @@
+using System.Collections.Generic;
+using System.IO;
+using AngleSharp.Dom;
+using AngleSharp.Html.Parser;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+
+namespace AngleSharp.Benchmarks
+{
+    /// <summary>
+    /// Focused gate for selector matching performance. Unlike <see cref="SelectorBenchmark"/>
+    /// - which is a broad short-run sweep - this uses the default job for low noise and covers
+    /// both the "all matches" and the "first match" traversal, which are separate code paths.
+    /// </summary>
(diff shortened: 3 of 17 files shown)


FlorianRappl (approved):
LGTM!

PR #1272, opened 2026-07-30
Generate the package with dotnet pack instead of the hand-written nuspec

lahma:
Companion to AngleSharp/AngleSharp.Js#134 — you asked to check whether the same packaging approach works on the core repository before committing to it, so here it is with the package diff.

Short answer: it works, and the delta is smaller here than in AngleSharp.Js.

## The package diff

Built `Package` on `devel` and on this branch, then diffed the two generated nuspecs. `devel` on the left (`CopyFiles` + `NuGetPack` over the checked-in nuspec), this branch on the right (`dotnet pack`):

```diff
     <id>AngleSharp</id>
     <version>1.7.0</version>
     <authors>AngleSharp</authors>
-    <owners>Florian Rappl</owners>
-    <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <license type="expression">MIT</license>
     <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
     <icon>logo.png</icon>
@@
     <copyright>Copyright 2013-2026, AngleSharp.</copyright>
     <tags>html html5 css css3 xml dom dom4 parser engine hypertext markup language query selector attributes linq angle bracket web internet text headless browser</tags>
-    <repository type="git" url="https://github.com/AngleSharp/AngleSharp" />
+    <repository type="git" url="https://github.com/AngleSharp/AngleSharp" branch="refs/heads/modern-csproj" commit="c48371bf9baf40eff1e6cf22b1519b8ff439fb24" />
     <dependencies>
-      <group targetFramework=".NETStandard2.0">
-        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" />
-      </group>
+      <group targetFramework="net10.0" />
       <group targetFramework=".NETFramework4.6.2">
-        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" />
+        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" exclude="Build,Analyzers" />
       </group>
       <group targetFramework=".NETFramework4.7.2">
-        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" />
+        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" exclude="Build,Analyzers" />
       </group>
-      <group targetFramework="net6.0" />
-      <group targetFramework="net7.0" />
       <group targetFramework="net8.0" />
-      <group targetFramework="net10.0" />
+      <group targetFramework=".NETStandard2.0">
+        <dependency id="System.Text.Encoding.CodePages" version="8.0.0" />
+      </group>
     </dependencies>
```

Reading that as a list of actual changes:

- `net6.0` and `net7.0` groups are gone. Neither is in `TargetFrameworks`, so they were declaring dependency groups for frameworks the package has never contained a `lib/` folder for.
- All five real TFMs are present, and `System.Text.Encoding.CodePages` lands on exactly the three that need it. That is generated from the existing `IsTargetFrameworkCompatible` condition rather than restated by hand, so it cannot drift from the csproj.
- `owners` and `requireLicenseAcceptance` dropped. Both are deprecated; nuget.org resolves ownership from the account, and omitting `requireLicenseAcceptance` means the same as `false`.
- `repository` gains `branch` and `commit` — SourceLink, which the SDK does on its own.
- `exclude="Build,Analyzers"` spells out the default PackageReference asset behaviour rather than changing it.

The analyzer packages (`ConfigureAwaitChecker.Analyzer`, `Nullable`) correctly do not appear, since they are `PrivateAssets="all"`.

File listing is byte-for-byte identical — same five `lib/` folders with `dll` + `xml`, same `logo.png` and `README.md`:

```
$ diff <(cd devel-pkg && find . -type f | sort) <(cd dotnet-pack-pkg && find . -type f | sort)
$ # no output
```

## Assembly identity is unchanged

This was the part I most wanted to be sure of, since the package is strong-named and has a large .NET Framework audience. Published history bumps `AssemblyVersion` with every minor (1.5.0.0, 1.6.0.0, 1.7.0.0), so I deliberately left that flowing rather than pinning it:

```
devel  netstandard2.0 -> AngleSharp, Version=1.7.0.0, PublicKeyToken=e83494dcdc6d31ea
newPR  netstandard2.0 -> AngleSharp, Version=1.7.0.0, PublicKeyToken=e83494dcdc6d31ea
devel  net10.0        -> AngleSharp, Version=1.7.0.0, PublicKeyToken=e83494dcdc6d31ea
newPR  net10.0        -> AngleSharp, Version=1.7.0.0, PublicKeyToken=e83494dcdc6d31ea
```

FileVersion (`1.7.0.0`) and InformationalVersion (`1.7.0+c48371b`) match as well. The one intentional difference is `AssemblyCopyright`: `AssemblyInfo.cs` said `Copyright © AngleSharp, 2013-2024` while the nuspec said `2013-2026`. They now agree on the nuspec's value.

## Version authority

`nuke/Build.cs` already parsed `CHANGELOG.md` for the package version; it now also passes it as `-p:Version`. That removes the "bump it in two places" requirement AGENTS.md documented — `<Version>` in `Directory.Build.props` stays only as the fallback for builds that bypass the script. Since `CHANGELOG.md` and `<Version>` both said `1.7.0`, the built assembly is unchanged.

## Determinism was already right here

Worth noting since it came up on the AngleSharp.Js side: this repo already had `.SetContinuousIntegrationBuild(IsServerBuild)` on `Compile`, and the published package proves it works — AngleSharp 1.6.0's symbols contain zero absolute paths (692 documents mapped to `/_/`, 11 of the `Nullable` package's contentFiles to `/_1/`). AngleSharp.Js has no such call, which is why its published symbols carry the agent's `D:\a\...` paths and NuGet Package Explorer flags it. I have taken this repo's approach back to that PR rather than the other way round.

## What this touched beyond packaging

- **Test project to `net10.0`.** This raises it to the .NET 10 `AnalysisLevel`, which turned up `CA2022` (ignoring a `Stream.Read` return value) in `Mocks/FileEntry.cs` and `Library/FormSubmit.cs`. Both are real, if harmless on a `MemoryStream`. Fixed rather than suppressed — `FileEntry.Slice` now reads in a loop and writes only what it got, `Utf8StreamToString` uses `CopyTo`. Flagging it because it is the one behavioural edit in an otherwise build-only change, and it is the kind of thing that will recur on the other repositories when they move TFM.
- **CPM** via `src/Directory.Packages.props`, covering the library, tests and benchmarks.
- **Test tooling bumps**: `Microsoft.NET.Test.Sdk` 17.9.0 → 17.14.1, `NUnit3TestAdapter` 4.5.0 → 5.2.0, `GitHubActionsTestLogger` 2.3.3 → 2.4.1 (matching AngleSharp.Js), `Newtonsoft.Json` 13.0.1 → 13.0.4. NUnit stays on 3.14.0 — NUnit 4 moves the classic asserts to `ClassicAssert` and that is a separate change.
- **`DelaySign=false`** removed; it was conditioned on Windows and `false` is the default everywhere.
- **NUKE 10.0.0 → 10.1.0**, and the `NuGet.CommandLine` download dropped now that pack and push both go through `dotnet`.

## Deliberately not done

- No `.slnx` and no `global.json`, per your calls on the other PR. This repo's solution also has no orphaned config blocks to clean — the one project without a configuration is the `Solution Items` folder, which is legitimate.
- No Fallout migration. Agreed that wants to be its own change applied across the organisation.

## Verification

`Clean Package` end to end: 3741 tests pass on `net10.0`, `net462` and `net472`, in both `prefetched=false` and `prefetched=true` runs. Zero warnings with `TreatWarningsAsErrors` on.

---


src/AngleSharp.Benchmarks/Properties/AssemblyInfo.cs
```diff
@@ -1,7 +1,6 @@
 using System.Reflection;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyCopyright("Copyright © AngleSharp, 2013-2019")]
```
FlorianRappl, line 4: I think we can now fully remove the file, or is there any reason to keep it? Keeping the file was anyway just an oversight - that the copyright is duplicated / inconsistent was not desired.

src/AngleSharp.Core.Tests/Properties/AssemblyInfo.cs
```diff
@@ -1,7 +1,6 @@
 using System.Reflection;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyCopyright("Copyright © AngleSharp, 2013-2019.")]
```
FlorianRappl, line 4: Same story here - we can just remove the file.

FlorianRappl (approved):
LGTM!

src/AngleSharp.Benchmarks/Properties/AssemblyInfo.cs
lahma: Removed in 59cc966de. No reason to keep it — everything left in there was either empty or already the default: `AssemblyTrademark` and `AssemblyCulture` were `""`, `ComVisible(false)` is what absence means anyway, and the typelib `Guid` was declared but referenced nowhere (and neither assembly is COM-registered).

Both projects build clean and the 3741 tests still pass on net10.0/net462/net472 in both prefetched modes.

src/AngleSharp.Core.Tests/Properties/AssemblyInfo.cs
lahma: Gone too, same commit — `src/` now has no `AssemblyInfo.cs` at all. AngleSharp.Js#134 already had both of its files removed, so the two repositories match on this.

PR #1274, opened 2026-08-01
Use interface dispatch for tokenizer character reads

dv00d00:
# Types of Changes

## Prerequisites

- [ ] I have read the **CONTRIBUTING** document (the repository currently has no `CONTRIBUTING` file)

## Contribution Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] I have added a benchmark to cover my changes

## Description

Use the existing `IReadOnlyTextSource` reference directly for per-character tokenizer reads instead of checking cached concrete source references on every character.

The cached `CharArrayTextSource` and `ReadOnlyMemoryTextSource` references remain in place for the `ScanDataText` bulk-scanning optimization. The now-unused `WritableTextSource` cache is removed.

This also adds `TextSourceDispatchBenchmark`, which compares the previous cached concrete dispatch with interface dispatch for char-array, memory, and string-backed sources. Dedicated benchmark entry points avoid materializing unrelated benchmark parameter sources, so both the new microbenchmark and the existing end-to-end `ScanDataTextBenchmark` can run without the external page cache:

```console
dotnet run -c Release --project src/AngleSharp.Benchmarks -f net10.0 -- --text-source-dispatch
dotnet run -c Release --project src/AngleSharp.Benchmarks -f net10.0 -- --scan-data-text
```

### Benchmark results

BenchmarkDotNet 0.15.8, .NET 10.0.10, Intel Core i7-9750H, ShortRun. The end-to-end tokenizer benchmark at 10,000 repeats improved for every source type with unchanged allocations:

| Source | Before | After | Change | Allocated |
|---|---:|---:|---:|---:|
| String | 9.594 ms | 8.503 ms | -11.4% | 2954.8 KB |
| Char array | 5.379 ms | 4.895 ms | -9.0% | 2053.0 KB |
| Memory | 6.100 ms | 5.566 ms | -8.7% | 2053.0 KB |

The isolated 4,096-character dispatch benchmark shows the architecture-dependent trade-off explicitly: interface dispatch was 6% slower for `CharArrayTextSource`, 27% faster for `ReadOnlyMemoryTextSource`, and 22% faster for `StringTextSource`. The tokenizer-level benchmark is the representative result for this change.

### Validation

- `net472`, `net8.0`, and `net10.0` benchmark/core builds: passed with no warnings
- .NET 10 core test suite: 3,741 passed, 0 failed
- .NET 10 core test suite with `prefetched=true`: 3,740 passed, 0 failed

The .NET Framework target was compile-validated on macOS; runtime benchmarking for that target requires Windows.

### DIFF
--- a/src/AngleSharp.Benchmarks/Program.cs
+++ b/src/AngleSharp.Benchmarks/Program.cs
@@ -10,6 +10,24 @@ namespace AngleSharp.Benchmarks
         {
             Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 
+            if (
+                args.Length == 1
+                && args[0].Equals("--text-source-dispatch", StringComparison.OrdinalIgnoreCase)
+            )
+            {
+                BenchmarkRunner.Run<TextSourceDispatchBenchmark>();
+                return;
+            }
+
+            if (
+                args.Length == 1
+                && args[0].Equals("--scan-data-text", StringComparison.OrdinalIgnoreCase)
+            )
+            {
+                BenchmarkRunner.Run<ScanDataTextBenchmark>();
+                return;
+            }
+
             BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
         }
     }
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
LGTM!

CLA still needs a signature - then we are fine here.

PR #1277, opened 2026-08-03
Fast Dom Builder path and random improvements

dv00d00:
## Context

The experimental UTF-8 tokenizer has moved to AngleSharp.ReadOnlyDom.Streaming. This draft removes its duplicate implementation and supporting experiments from Core, leaving the smallest Core surface currently required by the Compact DOM work.

## What remains in Core

- generic handle-based HTML tree construction
- the generic custom-document parse entry point
- byte-backed text input used by the compact parser

## What is removed

- `Html/Parser/Utf8`
- the now-single-implementation `IHtmlTokenSource` adapter seam
- `System.IO.Pipelines` from the Core package
- tokenizer-only entity lookup, token reuse helpers, generator, benchmarks, tests, traces, and API report
- unrelated contiguous-source/tokenizer experiments from the same development branch

The cleanup commit removes roughly 12.3k lines from the current fork branch. The resulting GitHub diff against current upstream `devel` is 23 files, +1,390/-300, containing only the remaining Compact DOM integration surface.

## Verification

- `dotnet build src/AngleSharp/AngleSharp.Core.csproj -c Release` (`netstandard2.0`, `net8.0`, `net10.0`)
- Core tests in normal and `prefetched=true` modes: 3,748 passed in each mode
- AngleSharp.ReadOnlyDom test suite against this checkout: 179,310 passed
- benchmark project builds for `net10.0`
- NUKE project builds; existing NuGet vulnerability warnings remain unrelated

## Draft status

This is intentionally a draft while the public API and repository/package boundary for Compact DOM are discussed.


dv00d00:
Superseded by #1285, rebuilt from the current `devel` as two focused commits authored by `dv00d00`. The replacement keeps only the handle-based tree-construction backend and its tests; the byte-source and tokenizer experiments are intentionally excluded.

PR #1278, opened 2026-08-04
Keep auxiliary browsing contexts alive until their opener is gone

jafin:
# Types of Changes

## Prerequisites

## Contribution Type

## Problem

A browsing context opened as a **new window** (`target=`, `window.open`) is only ever reachable through weak references:

- `BrowsingContext._children` — `Dictionary<String, WeakReference<IBrowsingContext>>`
- `BrowsingContext._contextGroup` — `List<WeakReference<IBrowsingContext>>`
- `Document._attachedReferences` — `List<WeakReference>`

For an **iframe** this is harmless: `HtmlFrameElementBase._context` is a strong field, and the frame element is rooted by the DOM tree.

An **auxiliary (window) context has no owning element**. `DefaultNavigationHandler.NavigateAsync` resolves the target into a local, `CreateChildFor` registers only weak refs, and once that async method completes the new context and its `Active` document form an isolated cycle with **no strong root anywhere**. Any collection can then reclaim it, and a later `FindChild` returns `null` for a window that logically still exists.

## Fix

Root the auxiliary context in the **opening document**, mirroring the lifetime an opener/opened pair should have.

Frame contexts and the existing weak lookup structures are deliberately untouched — auxiliary contexts go into a separate list rather than `_attachedReferences`, so `FindChildRecursive`'s `GetAttachedReferences<BrowsingContext>()` walk behaves exactly as before. **Only the lifetime changes, not search semantics.**

## How this surfaced

Intermittent CI failures in `IframeTargets`, most often `AnchorTargetsSelfAsParentInNewWindow`, whose new-window document contains no iframes to incidentally keep it alive:

```
Failed AnchorTargetsSelfAsParentInNewWindow
  Expected: not null
  But was:  null
```

CI builds Release while local dev builds Debug — in Debug the JIT extends local lifetimes to end-of-method, keeping the context transitively alive, which is why it reproduces on CI and not locally.

The tests could **not** be fixed on their own: `DoClick` is fire-and-forget, and there is nothing to hold a strong reference to until `FindChild` returns — which is the racing call itself. A longer delay or a retry makes it *more* likely to fail, not less.

Three sibling tests use the same `FindChild("new-window")` pattern and were vulnerable to the identical race: `AnchorTargetsIframeFromDifferentWindow`, `AnchorTargetsIframeInDifferentWindow`, `AnchorTargetsCurrentWindowWithTop`.

## Verification

Reproduced deterministically with a standalone harness that forces a collection between the click and the lookup, A/B against the same base commit:

| Base | Forced GC | Result |
|---|---|---|
| unmodified | yes | **5/5 fail** (`newWindowContext=null`) |
| with fix | yes | **0/5 fail** |
| unmodified | no | 0/3 fail (race does not trigger) |

Full suite on `net10.0` Release: **3742 passed, 0 failed**, 2 pre-existing network-dependent skips. `IframeTargets` 11/11.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/BrowsingContextLifetime.cs
+++ b/src/AngleSharp.Core.Tests/Library/BrowsingContextLifetime.cs
@@ -26,6 +26,9 @@ namespace AngleSharp.Core.Tests.Library
                 "https://localhost/new-window.html" => CreateResponse(req.Address,
                     // language=html
                     "<html><body><h1>New window</h1></body></html>"),
+                "https://localhost/elsewhere.html" => CreateResponse(req.Address,
+                    // language=html
+                    "<html><body><h1>Elsewhere</h1></body></html>"),
                 "https://localhost/" => CreateResponse(req.Address,
                     // language=html
                     "<html><body><a id=\"link\" href=\"new-window.html\" target=\"new-window\">Load window</a></body></html>"),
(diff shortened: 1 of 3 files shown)


FlorianRappl (commented):
This is definitely something desirable - I am just not sure if that should be managed by the Document. I'd rather let it manage by the browsing context. It should know if a context is a direct descendant (e.g., an iframe) or was just opened from another document.

Right now I am therefore unsure if this is the right solution or if another solution is preferable.

jafin:
That makes sense, let me rework it onto the browsing context see how that comes out.

FlorianRappl (approved):
LGTM!

PR #1284, opened 2026-08-10
Match the annotation-xml encoding case-insensitively

### LINKED ISSUE #1283: annotation-xml with encoding=text/html is not treated as an HTML integration point
### Prerequisites

- [x] Can you reproduce the problem in a MWE? Yes, see below.
- [x] Are you running the latest version of AngleSharp? Reproduced on 1.7.1 and on `devel`.
- [x] Are you reporting to the correct repository? Yes, this is the core HTML tree construction.

### Description

A MathML `annotation-xml` element whose `encoding` attribute is `text/html` should be an HTML integration point, so its children are parsed with HTML rules. AngleSharp never treats it as one, so children are parsed as foreign (MathML) content instead.

Per the HTML Standard tree construction dispatcher, an `annotation-xml` element in the MathML namespace is an HTML integration point when its start tag had an `encoding` attribute whose value is an ASCII case-insensitive match for the string `text/html` (or `application/xhtml+xml`).

This came out of a security advisory that @FlorianRappl correctly closed as a correctness issue rather than a vulnerability. I confirmed that separately: the divergence is fail-safe for consumers such as sanitizers, because AngleSharp builds *more* markup than a browser rather than less. Filing it here as the conformance bug it is.

### Steps to Reproduce

```csharp
var doc = parser.ParseDocument(
    "<math><annotation-xml encoding=\"text/html\"><style><img src=x onerror=alert(1)></style></annotation-xml></math>");

var style = doc.QuerySelectorAll("*").First(e => e.LocalName == "style");
Console.WriteLine(style.NamespaceUri);          // namespace of <style>
Console.WriteLine(doc.Body.InnerHtml);
```

### Expected Behavior

`annotation-xml` with `encoding="text/html"` is an HTML integration point, so `<style>` is created in the HTML namespace and is a raw text element. Its contents are therefore **text**, and no `img` element exists.

The spec parsing algorithm agrees. Via html5lib:

```
encoding="text/html"              -> <html:style>&lt;img src=x onerror=alert(1)&gt;</html:style>
encoding="TEXT/HTML"              -> <html:style>&lt;img src=x onerror=alert(1)&gt;</html:style>
encoding="text/html; charset=..." -> <math:style/>   (correctly NOT an integration point)
```

Note the third line: the spec matches the literal string case-insensitively, it does not parse a MIME type, so a value carrying parameters is correctly *not* an integration point. AngleSharp already gets that case right.

### Actual Behavior

AngleSharp 1.7.1 puts `<style>` in the MathML namespace in every case, so it is not a raw text element and a live HTML `` element is constructed from its contents:

```
encoding="text/html"              style-ns=http://www.w3.org/1998/Math/MathML   img element exists = True
encoding="TEXT/HTML"              style-ns=http://www.w3.org/1998/Math/MathML   img element exists = True
encoding="text/html; charset=..." style-ns=http://www.w3.org/1998/Math/MathML   img element exists = True (escapes to body)
```

So both the exact lowercase `text/html` and any case variant fail to open an HTML integration point.

`IsHtmlTip` in `src/AngleSharp/Html/Parser/HtmlDomBuilder.cs` compares the raw `encoding` attribute value with ordinal equality against `MimeTypeNames.Html` and `MimeTypeNames.ApplicationXHtml`. An `OrdinalIgnoreCase` comparison would fix the case handling. Whether that alone is sufficient depends on where the integration-point decision is consumed, since the lowercase case fails too, which suggests the dispatcher may not be reaching this check on the path that matters.

I am happy to open a PR with a fix and tree-construction tests if that is useful, just say which shape you would prefer.

I used an AI assistant while investigating this. The parse outputs above are from runs I performed myself against AngleSharp 1.7.1 and html5lib.

arpitjain099:
## Prerequisites

## Contribution Type

- [x] Bug fix (non-breaking change which fixes an issue, please reference the issue id) - #1283

## Description

Fixes #1283.

`IsHtmlTip` compared the raw `encoding` attribute value against `MimeTypeNames.Html` and `MimeTypeNames.ApplicationXHtml` with ordinal equality. The spec matches that value ASCII case-insensitively, so `TEXT/HTML` was not recognised as opening an HTML integration point and the children were parsed as foreign content instead.

`ForeignNormalTag` already runs the identical test with `Isi` at line 3777, so the two halves of the parser disagreed on the same question. This makes the dispatcher agree with it, and reads the attribute through `GetAttribute` with the interned `AttributeNames.Encoding` the way that call does rather than indexing `Attributes` with a string literal.

### A correction to the issue

The issue I filed claims that exact lowercase `text/html` fails too, and it does not. I had not built it at that point. Running it on `devel` before the change:

```
encoding=text/html                  style-ns=http://www.w3.org/1999/xhtml         img=False
encoding=TEXT/HTML                  style-ns=http://www.w3.org/1998/Math/MathML   img=True
encoding=Text/Html                  style-ns=http://www.w3.org/1998/Math/MathML   img=True
encoding=application/xhtml+xml      style-ns=http://www.w3.org/1999/xhtml         img=False
encoding=text/html; charset=utf-8   style-ns=http://www.w3.org/1998/Math/MathML   img=True
```

So only the case variants were affected, and the dispatcher was reaching `IsHtmlTip` correctly all along. Sorry for sending you looking for a routing problem that was not there. The scope is smaller than the issue describes.

### Tests

`src/AngleSharp.Core.Tests/Html/HtmlIntegrationPoints.cs`. CONTRIBUTING asks for tests taken from the specification, and html5lib does cover this, but `HtmlWithMathML.cs` and `HtmlConstruction.cs` are generated from the `.dat` fixtures by `src/TestGeneration`, so I did not hand-edit them. If you would rather this lived in the generated suite I am happy to move it.

The cases cover the two accepted media types in several casings, a value carrying parameters, an unrelated value, an empty value and an absent attribute. `text/html; charset=utf-8` stays out of the integration point, which is correct: the spec matches the literal string rather than parsing a MIME type.

Before the change, 3 of the 10 fail, and they are exactly the case variants:

```
Failed AnnotationXmlWithHtmlEncodingIsAnIntegrationPoint("TEXT/HTML")
Failed AnnotationXmlWithHtmlEncodingIsAnIntegrationPoint("Text/Html")
Failed AnnotationXmlWithHtmlEncodingIsAnIntegrationPoint("APPLICATION/XHTML+XML")
Failed!  - Failed: 3, Passed: 7, Skipped: 0, Total: 10
```

After it, all 10 pass and the whole suite is green on `net10.0`:

```
Passed!  - Failed: 0, Passed: 3758, Skipped: 2, Total: 3760, Duration: 1 m 28 s
```

I left `CHANGELOG.md` alone since the top entry is the released 1.7.1 and the next version number is your call. Say the word and I will add a line under whichever heading you want.

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/HtmlIntegrationPoints.cs
+++ b/src/AngleSharp.Core.Tests/Html/HtmlIntegrationPoints.cs
@@ -1,0 +1,60 @@
+namespace AngleSharp.Core.Tests.Html
+{
+    using AngleSharp.Dom;
+    using NUnit.Framework;
+    using System;
+    using System.Linq;
+
+    /// <summary>
+    /// The tree construction dispatcher processes a token with HTML rules when the adjusted
+    /// current node is an HTML integration point. A MathML annotation-xml element is one when
+    /// its encoding attribute is an ASCII case-insensitive match for text/html or for
+    /// application/xhtml+xml. The html5lib-derived suites are generated from the .dat fixtures
+    /// by src/TestGeneration, so these cases live here instead.
+    /// </summary>
+    [TestFixture]
+    public class HtmlIntegrationPointTests
+    {
+        private static String AnnotationXmlWith(String encoding) =>
+            "<math><annotation-xml encoding=\"" + encoding + "\"><style><img src=x></style></annotation-xml></math>";
+
+        [TestCase("text/html")]
+        [TestCase("TEXT/HTML")]
+        [TestCase("Text/Html")]
+        [TestCase("application/xhtml+xml")]
+        [TestCase("APPLICATION/XHTML+XML")]
+        public void AnnotationXmlWithHtmlEncodingIsAnIntegrationPoint(String encoding)
+        {
+            var document = AnnotationXmlWith(encoding).ToHtmlDocument();
+            var style = document.All.First(m => m.LocalName == "style");
+
+            Assert.AreEqual(NamespaceNames.HtmlUri, style.NamespaceUri);
+            // In the HTML namespace style is a raw text element, so the img start tag
+            // is character data and no element is constructed from it.
+            Assert.IsFalse(document.All.Any(m => m.LocalName == "img"));
+        }
+
+        [TestCase("text/html; charset=utf-8")]
+        [TestCase("application/xhtml+xml; charset=utf-8")]
+        [TestCase("text/xml")]
+        [TestCase("")]
+        public void AnnotationXmlWithOtherEncodingIsNotAnIntegrationPoint(String encoding)
+        {
+            var document = AnnotationXmlWith(encoding).ToHtmlDocument();
+            var style = document.All.First(m => m.LocalName == "style");
+
+            // The value is matched as a literal string, not parsed as a MIME type, so a
+            // parameter makes it something other than text/html.
+            Assert.AreEqual(NamespaceNames.MathMlUri, style.NamespaceUri);
+        }
+
+        [Test]
+        public void AnnotationXmlWithoutEncodingIsNotAnIntegrationPoint()
+        {
+            var document = "<math><annotation-xml><style><img src=x></style></annotation-xml></math>".ToHtmlDocument();
+            var style = document.All.First(m => m.LocalName == "style");
+
+            Assert.AreEqual(NamespaceNames.MathMlUri, style.NamespaceUri);
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


FlorianRappl:
Hi @arpitjain099 just in case you missed it the CLA still needs signature. Much appreciated!

arpitjain099:
Hi @FlorianRappl thank you, just signed the CLA. Please review the PR. Much thanks!

FlorianRappl (approved):
LGTM! Much appreciated

PR #1285, opened 2026-08-20
Add handle-based HTML tree construction backend

dv00d00:
# Types of Changes

## Prerequisites

- [x] I have read the **CONTRIBUTING** document (the repository currently has no CONTRIBUTING document)

## Contribution Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Description

AngleSharp currently supports custom DOM construction through class-based `IConstructableElement` nodes. That still requires one CLR object per parsed node, which prevents arena-backed or index-backed consumers from reusing the spec-compliant HTML tree-construction algorithm without paying that allocation cost.

This change:

- generalizes the internal tree builder over a value-type node identity;
- adds `IHtmlTreeConstructionFactory<TDocument, TNode>` and `IHtmlTreeConstructionNode<TNode>` as the public backend seam;
- separates the parser-required document state from the optional browsing-host lifecycle;
- keeps the existing AngleSharp DOM path behind an internal adapter;
- exposes synchronous text-source and asynchronous stream entry points for handle backends;
- adds parity tests covering foster parenting, SVG/foreign content, and both parser entry points.

The concrete downstream consumer is the arena-backed Compact DOM in [AngleSharp.ReadOnlyDom](https://github.com/dv00d00/AngleSharp.ReadOnlyDom). It represents nodes as an arena reference plus an integer handle and returns `null` for `AsDomElement`.

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/HandleTreeConstructionTests.cs
+++ b/src/AngleSharp.Core.Tests/Library/HandleTreeConstructionTests.cs
@@ -1,0 +1,48 @@
+namespace AngleSharp.Core.Tests.Library;
+
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using AngleSharp.Dom;
+using AngleSharp.Html.Construction;
+using AngleSharp.Html.Parser;
+using AngleSharp.Text;
+using NUnit.Framework;
+
+[TestFixture]
+public sealed class HandleTreeConstructionTests
+{
+    private const string Markup =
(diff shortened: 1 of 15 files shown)


dv00d00:
CI note: both Linux and Windows completed the initial `dotnet build` successfully with 0 warnings and 0 errors. The jobs then failed inside the Fallout 10.4 orchestrator before its Restore/Compile/Test targets ran: the current GitHub runner installed .NET SDK 10.0.400, and Fallout could not load its expected `NuGet.Frameworks, Version=7.9.0.0` while evaluating `IsTargetFrameworkCompatible`. This is independent of the PR diff; the same branch passes the full 3,752-test Core run locally. I have left the failed run intact so the infrastructure error is visible in the logs.

PR #1286, opened 2026-08-20
Parse immutable byte buffers without a stream

dv00d00:
# Types of Changes

## Prerequisites

- [ ] I have read the **CONTRIBUTING** document (the repository currently has no CONTRIBUTING document)

## Contribution Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] I have updated the documentation accordingly

## Description

Applications that already hold an HTTP response as `byte[]` or `ReadOnlyMemory<byte>` currently have to wrap it in a `MemoryStream` or decode it to a `string` before parsing. Both routes add buffering or allocations that are unnecessary for a fully loaded immutable response.

This change adds:

- `ReadOnlyByteTextSource`, which retains the original bytes and decodes into a pooled character buffer;
- byte order mark detection and the same HTML encoding-restart behavior as the accumulating stream source;
- `HtmlParser.ParseDocument(ReadOnlyMemory<byte>, Encoding?)` on netstandard2.0, net8.0, net10.0, and the existing .NET Framework targets;
- tests for BOM handling, sliced memory, authoritative encodings, late meta charset restarts, and parity with the stream path;
- an HttpClient example and a repeatable BenchmarkDotNet benchmark.

Typical usage:

```csharp
using var response = await client.GetAsync("https://example.com/");
response.EnsureSuccessStatusCode();
ReadOnlyMemory<byte> html = await response.Content.ReadAsByteArrayAsync();
using var document = new HtmlParser().ParseDocument(html);
```

No `MemoryStream` and no intermediate `string` are required. An authoritative transport encoding can be supplied as the second argument; otherwise BOM detection starts the source and an HTML encoding declaration may restart decoding.

### Benchmark

BenchmarkDotNet ShortRun on the existing 123,989-byte `page.html` fixture, .NET 10:

| Input path | Mean | Allocated |
|---|---:|---:|
| `MemoryStream` | 5.739 ms | 2.08 MB |
| `ReadOnlyMemory<byte>` | 4.640 ms | 1.32 MB |

The direct buffer path was about 19% faster and allocated about 37% less in this run.

### Validation

- `dotnet build src/AngleSharp/AngleSharp.Core.csproj` — netstandard2.0, net8.0, net10.0; 0 warnings, 0 errors
- `dotnet build src/AngleSharp.Benchmarks/AngleSharp.Benchmarks.csproj` — net472, net8.0, net10.0; 0 warnings, 0 errors
- `dotnet test src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj -f net10.0` — 3756 passed, 2 skipped, 0 failed

This is independent of #1285 and contains none of the handle-based tree-construction changes. It extracts only the byte-buffer capability that was previously mixed into #1277.

### DIFF
--- a/docs/tutorials/03-Examples.md
+++ b/docs/tutorials/03-Examples.md
@@ -41,6 +41,32 @@ So we define some source code, call the `OpenAsync` method of a an `BrowsingCont
 
 In this case we use the provided source code to determine the content of the request's response. This content of the response is then parsed into an HTML document. Afterwards we serialize the DOM back to a string. Finally, we output this string in the console.
 
+## Parsing Raw Bytes from HttpClient
+
+When an HTTP response is already available as bytes, it can be parsed directly without wrapping the
+buffer in a `MemoryStream` or decoding it to a `string` first.
+
+```C#
+using System;
+using System.Net.Http;
+using AngleSharp.Html.Parser;
+
+using var client = new HttpClient();
+using var response = await client.GetAsync("https://example.com/");
+response.EnsureSuccessStatusCode();
+
+ReadOnlyMemory<byte> html = await response.Content.ReadAsByteArrayAsync();
+
+var parser = new HtmlParser();
+using var document = parser.ParseDocument(html);
+
+Console.WriteLine(document.Title);
+```
+
+Without an explicit encoding, AngleSharp detects a byte order mark and starts with UTF-8 while still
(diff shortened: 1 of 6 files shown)


FlorianRappl (approved):
LGTM!

FlorianRappl (approved):

PR #1287, opened 2026-08-20
Allow custom DOMs to retain whitespace-only text nodes

dv00d00:
# Types of Changes

## Prerequisites

- [ ] I have read the **CONTRIBUTING** document (the repository currently has no CONTRIBUTING document)

## Contribution Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Description

Generic custom-DOM construction currently passes `false` to `AppendText` and `InsertText` for whitespace-only runs. A custom DOM therefore turns:

```html
<span>Hello</span> <span>world</span>
```

into `Helloworld`, and there is no public way to retain the separating text node. The standard AngleSharp DOM already preserves it.

This adds the opt-in `HtmlParserOptions.IsKeepingWhitespaceTextNodes` option. Its default is `false`, so existing custom backends keep their current allocation and tree-shape behavior. The builder combines it with its constructor-level setting using OR, which also guarantees that the standard DOM continues to preserve whitespace regardless of the option value.

The production diff is intentionally small: one option property, one mutable builder flag, and one assignment when parser options are applied. Tests cover the custom default, custom opt-in, and unchanged standard DOM behavior.

### Validation

- `dotnet build src/AngleSharp/AngleSharp.Core.csproj` — netstandard2.0, net8.0, net10.0; 0 warnings, 0 errors
- `dotnet test src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj -f net10.0` — 3752 passed, 2 skipped, 0 failed

This is independent of #1285 and #1286.

### DIFF
--- a/src/AngleSharp/Html/Parser/HtmlParserOptions.cs
+++ b/src/AngleSharp/Html/Parser/HtmlParserOptions.cs
@@ -59,6 +59,13 @@ namespace AngleSharp.Html.Parser
         public Boolean IsKeepingSourceReferences { get; set; }
 
         /// <summary>
+        /// Gets or sets if whitespace-only text nodes should be kept when parsing into a custom
+        /// document through a generic tree-construction entry point. Custom documents omit these
+        /// nodes by default, while the standard AngleSharp DOM always keeps them.
+        /// </summary>
+        public Boolean IsKeepingWhitespaceTextNodes { get; set; }
+
+        /// <summary>
         /// Gets or sets if the parsing of character references should
         /// be avoided.
         /// Note: With this option there is no way to determine from
--- a/src/AngleSharp.Core.Tests/Html/CustomTreeWhitespace.cs
+++ b/src/AngleSharp.Core.Tests/Html/CustomTreeWhitespace.cs
@@ -1,0 +1,54 @@
+namespace AngleSharp.Core.Tests.Html;
+
+using System;
+using AngleSharp.Dom;
+using AngleSharp.Html.Parser;
+using AngleSharp.Text;
+using NUnit.Framework;
+
+[TestFixture]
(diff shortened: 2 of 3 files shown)


dv00d00:
CI note: both Linux and Windows completed the initial `dotnet build` successfully with 0 warnings and 0 errors. The jobs then hit the same current Fallout 10.4 / .NET SDK 10.0.400 infrastructure failure as #1285 and #1286: Fallout cannot load its expected `NuGet.Frameworks, Version=7.9.0.0` while evaluating `IsTargetFrameworkCompatible`, before its own Restore/Compile/Test targets run. The branch passes the full 3,754-test Core run locally; the failed run is left intact for the infrastructure logs.

src/AngleSharp/Html/Parser/HtmlParserOptions.cs
```diff
@@ -58,6 +58,13 @@ public struct HtmlParserOptions
         /// </summary>
         public Boolean IsKeepingSourceReferences { get; set; }
 
+        /// <summary>
+        /// Gets or sets if whitespace-only text nodes should be kept when parsing into a custom
+        /// document through a generic tree-construction entry point. Custom documents omit these
+        /// nodes by default, while the standard AngleSharp DOM always keeps them.
+        /// </summary>
+        public Boolean IsKeepingWhitespaceTextNodes { get; set; }
```
FlorianRappl, line 66: I am not sure I like this. Reason: The default (false) implies that white space text nodes are removed / skipped, which is not the default behavior for standard DOM / the parser.

Potentially we either need to transport the specific meaning (i.e., exclusive for custom DOM construction) in the name / another (extended) set of options. I am yet unsure how to deal with it.

PR #1290, opened 2026-09-01
Fixed tokenizer buffer overrun on encoding change

jafin:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

Parsing untrusted markup through `ParseDocument(ReadOnlyMemory<Byte>)` can throw
`IndexOutOfRangeException` from `ArrayPoolBuffer.Append`, so a `meta` charset declaration in
attacker-controlled input crashes the parser.

`BaseTokenizer` rents its character buffer for the source's character count, which is only
enough for one pass under one encoding. Handling a `meta` charset declaration breaks both
halves of that assumption:

* `HtmlDomBuilder.Restart()` tokenizes the source again into the same buffer, whose write
  position is never rewound, so the second pass appends behind the tokens of the abandoned
  one. Reachable when the already-consumed prefix decodes differently under the new encoding.
* Re-decoding the same bytes with a single-byte encoding turns every multi-byte sequence into
  that many characters, so the source itself grows past the rented length. Reachable with an
  ASCII prefix, where the source keeps its position and no restart happens at all.

Either one alone runs `Append` off the end of the rented array. Only `ReadOnlyByteTextSource`
combines a known content length (which selects `ArrayPoolBuffer` over `StringBuilderBuffer`)
with the possibility of an encoding restart, so `ParseDocument(ReadOnlyMemory<Byte>)` without
an explicit encoding is the affected entry point.

### The fix

`Restart()` now rewinds the buffer, and handling the declaration grows it to the source's new
character count. The growth abandons the old array rather than returning it to the pool,
because the `StringOrMemory` values already handed out are slices of it and their lifetime is
tied to the buffer. Nothing changes on the per-character append path.

### Tests

Two regression tests in `Vulnerabilities/ParserVulnerabilityTests.cs`, one per trigger. Both
fail on `devel` with `IndexOutOfRangeException` and pass with the fix. The full suite passes in
both `prefetched` modes.

Found by fuzzing the parser with SharpFuzz/libFuzzer; the smallest reproducer libFuzzer arrived
at was 50 bytes.

### DIFF
--- a/src/AngleSharp.Core.Tests/Vulnerabilities/ParserVulnerabilityTests.cs
+++ b/src/AngleSharp.Core.Tests/Vulnerabilities/ParserVulnerabilityTests.cs
@@ -1,5 +1,8 @@
 namespace AngleSharp.Core.Tests.Vulnerabilities;
 
+using System;
+using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using AngleSharp.Html.Parser;
 using NUnit.Framework;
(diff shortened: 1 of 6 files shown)


FlorianRappl (approved):
LGTM!

PR #1291, opened 2026-09-01
Fixed selector text dropping the case-insensitive modifier

jafin:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

`ISelector.Text` drops the ASCII case-insensitive modifier of an attribute selector, so a
selector does not survive being serialized and parsed again:

```csharp
var selector = new CssSelectorParser().ParseSelector("[href$=\"B\" i]");
selector.Text;                                    // [href$="B"]  -- the " i" is gone
document.QuerySelectorAll("[href$=\"B\" i]");     // 1 element
document.QuerySelectorAll(selector.Text);         // 0 elements
```

Every attribute selector took the `insensitive` flag, turned it straight into a
`StringComparison` for matching and kept nothing else, so `Text` had nothing left to write.
`Text` is what consumers store, compare, log and re-parse, and the result silently parses as a
case-sensitive selector. All seven value-carrying operators are affected: `=`, `~=`, `|=`,
`^=`, `$=`, `*=` and `!=`.

### The fix

The flag moves to `BaseAttrSelector`, which now owns both the `Comparison` used to match and
the `Modifier` used to serialize, derived from the same field — so the two cannot drift apart
again. The subclasses lose their own `_comparison` field and read `Comparison`, which is a
readonly field behind a property and costs nothing on the matching path.

`ISelectorVisitor.Attribute(name, op, value)` has the same gap and is left alone here, since
giving it the modifier means changing a public interface.

### Tests

`CssSelectorParserTests` gains a case for the reported selector and a `TestCase` per operator,
asserting the modifier survives serialization and a re-parse. All eight fail on `devel` and
pass with the fix; the full suite passes.

Found by fuzzing the selector parser with SharpFuzz/libFuzzer, checking that a parsed selector
and its own serialization select the same elements.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelectorParserTests.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelectorParserTests.cs
@@ -1,5 +1,6 @@
 namespace AngleSharp.Core.Tests.Css
 {
+    using System;
     using AngleSharp.Css.Parser;
     using NUnit.Framework;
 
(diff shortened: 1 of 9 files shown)


FlorianRappl (approved):
Great!

PR #1292, opened 2026-09-01
Fixed selector text dropping the nth-child of clause

jafin:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

`ISelector.Text` drops the `of S` clause of `:nth-child` and `:nth-last-child`, so the
serialization names a different selector:

```csharp
var selector = new CssSelectorParser().ParseSelector("li:nth-child(2 of .item)");
selector.Text;          // li:nth-child(0n+2)  -- the filter is gone
selector.Specificity;   // (0, 0, 2, 1), but the re-parsed text gives (0, 0, 1, 1)
```

`ChildSelector.Text` wrote back only the `An+B` part. The of-clause decides which siblings are
counted at all, so losing it widens the selector, and it takes the inner selector out of the
specificity with it.

### The fix

`ChildSelector.Text` appends ` of <filter>` whenever the filter is not the universal selector.
Only `:nth-child` and `:nth-last-child` accept the clause (`withOptionalSelector: true` in
`CssSelectorConstructor`); everywhere else the filter is `AllSelector.Instance` and the output
is unchanged.

### Tests

`CssSelectorParserTests` gains a case for the reported selector and a `TestCase` for each of
the two pseudo-classes that accept the clause, asserting that the text and the specificity both
survive a serialize and re-parse. All three fail on `devel` and pass with the fix; the full
suite passes.

Found by fuzzing the selector parser with SharpFuzz/libFuzzer, checking that a parsed selector
and its own serialization agree on specificity and on the elements they select.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelectorParserTests.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelectorParserTests.cs
@@ -1,5 +1,6 @@
 namespace AngleSharp.Core.Tests.Css
 {
+    using System;
     using AngleSharp.Css.Parser;
     using NUnit.Framework;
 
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1299, opened 2026-09-02
Fix ArgumentOutOfRangeException on \D800 and \DFFF CSS escapes

meziantou:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

A selector containing the CSS escape `\D800` or `\DFFF` throws `ArgumentOutOfRangeException` instead of matching, from every selector position (class, id, attribute name/value, type, namespace).

`Int32.IsInvalid` (`Text/CharExtensions.cs`) used **exclusive** bounds for the surrogate range:

```csharp
c == 0 || c > Symbols.MaximumCodepoint || (c > 0xD800 && c < 0xDFFF)
```

so the two boundary code points U+D800 and U+DFFF reported *valid* and fell through to `Char.ConvertFromUtf32` in `Css/Parser/CssStringSourceExtensions.cs`, which throws on any lone surrogate. The interior of the range (`\DBFF`, `\DC00`) was already handled correctly.

Per [css-syntax § consume an escaped code point](https://drafts.csswg.org/css-syntax/#consume-escaped-code-point), a surrogate escape must yield U+FFFD. The fix makes the bounds inclusive, so the boundaries take the same replacement-character path as the rest of the range. That `Int32` overload has exactly one call site, so nothing else shifts — the `IsInvalid` on `IElement` and on `Url` are unrelated members.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -1361,6 +1361,31 @@ nav h1, nav h2, nav h3, nav h4, nav h5, nav h6";
             Assert.AreEqual(selectorText, selector.Text);
         }
 
+        // https://drafts.csswg.org/css-syntax/#consume-escaped-code-point - a surrogate escape yields U+FFFD.
+        [TestCase(@"\D800")]
+        [TestCase(@"\DBFF")]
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM!

PR #1310, opened 2026-09-02
Fixed IStringMap.Remove leaving the data-* attribute in place

### LINKED ISSUE #1303: IStringMap.Remove leaves the data-* attribute in place instead of removing it
Observed while generating JavaScript bindings for AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3634, "AngleSharp + Jint"), on AngleSharp 1.7.x.

`IStringMap.Remove(name)` (the `dataset` map) sets the value to `null` and leaves the `data-*` attribute on the element, so `element.HasAttribute("data-foo")` stays true after removal and the map still enumerates the name. The DOM's `DOMStringMap` deleter (https://html.spec.whatwg.org/multipage/dom.html#dom-domstringmap-removeitem) removes the attribute.

A script-facing binding has to filter `null` values out of the map's supported names to keep `delete el.dataset.foo` observable, which is the workaround we carry. Happy to send a PR if you would take one.

lahma:
`IStringMap.Remove(name)` set the `data-*` attribute's value to `null` instead of removing the attribute, so `HasAttribute("data-foo")` stayed true afterwards, the map kept enumerating the name, and the attribute was still serialized. It now deletes the attribute from the element, which is what the `DOMStringMap` deleter specifies (https://html.spec.whatwg.org/multipage/dom.html#dom-domstringmap-removeitem).

The name is still validated the same way before the removal, and removing an unknown name remains a no-op. The tests cover the attribute being gone, the map no longer enumerating the name, and the serialization.

Fixes #1303

### DIFF
--- a/src/AngleSharp.Core.Tests/Html/DOM.cs
+++ b/src/AngleSharp.Core.Tests/Html/DOM.cs
@@ -186,6 +186,20 @@ namespace AngleSharp.Core.Tests.Html
         }
 
         [Test]
+        public void DOMStringMapRemoveDeletesTheAttribute()
+        {
+            var document = new HtmlDocument();
+            var div = new HtmlDivElement(document);
+            div.SetAttribute("data-some", "test");
+            div.SetAttribute("data-another", "value");
+            div.Dataset.Remove("some");
+            Assert.IsFalse(div.HasAttribute("data-some"));
+            Assert.AreEqual("<div data-another=\"value\"></div>", div.ToHtml());
+            Assert.AreEqual(1, div.Dataset.Count());
+            Assert.AreEqual("another", div.Dataset.Single().Key);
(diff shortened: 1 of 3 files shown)


lahma:
The checks shown are this branch's original green run. I briefly re-ran it with an identical tree to test whether the red CI on #1314 was environmental; it came back with 37 failures, so I restored the original commit.

`http://anglesharp.azurewebsites.net/` is answering `403 Web App - Unavailable` at the moment, so every live-network test in `Library/FormSubmit.cs` and `Library/ContextLoading.cs` fails regardless of the change under test. That is what #1314 is hitting, not its own diff.

FlorianRappl (commented):
LGTM!

PR #1311, opened 2026-09-02
Added the DomLiterals annotation to AdjacentPosition

### LINKED ISSUE #1306: AdjacentPosition lacks the [DomLiterals] annotation the other WebIDL enumerations carry
Observed while generating JavaScript bindings for AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3634), on AngleSharp 1.7.x.

`AngleSharp.Dom.AdjacentPosition` is the WebIDL enumeration behind `insertAdjacentElement`/`insertAdjacentHTML`/`insertAdjacentText` (`beforebegin`, `afterbegin`, `beforeend`, `afterend`), but unlike the other string enumerations it carries no `[DomLiterals]` markers, so a generator reading the attributes as WebIDL cannot tell it from an ordinary enum and has to guess from the member names. Adding the same annotation the other enumerations carry would close that gap; happy to send the one-file PR.

lahma:
`AdjacentPosition` is the string enumeration behind `insertAdjacentElement`/`insertAdjacentHTML`/`insertAdjacentText` (https://html.spec.whatwg.org/multipage/dom.html#dom-element-insertadjacentelement), and its members already carry the spec literals as `[DomName]`, but the enum itself was not marked `[DomLiterals]` the way `ShadowRootMode` is. A binding generator reading the attributes as WebIDL therefore could not tell it from an ordinary numeric enum.

This adds the enum-level `[DomLiterals]`; no enum-level `[DomName]`, since the spec models the position argument as a plain `DOMString` rather than a named IDL enumeration, which is also how `DocumentReadyState` is annotated. The new test pins the annotation and the four literals.

Fixes #1306

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/DomAnnotationTests.cs
+++ b/src/AngleSharp.Core.Tests/Library/DomAnnotationTests.cs
@@ -1,0 +1,31 @@
+namespace AngleSharp.Core.Tests.Library
+{
+    using AngleSharp.Attributes;
+    using AngleSharp.Dom;
+    using NUnit.Framework;
+    using System;
+    using System.Reflection;
+
+    [TestFixture]
+    public class DomAnnotationTests
+    {
+        [Test]
+        public void AdjacentPositionIsAnEnumerationOfLiterals()
+        {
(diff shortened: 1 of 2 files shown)


FlorianRappl (approved):
LGTM

PR #1312, opened 2026-09-02
Fixed Document.CurrentScript answering the deferred-script queue

### LINKED ISSUE #1308: Document.CurrentScript answers the deferred-script queue head, not the classic script being run
Observed while building a page runtime on AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3648), on AngleSharp 1.7.x.

`Document.CurrentScript` returns the head of the *deferred*-script queue, so during the execution of an ordinary classic `<script>` — the one case pages actually read it in (`document.currentScript.dataset.config`, `document.currentScript.src`) — it answers `null`. The HTML standard's `currentScript` (https://html.spec.whatwg.org/multipage/dom.html#dom-document-currentscript) is the script element whose script is currently being run by the *prepare a script* steps, set around the execution of each classic script and `null` otherwise.

A host running scripts through `IScriptingService` can set nothing here because the property is read-only, so it has to shadow `currentScript` on its own document wrapper. Happy to send a PR.

lahma:
`Document.CurrentScript` returned the head of the deferred-script queue, so it was null exactly when a page reads it — during the execution of an ordinary classic `<script>` — and, for a deferred script, named the next queued element rather than the running one. It is now set around the evaluation of each classic script in `ScriptRequestProcessor` and restored afterwards, so it is null outside script execution and correct for inline, external and deferred classic scripts.

A script element whose `type` is `module` leaves it null, as the execution steps specify (https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-element, exposed by https://html.spec.whatwg.org/multipage/dom.html#dom-document-currentscript). The deferred queue itself is untouched; only the property stopped reading from it. Tests cover the inline, external, deferred and module cases through the scripting service.

Fixes #1308

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/BasicScripting.cs
+++ b/src/AngleSharp.Core.Tests/Library/BasicScripting.cs
@@ -4,10 +4,12 @@ namespace AngleSharp.Core.Tests.Library
     using AngleSharp.Browser;
     using AngleSharp.Core.Tests.Mocks;
     using AngleSharp.Dom;
+    using AngleSharp.Html.Dom;
     using AngleSharp.Io;
     using AngleSharp.Text;
     using NUnit.Framework;
     using System;
+    using System.Collections.Generic;
     using System.Reflection;
     using System.Text;
     using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


FlorianRappl (approved):
LGTM!

PR #1313, opened 2026-09-02
Reflected DOMString attributes return the empty string when absent

### LINKED ISSUE #1304: Reflected DOMString attributes return null for an absent content attribute where the DOM specifies the empty string
Observed while generating JavaScript bindings for AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3634), on AngleSharp 1.7.x.

Reflected `DOMString` attributes return `null` when the content attribute is absent — `element.Id`, `element.ClassName` and the rest of the `[DomName]` string properties over `GetAttribute` — where the reflection rules (https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes, "if the attribute is not present, return the empty string") specify `""`. A binding that projects the property straight through therefore hands script `null` where a browser hands `""`, which breaks `el.id.length`-style code; we currently substitute the empty string in the generated getter.

If changing the CLR return value is too breaking for 1.x, an attribute marking the property as a reflected DOMString would let a binding do the substitution without a hand-written table. Happy to send a PR either way.

lahma:
A reflected `DOMString` IDL attribute returns the empty string when the content attribute is absent (https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes), but AngleSharp returned `null`, so a script binding that projects the property straight through hands `null` where a browser hands `""`. This is the first step, limited to the members that every element carries, so the change stays reviewable; the CLR types stay `String?` and the setters are untouched.

Changed: `IElement.Id` (`id`), `IElement.ClassName` (`class`), `IElement.Slot` (`slot`), `IHtmlElement.Title` (`title`), `IHtmlElement.Direction` (`dir`) and `IHtmlSlotElement.Name` (`name`, the counterpart of `slot`, which has to move with it so an unnamed `<slot>` still matches an element without a `slot` attribute). `IHtmlElement.AccessKey` already did this.

Three call sites relied on the `null`: `:target` now requires a non-empty `id` rather than a non-null one; `IsPersistent`/`IsPreferred`/`IsAlternate` now ask whether the link has a `title` attribute, which is what the stylesheet-set rules actually say (https://html.spec.whatwg.org/multipage/semantics.html#link-type-stylesheet); and a text node's default slot is looked up by the empty name. Two `nth-child` selector tests asserted `null` for a missing `class` and now assert `""`, with a comment saying why.

A follow-up can cover the remaining per-element reflections (`alt`, `src`, `href`, `type`, `placeholder`, `name`, `value`, …). Those need to be taken one at a time, because several look like plain reflections and are not: `crossOrigin` is `DOMString?` and must keep returning `null`, `contentEditable` is an enumerated attribute whose missing default is `"inherit"`, `HTMLOptionElement.value`/`label` fall back to the element's text, and `input.value` has value modes. `dir` is also "limited to only known values", so an unknown value should reflect as `""` too — left out of this PR.

Fixes #1304

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -809,7 +809,8 @@ nav h1, nav h2, nav h3, nav h4, nav h5, nav h6";
             Assert.AreEqual("span", result[1].GetTagName());
             Assert.AreEqual("span", result[2].GetTagName());
             Assert.AreEqual("italic", result[0].ClassName);
-            Assert.AreEqual(null, result[1].ClassName);
+            // A missing class attribute reflects as the empty string, not null.
+            Assert.AreEqual("", result[1].ClassName);
             Assert.AreEqual("this", result[2].ClassName);
             Assert.AreEqual("2", result[0].TextContent);
             Assert.AreEqual("4", result[1].TextContent);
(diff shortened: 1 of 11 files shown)


FlorianRappl (approved):
LGTM!

PR #1314, opened 2026-09-02
Added the DomSameObject annotation for the [SameObject] IDL members

### LINKED ISSUE #1305: querySelectorAll's return type and [SameObject] collections, as seen from a script binding
Two IDL-shape observations from generating JavaScript bindings for AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3634), on AngleSharp 1.7.x. Neither is wrong for a .NET API; both are visible the moment the objects are handed to script, so noting them for the record rather than asking for a break.

1. `IParentNode.QuerySelectorAll` returns `IHtmlCollection<IElement>`, where the DOM specifies a static `NodeList` (https://dom.spec.whatwg.org/#dom-parentnode-queryselectorall); a binding that maps by CLR type shows `[object HTMLCollection]` for the result and has to special-case the method.
2. `IElement.Children` (and the other `[SameObject]` collections: `Attributes`, `ClassList`, `ChildNodes`) returns a fresh collection on each access, where WebIDL's `[SameObject]` promises one object per element, so `el.children === el.children` is false unless the binding caches per element.

If a future major wanted to align, a `NodeList`-typed return for `QuerySelectorAll` and a cached collection per element would do; otherwise a `[DomSameObject]`-style attribute would let bindings cache without a hand-written list.

lahma:
`DomSameObjectAttribute` is a marker in `AngleSharp.Attributes`, in the same shape as `DomPutForwards`/`DomLenientThis` and with no behaviour of its own, so a binding can tell which members WebIDL declares `[SameObject]` and cache one projected object per owner instead of carrying a hand-written list. The members below were taken from the extracted IDL of the three standards, not from the prose.

DOM (https://dom.spec.whatwg.org/): `Node.childNodes`, `Element.attributes`, `Element.classList`, `ParentNode.children`, `Document.implementation`, `MutationRecord.target`/`addedNodes`/`removedNodes`, `NodeIterator.root`, `TreeWalker.root`.

HTML (https://html.spec.whatwg.org/multipage/): `Document.all`/`anchors`/`images`/`embeds`/`plugins`/`links`/`forms`/`scripts` (`embeds` and `plugins` are one property here, so one marker covers both), `HTMLElement.dataset`, `HTMLFormElement.elements`, `HTMLFieldSetElement.elements`, `HTMLSelectElement.options`/`selectedOptions`, `HTMLDataListElement.options`, `HTMLOutputElement.htmlFor`, `HTMLTableElement.rows`/`tBodies`, `HTMLTableSectionElement.rows`, `HTMLTableRowElement.cells`, `HTMLLinkElement.relList`/`sizes`, `HTMLAnchorElement.relList`, `HTMLAreaElement.relList`, `HTMLIFrameElement.sandbox`, `HTMLMapElement.areas`, `HTMLMediaElement.audioTracks`/`videoTracks`/`textTracks`.

CSSOM (https://drafts.csswg.org/cssom/): `StyleSheet.media`, and `styleSheets` on both `IDocumentStyle` and `IShadowRoot`.

Left out on purpose: `HTMLTemplateElement.content` and `HTMLInputElement.files` look like candidates but are not `[SameObject]` in the IDL; `validity` is `[SameObject]` only on `HTMLFieldSetElement`, while AngleSharp declares it once on the shared `IValidation`, so marking it there would over-apply it to five other controls. `CSSStyleSheet.cssRules` and `ElementCSSInlineStyle.style` live in AngleSharp.Css. The test walks the assembly and asserts every annotated member is a read-only property of a reference type carrying `[DomName]`, plus a list of members that must and must not carry it.

Fixes #1305

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/DomSameObjectTests.cs
+++ b/src/AngleSharp.Core.Tests/Library/DomSameObjectTests.cs
@@ -1,0 +1,73 @@
+namespace AngleSharp.Core.Tests.Library
+{
+    using AngleSharp.Attributes;
+    using AngleSharp.Dom;
+    using AngleSharp.Html.Dom;
+    using NUnit.Framework;
+    using System;
+    using System.Linq;
+    using System.Reflection;
+
+    [TestFixture]
+    public class DomSameObjectTests
+    {
+        private static PropertyInfo[] GetAnnotatedProperties()
+        {
+            return typeof(IElement).Assembly.GetTypes()
+                .SelectMany(m => m.GetProperties())
+                .Where(m => m.GetCustomAttribute<DomSameObjectAttribute>() != null)
+                .ToArray();
+        }
+
+        [Test]
+        public void SameObjectMembersAreReadOnlyPropertiesOfAReferenceType()
+        {
+            var properties = GetAnnotatedProperties();
+
+            Assert.IsNotEmpty(properties);
+
+            foreach (var property in properties)
+            {
+                var name = property.DeclaringType.Name + "." + property.Name;
+                Assert.IsFalse(property.PropertyType.IsValueType, name + " must return a reference type");
+                Assert.IsNotNull(property.GetMethod, name + " must be readable");
+                Assert.IsNull(property.SetMethod, name + " must be read only");
+                Assert.IsNotEmpty(property.GetCustomAttributes<DomNameAttribute>(), name + " must be a DOM member");
+            }
+        }
+
+        [TestCase(typeof(INode), nameof(INode.ChildNodes))]
+        [TestCase(typeof(IElement), nameof(IElement.Attributes))]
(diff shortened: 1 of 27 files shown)


lahma:
The red CI is the test host rather than this change: `http://anglesharp.azurewebsites.net/` is answering `403 Web App - Unavailable` right now, and every failure is one of the 37 live-network tests in `Library/FormSubmit.cs` and `Library/ContextLoading.cs`. The same 37 fail on #1310, whose only change is one line in `StringMap.Remove`, and which was green at 15:40 UTC before the app stopped.

The whole suite is green locally on net10.0. Happy to push again once the app is started.

FlorianRappl (approved):
LGTM!

The web app quota is currently exceeded.

Issue #687, opened 2018-07-06
Create new Console Application and OpenAsync just stops working


LeeAntW:
I finally got my code work in your **FormSubmissionDemo**.App working using AngleSharp and it all works fine.
However, i created a new Console App in VS2017 and when i try to execute the line:-
**await browsingContext.OpenAsync(########);**

The Console application window closes and the application just stops with no error message.

I see i don't have a reference to **System.Threading.Tasks** but it doesn't come up when i click add references i browse to the  **System.Threading.Tasks** from your **FormSubmissionDemo**.App but still doesn't work.

Is this becuase i am using a newer version of Framework 4.6.1?

I then uninstalled the GIT anglesharps packs and added your source code so i could debug the code and the code stops working on this step (**BrowsingContextExtensions.cs**)
On line  **using (var response = await download.Task.ConfigureAwait(false))**  in the method below:-

public static async Task<IDocument> OpenAsync(this IBrowsingContext context, DocumentRequest request, CancellationToken cancel)
        {
            if (request == null)
                throw new ArgumentNullException(nameof(request));

            var loader = context.Loader;

            if (loader != null)
            {
                var download = loader.DownloadAsync(request);
                cancel.Register(download.Cancel);

                **using (var response = await download.Task.ConfigureAwait(false))**
                {
                    if (response != null)
                    {
                        return await context.OpenAsync(response, cancel).ConfigureAwait(false);
                    }
                }
            }

            return await context.OpenNewAsync(request.Target.Href).ConfigureAwait(false);
        }

PR #1315, opened 2026-09-02
The HTML parser tracks an exception escaping host code instead of throwing it or dropping it

### LINKED ISSUE #1309: Host-owned navigation and parsing: location setters navigate on the caller's thread, sync-parse script exceptions are dropped, ReadyState cannot be set by a host
Three loose ends observed while building a page runtime on AngleSharp's DOM in Jint (https://github.com/sebastienros/jint/pull/3648), on AngleSharp 1.7.x — grouped because they are all about who owns navigation and parsing when a host drives the document:

1. Setting `location.host` (and `pathname`, `port`, `protocol`, `search`) starts a navigation on the *setter's* thread through a fire-and-forget `async void` in `Document.LocationChanged`; the caller cannot observe, await or cancel it, and an exception in it is unobservable. It is inert only when no `IRequester` is registered. A host that owns navigation would want the setter to raise an event or return the task.
2. `HtmlDomBuilder.Parse` (the synchronous parse) assigns `_waiting = RunScript(script)` and never observes the task, so an exception thrown by a script whose task does not complete synchronously is silently dropped.
3. `Document.ReadyState`'s setter is unreachable outside the assembly, so a host that owns the parse (running scripts itself, deciding when `DOMContentLoaded` fires) cannot report `interactive`; the document advances straight to `complete` during the parse.

None blocks us today; each would let a host drive the document more faithfully. Happy to send PRs for any you would take.

lahma:
`HtmlDomBuilder` keeps the task a parse still has to wait for in `_waiting`. `ParseAsync` awaits it, so an exception escaping host code on that task is rethrown out of `ParseDocumentAsync` — even though the parser itself never throws. The synchronous `Parse` never awaits that task at all, so the very same failure is dropped without a trace, leaving only a faulted task for the finalizer thread to notice.

Every assignment now goes through `WaitFor`, which wraps the task in one that observes the failure and reports it through `IConstructableDocument.TrackError` — the browsing context's `error` event carrying a `TrackEvent`, i.e. the channel `ScriptRequestProcessor` already uses for a failing scripting service. No new event type, no new public API. A task that already ran to completion is not wrapped, so a document with no scripts and no pending downloads allocates nothing extra.

## What was observed on `devel` (ff20ab6)

I built a repro for each half of the issue before writing any code, since an issue body is a hypothesis.

**The exception half is real, but not quite where the issue said.** `ScriptRequestProcessor.RunAsync` already wraps `IScriptingService.EvaluateScriptAsync` in a `try`/`catch` that calls `IBrowsingContext.TrackError`, so a service doing `await Task.Yield(); throw;` *is* reported today — 1 tracked error in all three parse shapes. What is not covered is everything else on the `_waiting` chain: the events the processor fires around the script, and `Document.FinishLoadingAsync`. With a host listener that throws:

| listener that throws | `ParseDocument` on `devel` | `ParseDocumentAsync` on `devel` |
| --- | --- | --- |
| `beforescriptexecute` | returns; **0** errors tracked | **throws** `InvalidOperationException` out of the parser |
| `afterscriptexecute` | returns; **0** errors tracked | **throws** |
| `DOMContentLoaded` | returns with `readyState == Interactive`; **0** errors tracked | **throws** |

After this change every one of those six cells returns a document and raises the context's `error` event exactly once, carrying the original exception.

One thing the sync path cannot fix: when the scripting service really is asynchronous, `ParseDocument` returns before the script has run, so the error is tracked after the call returns (measured: 0 tracked errors immediately after `ParseDocument`, 1 after 300 ms, and which of the two you see varies between runs). That is inherent to asking for a synchronous parse of an asynchronous script; the point here is only that the failure is no longer lost.

**The `readyState` half turned out to already hold, so there is no code for it.** `Document.FinishLoadingAsync` already sets `Interactive` before the deferred scripts and `DOMContentLoaded`, and `Complete` after them — `Complete` is in fact unreachable without passing through `Interactive`. Observed sequence on `devel` for `<body><script type='c-sharp'>…</script><p>text</p>`, identical for the sync and the async parse and with or without a scripting service:

```
readystatechange(Interactive) -> DOMContentLoaded -> readystatechange(Complete) -> load
```

So my third bullet in #1309 ("the document advances straight to complete during the parse") was simply wrong, and I have said so on the issue. The sequence is now pinned by a test instead, and `ReadyState` stays non-public exactly as you asked.

## Tests

Six new tests in `Library/BasicScripting.cs`, plus a `FailingScriptEngine` mock whose task faults only after yielding.

Run against unfixed `devel` first, as they should be: **four fail there** — the two synchronous ones with *"The browsing context did not track any error"*, the two asynchronous ones by `InvalidOperationException` escaping `ParseDocumentAsync`. The remaining two are characterization tests for the two premises that already held (`FailingScriptingServiceIsTrackedInsteadOfFaultingTheParse_Issue1309`, `ReadyStateBecomesInteractiveBeforeDomContentLoaded_Issue1309`); they pass on `devel` and are here to keep both contracts from regressing. All six pass with the change, on `net10.0` and on `net472`.

Full suite, Release, `net10.0`: 3798 passed / 37 failed / 2 skipped with `prefetched=false`, and 3796 / 37 / 4 with `prefetched=true`. The 37 are the `Library/FormSubmit.cs` (32) and `Library/ContextLoading.cs` (5) tests that talk to `http://anglesharp.azurewebsites.net/`, which is answering `403` at the moment; they are unrelated to this change.

## Left out

The location-setter half of #1309 — `location.host`/`pathname`/`port`/`protocol`/`search` starting a navigation on the setter's thread through the fire-and-forget `async void` in `Document.LocationChanged`. You did not agree to it, and it is arguably a host concern rather than a parser one, so it is not touched here. Happy to send it separately if you ever want it.

Fixes #1309 — although if you would rather keep the issue open for that remaining half, feel free to drop this line when merging.

https://claude.ai/code/session_01NqCcJrL3MJecCPRBMQsZyC

### DIFF
--- a/src/AngleSharp.Core.Tests/Library/BasicScripting.cs
+++ b/src/AngleSharp.Core.Tests/Library/BasicScripting.cs
@@ -1,10 +1,12 @@
-namespace AngleSharp.Core.Tests.Library
+﻿namespace AngleSharp.Core.Tests.Library
 {
     using AngleSharp;
     using AngleSharp.Browser;
     using AngleSharp.Core.Tests.Mocks;
     using AngleSharp.Dom;
     using AngleSharp.Html.Dom;
+    using AngleSharp.Html.Dom.Events;
+    using AngleSharp.Html.Parser;
     using AngleSharp.Io;
     using AngleSharp.Text;
     using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


lahma:
CI triage: both jobs report exactly **37 failures**, all in `Library/FormSubmit.cs` and `Library/ContextLoading.cs`, i.e. the tests that talk to `http://anglesharp.azurewebsites.net/`. Everything else passes (3798 on `net10.0`, 3793 on `net462`/`net472`).

They are not from this change: PRs #1310 and #1314 — untouched branches over completely different code — fail with the identical annotations (`FormSubmit.cs#621`, `#687`, …, all `NullReferenceException` on `QuerySelector("#input")` because the echo page never comes back). The host answers `403` from here as well. The same 37 fail on an unmodified `devel` checkout locally.

The six new tests pass on `net10.0`, `net472` and `net462`.

FlorianRappl (approved):
LGTM!

The web app quota is currently exceeded.

PR #1316, opened 2026-09-03
Walk CSS combinators through a struct cursor instead of a delegate

meziantou:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

`CssCombinator` exposed a `Func<IElement, IEnumerable<IElement>> Transform` implemented by
`yield return` methods, and `ComplexSelector.MatchCascade` invoked it **once per candidate
element per combinator position**. Every combinator step therefore allocated a compiler-generated iterator on the hot selector-matching path.

This replaces the delegate with a `CssCombinatorKind` enum plus a `CombinatorCursor` struct holding a hand-rolled state machine, so the traversal lives on the stack:

- `Child` / `Deep` / `AdjacentSibling` / `Namespace` reach at most one element.
- `Descendant` walks `ParentElement` through a stored field.
- `Sibling` keeps the existing single-scan index walk (the non-quadratic path added
  earlier), now as an index field rather than an iterator.
- `Column` stays the only kind that materialises a list, since `||` needs the table rows
  collected in a defined order.

`Transform` was only ever consumed by `MatchCascade`, so the change is contained. The
column-cell helpers moved out of the nested `ColumnCombinator` into `CssCombinator.GetColumnCells`; while moving them I folded the duplicated colspan parsing
into one `GetColumnSpan` and normalised `int` to `Int32` per the project convention.

### Measurements

Allocations, 16k-element document, `GC.GetAllocatedBytesForCurrentThread`, net10.0:

| Selector | before | after | ratio |
|---|---|---|---|
| `p` *(control, no combinator)* | 131,476 B | 131,476 B | 1.00 |
| `div p` | 467,476 B | 131,476 B | 0.28 |
| `section div p` | 803,476 B | 131,476 B | 0.16 |
| `div > p` | 419,476 B | 131,476 B | 0.31 |
| `p + p` | 353,916 B | 65,916 B | 0.19 |
| `p ~ p` | 497,916 B | 65,916 B | 0.13 |
| `body div p span` *(0 matches)* | 112,148 B | 148 B | 0.001 |

Every multi-part selector now allocates exactly what the combinator-free control does; the
remainder is the result list, and combinator traversal contributes nothing.

| Selector | before | after | Ratio | Alloc Ratio |
|---|---|---|---|---|
| `.note` *(control)* | 17.72 us | 17.13 us | 0.97 | 1.00 |
| `div` *(control)* | 16.05 us | 15.35 us | 0.96 | 1.00 |
| `div p` | 25.88 us | 22.13 us | 0.86 | 0.19 |
| `div > p` | 22.65 us | 19.43 us | 0.86 | 0.22 |
| `div p a` | 32.46 us | 27.83 us | 0.86 | 0.02 |
| `div + p` | 52.27 us | 46.04 us | 0.88 | 0.04 |
| `div ~ p` | 55.37 us | 53.73 us | 0.97 | 0.16 |
| `h1#title + div > p` | 42.18 us | 39.21 us | 0.93 | 0.007 |

Please read the controls before the results. `div` and `.note` contain no combinator, so
both builds run identical code there, yet they measure 0.96 / 0.97 — a systematic ~3-4%
bias. Netting that out, the honest timing gain is roughly 11% for the descendant/child
selectors, ~9% for `div + p`, ~4% for `h1#title + div > p`, and parity for `div ~ p`. The
allocation column is deterministic and needs no such hedge.

### DIFF
--- a/src/AngleSharp.Core.Tests/Css/CssSelector.cs
+++ b/src/AngleSharp.Core.Tests/Css/CssSelector.cs
@@ -58,6 +58,20 @@ namespace AngleSharp.Core.Tests.Css
         }
 
         [Test]
+        public void DeepCombinatorCrossesTheShadowBoundary()
+        {
+            var document = "<div id='host'></div><div id='other'></div>".ToHtmlDocument();
+            var host = document.QuerySelector("#host");
+            var shadowRoot = host.AttachShadow(mode: ShadowRootMode.Open);
+
+            shadowRoot.InnerHtml = "<p id='inner'>text</p>";
+
+            Assert.AreEqual(1, shadowRoot.QuerySelectorAll("#host >>> p").Length);
+            Assert.AreEqual("inner", shadowRoot.QuerySelectorAll("#host >>> p")[0].GetAttribute("id"));
+            Assert.AreEqual(0, shadowRoot.QuerySelectorAll("#other >>> p").Length);
+        }
+
+        [Test]
         public void StrangeDashSelector()
         {
             var source = @"<ul>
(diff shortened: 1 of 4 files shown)


FlorianRappl (approved):
LGTM!

PR #1322, opened 2026-09-08
Fix DOM doctype name validation

### LINKED ISSUE #1320: CreateDocumentType rejects valid DOM doctype names
### Prerequisites

- [x] Can you reproduce the problem in a MWE?
- [x] Are you reporting to the correct repository?

### Description

`IImplementation.CreateDocumentType` validates a doctype name as an XML name and qualified name. The current DOM Standard uses a separate doctype-name rule, including acceptance of the empty string.

Independently reproduced with NuGet AngleSharp 1.8.0 and current `devel` at `1b8579bea408b816891c61d047eb80152fc4f9af`, on macOS arm64 with .NET SDK 10.0.400 / runtime 10.0.11.

This belongs in AngleSharp.Core: the public DOM factory itself rejects the input in a standalone C# program, without a JavaScript engine or CSS plugin. Downstream context: https://github.com/sebastienros/jint/issues/3950. This report concerns only doctype names; element and attribute names have different rules.

### Steps to Reproduce

Create a .NET console app with `<PackageReference Include="AngleSharp" Version="1.8.0" />` and run:

```csharp
using AngleSharp.Dom;
using AngleSharp.Html.Parser;

var document = new HtmlParser().ParseDocument("<!doctype html><body></body>");

foreach (var name in new[] { "", "1foo", "edi:{", "a:", "a:b:c", "/", "a/b" })
{
    try
    {
        var type = document.Implementation.CreateDocumentType(name, "public", "system");
        Console.WriteLine($"'{name}': OK '{type.Name}'");
    }
    catch (DomException error)
    {
        Console.WriteLine($"'{name}': {error.Name}: {error.Message}");
    }
}
```

### Expected Behavior

All seven calls succeed and preserve the supplied name, public ID, and system ID. Colons and slashes are allowed; namespace validation does not apply.

Names containing any ASCII whitespace (tab, LF, FF, CR, space), U+0000, or `>` must still throw `DomException(DomError.InvalidCharacter)`, corresponding to DOM's InvalidCharacterError. Vertical tab and non-breaking space are allowed because they are not ASCII whitespace.

References:
- [DOM valid doctype name](https://dom.spec.whatwg.org/#valid-doctype-name)
- [DOM createDocumentType](https://dom.spec.whatwg.org/#dom-domimplementation-createdocumenttype)
- [WPT DOMImplementation-createDocumentType.html](https://github.com/web-platform-tests/wpt/blob/master/dom/nodes/DOMImplementation-createDocumentType.html)

### Actual Behavior

Identical results on 1.8.0 and the current source:

```text
'': InvalidCharacter: Invalid character detected.
'1foo': InvalidCharacter: Invalid character detected.
'edi:{': InvalidCharacter: Invalid character detected.
'a:': Namespace: The operation is not allowed by namespaces in XML.
'a:b:c': Namespace: The operation is not allowed by namespaces in XML.
'/': InvalidCharacter: Invalid character detected.
'a/b': InvalidCharacter: Invalid character detected.
```

Control probes for `html`, `:`, and `:a` already succeed. ASCII whitespace, U+0000, and `>` already fail correctly.

Replace the XML/qualified-name checks in this factory with the doctype-specific rule, preserving the existing .NET null argument check. No change to shared XML-name helpers is needed.

Checked the repository FAQ and searched open/closed issues and PRs on 2026-09-08 for `createDocumentType`, `"doctype name"`, `IsXmlName`, `"name validation"`, and `"Invalid character"`. No duplicate was found; the SVG parsing and attribute-name reports describe different paths.

sebastienros:
# Types of Changes

## Prerequisites

## Contribution Type

- [ ] New feature
- [ ] Breaking change
- [x] I have updated the documentation accordingly (changelog and linked issue)
- [x] All new and existing tests passed (excluding the seven cases not executed, noted below)

## Description

Fixes #1320.

`CreateDocumentType("", "", "")`, slash-containing names, and names such as `a:b:c` currently fail XML-name or namespace validation. Apply the [DOM doctype-name rule](https://dom.spec.whatwg.org/#valid-doctype-name) directly in this factory: accept these names, while rejecting ASCII whitespace, U+0000, and `>` with `DomError.InvalidCharacter`. Preserve the existing .NET null-argument check and public/system identifiers.

Shared XML-name helpers and element/attribute validation are unchanged. Regression tests cover empty, colon, slash, Unicode and whitespace boundaries, forbidden characters at different positions, owner document, and identifiers. The focused suite reproduced 14 failures before the fix and passes all 28 cases afterward.

Validation: the full `net10.0` suite passed: 3,906 tests passed, zero failed, and seven were not executed (six framework-gated cookie tests and one disabled external Google test, as recorded in the TRX). Library builds pass for `net8.0` and `netstandard2.0` with zero warnings/errors. Validation uses .NET SDK 10.0.400 on macOS arm64. Windows framework targets were not run locally.

Comparative benchmarks are deferred until a quiet host is available; no performance claim is made.


FlorianRappl:
Not convinced. Any real data to back this up? A website that breaks?

I see a ton of AI slop here just generated from some WPT transported tests - many of which may not make sense. It would be good to pre-assess first. The reason is simple. AngleSharp is out there for 10+ years and parsed hundreds of thousands of pages - all successful. While there certainly some issues - I am not sure that my time and attention is spend best on those tiny completely irrelevant spec-holes that actually might lead to other issues.

There are now 16+ new issues and I am not sure I want to spend time walking through them (especially if most titles already indicate its actually a non-issue - but then again, in that noise I miss real issues).

So can we triage first?

Please reduce noise. Many thanks!

PR #1344, opened 2026-09-12
Make every attribute write observable, and version the document

lahma:
Two changes to the same seam: an attribute change has to be observable, and a document should be able to say whether anything in it changed.

## 1. A `classList` write notified nothing

`Element.ClassList` wires `TokenList.Changed` to `UpdateAttribute(AttributeNames.Class, value)`, and `UpdateAttribute` was `SetOwnAttribute(name, value, suppressCallbacks: true)`. That single flag suppressed **three** things where only **one** needed suppressing:

| suppressed | needed? |
| --- | --- |
| the sync of the new value back into the `TokenList` that raised the change (`DefaultAttributeObserver` to `Element.UpdateClassList`) | yes - it re-enters the list mid-write |
| every `IAttributeObserver.NotifyChange` | no |
| `Owner.QueueMutation(MutationRecord.Attributes(...))` | no |

So `el.classList.add("x")` produced no mutation record and no observer notification, while `el.setAttribute("class", ...)` and `el.className = ...` produced both. DOM requires the attribute change steps and "queue a mutation record of attributes" for *every* attribute change, `DOMTokenList` included, and browsers do fire `MutationObserver` for `classList.add`. Downstream this is a `MutationObserver` with `attributeFilter: ["class"]` silently missing every class change, and a custom element never seeing a `classList` write.

**The suppression is narrowed rather than removed**, in two parts.

**The write itself becomes ordinary.** `UpdateAttribute` is now a plain `SetOwnAttribute`, so a token-list write runs the attribute change steps and queues a record exactly as `setAttribute` does.

**The sync reads the attribute instead of the value it was handed.** `Element.UpdateClassList` and its eight siblings took the value `IAttributeObserver.NotifyChange` delivered. That value is captured when the write starts, and it goes stale: `ConfigurationExtensions.With` **prepends** (`Common.Concat` yields the new element before the existing ones), so a host-registered observer runs *before* `DefaultAttributeObserver` and the built-in sync is always **last**. By the time it runs, an earlier observer may already have written something else — for an embedder, that is user script in a custom element's `attributeChangedCallback`.

Every rule over that delivered value has a hole, and they are the same hole from different sides — the list holding one value while the content attribute holds another:

| sequence during one dispatch | skip echo, never clear | + clear on mismatch | no guard at all | **read the attribute** |
| --- | --- | --- | --- | --- |
| nothing else writes | ok | ok | ok | ok |
| observer writes `zzz` | ok | **diverges** | **diverges** | ok |
| observer writes `zzz`, then `alpha` back | **diverges** | ok | ok | ok |
| observer writes the list itself, then `alpha` back | **diverges** | **diverges** | ok | ok |

So the sync takes no value at all now — the shape `DefaultAttributeObserver` already uses for the `img` `src`/`srcset`/`sizes` observers, for exactly this reason. The attribute is the truth, a token list exists to mirror it, and reading it converges however many writes were interleaved.

**What is left of the guard, and what it is not.** `TokenList` still records the string its own write serialized and skips an `Update` that matches it while it is still the one in flight (dropped as soon as any value arrives, saved and restored across nesting). That is now *only* an optimisation — it skips the split and its substrings and keeps the token instances the caller handed us. It is **not** a recursion guard: `Update` does not raise `Changed`, so an unguarded re-entry terminates on its own. Correctness rests on the re-read, not on the guard, and the code comment says so.

### Which other `UpdateAttribute` callers changed, and why none was left behind

There are ten call sites, all of them "an IDL member wrote the content attribute", and all of them are now observable:

| member | observer callback it re-enters |
| --- | --- |
| `Element.ClassList` | `Element.UpdateClassList` |
| `HtmlElement.DropZone` | `HtmlElement.UpdateDropZone` |
| `HtmlIFrameElement.Sandbox` | `HtmlIFrameElement.UpdateSandbox` |
| `HtmlLinkElement.RelationList` / `.Sizes` | `HtmlLinkElement.UpdateRel` / `UpdateSizes` |
| `HtmlUrlBaseElement.RelationList` / `.Ping` | `UpdateRel` / `UpdatePing` |
| `HtmlOutputElement.HtmlFor` | `HtmlOutputElement.UpdateFor` |
| `HtmlTableCellElement.Headers` | `HtmlTableCellElement.UpdateHeaders` |

**All nine sync callbacks changed the same way** — `UpdateClassList`, `UpdateDropZone`, `UpdateSandbox`, `UpdateRel` (on both `HtmlLinkElement` and `HtmlUrlBaseElement`), `UpdateSizes`, `UpdatePing`, `UpdateFor`, `UpdateHeaders` — so every reflected token list gets the fix, not just `class`. The guard lives in `TokenList` itself, which `SettableTokenList` derives from, so there is no per-list logic to get wrong either. Tests cover `class` and `sandbox` as representatives of the two base types. The tenth, `HtmlLinkElement.UpdateRel`, also rebuilds the link relation and calls `LoadRelation()` - so `link.RelationList.Add("stylesheet")` now re-runs link processing. That is deliberate: `link.Relation = "alternate stylesheet"` (the string setter) has always done exactly that, and the token list writing the same attribute without it was the inconsistency. `LoadRelation` is still gated on `Href != null` and `_relationLoaded`, so it cannot fire twice for the same href.

### One more member of the same family

`ElementExtensions.SetBoolAttribute` removed the content attribute with `RemoveNamedItemOrDefault(name, suppressMutationObservers: true)`. Setting a reflected boolean (`disabled`, `checked`, `required`, `readonly`, `multiple`, `hidden`, ...) was observable; clearing it was not. Fixed. As a side effect `link.IsDisabled = false` now reaches `HtmlLinkElement.UpdateDisabled` and actually re-enables the sheet.

### Deliberately left alone

- `NamedNodeMap.FastAddItem` and `IConstructableElement.SetAttributes` - the tree builder's bulk path. It has no duplicate check and no change steps by design; it is construction, not mutation, and the spec's attribute change steps do not run during parsing either.
- `Element.SetupElement` - already notifies observers for every attribute of a freshly built element, and queues nothing, which is correct: nothing changed.
- The `suppressObservers` argument on the four `Node.cs` sites - every `true` there is an internal step of a larger operation that queues one aggregated record of its own.

## 2. `Document.MutationVersion`

```csharp
public Int64 MutationVersion { get; }
```

A monotonic counter that changes whenever a node is inserted into or removed from the document's tree, an attribute of an element in it is added/removed/changed, or character data in it changes. It lets a consumer ask "may anything have changed since I last looked?" - the use case is a binding layer caching a live collection's result - without registering a `MutationObserver`, which is the only thing available today, costs a `MutationRecord` allocation per mutation and delivers asynchronously.

### Public or internal?

**Chosen: a public property on the public `abstract class Document`.** Reaching it from an `IDocument` is one cast: `((Document)document).MutationVersion`.

**Rejected: adding it to `IDocument`.** That is a breaking change for every implementor of the interface, and the usual escape hatch - a default interface member - is not available: `netstandard2.0` and `net462` are shipping targets and neither supports DIMs. Per the CONTRIBUTING versioning rules this would force a MAJOR, for a property.

**Also rejected: an `IDocument.GetMutationVersion()` extension** that falls back to `0` for a foreign implementation. A constant `0` reads as "nothing ever changes", so a cache built on it would be wrong rather than loud - the exact failure this feature exists to avoid. A cast that throws is the better diagnostic.

### Where it is advanced, and the honest enumeration

`DocumentExtensions.QueueMutation` looks like the one unconditional funnel for every childList, attribute and character-data mutation. **It is not, and putting the counter there would have made the promise dishonest.** These bypass it:

| path | inside the promise? | how |
| --- | --- | --- |
| `Node.InsertBefore` / `RemoveChild` / `ReplaceChild` / `ReplaceAll` (`suppressObservers: false`) | yes | reaches `InsertNode`/`AddNode`/`RemoveNode` |
| the same four with `suppressObservers: true` (`Document.Clear`, `Open`, the internal steps of `ReplaceAll` and `ReplaceChild`) | yes | same three mutators - the flag suppresses the *record*, not the write |
| `Node.InsertNode` / `AddNode` / `RemoveNode` called directly - the whole `HtmlDomBuilder` tree construction, `document.write`, foster parenting, the adoption agency, `HtmlTemplateElement.Content` | **yes, and this is why the counter is not on `QueueMutation`** | the counter is in those three methods |
| `Node.AppendText` / `InsertText` | yes | `CharacterData.Replace`, or `AddNode`/`InsertNode` |
| `setAttribute` / `removeAttribute` / `Attr.Value` / `className` / `id` / every reflected attribute / `dataset` | yes | `NamedNodeMap` mutators |
| `classList.add/remove/toggle` and the other token lists | yes (after change 1 they also queue a record) | `NamedNodeMap` mutators |
| `NamedNodeMap` with `suppressMutationObservers: true` | yes | the version is advanced before the suppression is consulted |
| `CharacterData.Replace` / `Data` / `NodeValue` / `TextContent` / `Append` / `Insert` / `Delete` | yes | `CharacterData.Replace` |
| `Element.AddAttribute` (public) and `IConstructableElement.SetAttributes` (the parser's bulk path, via `FastAddItem`) | yes | one step per element for the whole batch; the version has to move, not to count |
| `Element.CloneElement` to `FastAddItem` | n/a | the clone is detached; the insertion that attaches it advances the counter |
| `document.AdoptNode` | n/a | the removal from the old parent advances the old document, the insertion advances the new one; adoption alone changes neither tree |
| `Node.ChildNodes` setter (internal, **no callers**) | no | dead code; noted here so it is not resurrected without a step |

The one loose end is a **false positive**, not a false negative: a mutation of a node that is detached but still owned by the document - building a subtree before appending it, `Clone(deep: true)` - advances the counter. That is allowed by the contract as stated (a change means something *may* have changed) and costs a consumer a recomputation, never a stale answer. Making it precise would mean a connectivity check (a parent walk) on every parser insertion, which is not worth it.

So the promise holds: **an unchanged version means nothing observable changed.**

### Overflow and type

`Int64`, advanced with `++`. It does not wrap in any realistic document lifetime. The documented contract is nevertheless **equality only** - the value is opaque, a single API call may advance it more than once, and it may advance for a write that turned out to be a no-op - so that ordering is never depended on and the field could be narrowed later without breaking a consumer.

### Cost when nothing observes

One increment, no allocation, no branch beyond a null check. `Node` reaches it through `_owner ?? this as Document`, where the `as` only ever evaluates for a document node - the one node whose `Owner` is deliberately null. `Element.SetAttributes` takes one step for a whole attribute batch rather than one per attribute.

## Tests

`src/AngleSharp.Core.Tests/Library/AttributeObservability.cs` (16) and `src/AngleSharp.Core.Tests/Library/MutationVersion.cs` (21), both new files.

**Run against the unmodified library, 13 of the 16 attribute-observability tests fail** - `classList` add/remove/toggle producing no record, no record under `attributeFilter: ["class"]`, no `IAttributeObserver` notification, the `rel` and `sandbox` token lists, and both halves of the boolean-attribute removal. Of the three that pass on `devel`, one is the control (a `classList` write that changes nothing must stay silent) and the other two are the write-away-then-back pair, which cannot fail there because the observer never runs at all — those two fail against the intermediate versions of this change instead, which is where they earn their keep. Four of the sixteen assert the list and the content attribute agree after an observer has written the attribute during the dispatch; between them they fail against *every* rule in the table above except the one shipped. `MutationVersion.cs` does not compile against the unmodified library, which is the point.

The version tests cover an insert, a remove, a replace, an `innerHTML` write, a text change and append, `setAttribute`, an `Attr.Value` write, `removeAttribute`, a cleared boolean attribute, a `className` write, a `classList.add`, a `dataset` write, the construction path (`AddNode`, `AddAttribute`), and `document.write` from a script during parsing; that reads are silent; that a write which changes nothing is silent; and that two documents are independent.

Full suite: **0 failed** on `net10.0` (4074 passed, 1 skipped at the time of writing), in both `prefetched=false` and `prefetched=true`. The solution builds clean on `netstandard2.0`, `net462`, `net472`, `net8.0` and `net10.0` with `TreatWarningsAsErrors`. **No existing test depended on the old silence.**

## The cost, and what was done about it

Making `classList` observable means it now pays what `setAttribute` always paid. Measured paired A/B on an idle machine, that was **+526 %** and **54.69 KB → 171.88 KB** per thousand `classList.toggle` calls, with every other row flat — the right shape, since those paths already ran the attribute change steps.

Almost all of it was avoidable and predates this branch; `classList` simply became the first caller to notice:

- `Element.AttributeChanged` built the record **as an argument** to `QueueMutation`, so `MutationRecord.Attributes(...)` allocated on *every* attribute change and `QueueMutation` then discarded it unless an observer was registered. It now asks `Document.HasMutationObservers` first. `QueueMutation`'s own early-out already proved the call was a no-op in that case, so only the timing of the allocation changes, not the behaviour.
- `QueueMutation` snapshotted the observer list with a LINQ `ToArray()` *before* testing whether it was empty — and on `net462`/`net472` an empty `ToArray()` still allocates. The snapshot moved behind the guard and onto `List<T>.ToArray` via `MutationHost.Snapshot()`. It stays a snapshot: an observer callback may connect or disconnect while records are being delivered.
- `Document.AttributeObservers` was backed by an array but typed `IReadOnlyList<>`, so every `foreach` boxed an enumerator — 32 bytes per attribute change, in `AttributeChanged` and `SetupElement` both. Typed as the array it already is. Internal property, so no API change, and the dispatch is untouched: same observers, same order, same calls.

That restores `ClassListToggle` to **54.69 KB per thousand toggles**, its value before `classList` became observable, and takes `SetAttribute` to the same figure — below where it was on `devel`, since it was paying the record and the enumerator too. Allocation is deterministic so it is quoted here; **the timing re-measurement is the reviewer's**, on an idle box, not mine.

Each of the three is its own commit, so any of them can be dropped or split out without disturbing the conformance fix.

## Known, separate, not in this PR

`Node.cs` builds a `new NodeList()` for its childList records ahead of the same `suppressObservers` guard, and `CharacterData.Replace` builds its record unconditionally — which puts an allocation on the parser's text path. Both predate this branch and are left for their own change.

`SettableTokenList.Value { set => Update(value); }` never raises `Changed`, so setting `.Value` on any of the seven `ISettableTokenList` members (`Sizes`, `Sandbox`, `Ping` on anchor and area, `DropZone`, `HtmlFor`, `Headers`) updates the token list but **never writes the content attribute at all**. That is a different defect from this one - the write does not happen, rather than happening silently - and is left for its own change.

## To be measured

`src/AngleSharp.Benchmarks/MutationBenchmark.cs` is new: an `AppendChild` loop, a `SetAttribute` loop, a `ClassListToggle` loop (change 1's added notification cost) and a `SetCharacterData` loop, against a `CreateElements` baseline that is also the control - it allocates the same elements without touching a tree, an attribute or character data, so it must not move. `[MemoryDiagnoser]` is on and each mutating row owns its own document, built in `[GlobalSetup]`, so one row's residue cannot decide another's cost.

```
dotnet run -c Release --project src/AngleSharp.Benchmarks -f net8.0 -- --filter "*MutationBenchmark*"
dotnet run -c Release --project src/AngleSharp.Benchmarks -f net8.0 -- --filter "*ElementCreationParserBenchmark*"
```

The second one is the row that matters most: the counter is advanced from `Node.AddNode`, which is the parse hot path. **No numbers are quoted here** - the machine these were written on was shared, so any figure taken on it would be contaminated. The rows were executed once with `--job short` only to prove they run.


FlorianRappl (commented):
We have a conflict and this one is a bit dodgy in multiple areas. It really needs to show performance numbers to justify itself.

lahma:
Thanks — that is fair on all three counts, and two of them are mine to fix rather than to argue.

**The conflict is gone.** Rebased onto `devel` past #1345; both sets of changelog bullets kept.

**This should have been a draft.** My standing rule now is that a pull request on a repository I don't maintain stays a draft until I have reviewed it myself and it has its numbers. This one went out before that rule existed, and it landed in your queue in exactly the state you should not have had to look at. Sorry for that.

**Numbers are being measured now** — but I want to be precise about what they can and cannot justify here, because I think that is the root of the "dodgy" feeling:

- `Document.MutationVersion` is a **cost** question, and numbers settle it. The counter sits on `Node.AddNode`, which is the parser's hot path, so `ElementCreationParserBenchmark.ParsePage` is the row that matters, alongside the new `MutationBenchmark`. If the increment is visible there, the feature is not worth it and I will withdraw that half.
- The `classList` and `SetBoolAttribute` changes are a **conformance** question, and numbers cannot justify them at all — only show they cost nothing. Today `el.classList.add('x')` runs no attribute change steps and queues no mutation record, while `el.setAttribute('class', …)` and `el.className = …` do both; browsers fire `MutationObserver` for all three. Clearing `disabled` was silent where setting it was observable. Those are defects whether or not anything is faster.

**Which is the real problem with this PR: it is two changes.** A behaviour change to when observers fire, and a new public member. You are being asked to accept both at once, on the strength of one description. **I would rather split it** — a conformance PR (`classList` + the boolean-attribute clear, with the observability tests, no new API) and a separate `MutationVersion` PR that lives or dies on its benchmark row. Say the word and I will do that; the branch is already three commits along those lines.

In case it is useful before you decide, the two things in here that took the longest to get right, because they are where I would look first if I were reviewing it:

1. **`ConfigurationExtensions.With` prepends** (`Common.Concat` yields the new element first), so a host `IAttributeObserver` runs *before* `DefaultAttributeObserver` and the built-in token-list sync is always last — and it was being handed the value captured when the write started, which by then may be stale. Guarding that value could not be made correct; the fix is that `UpdateClassList` and its eight siblings now **re-read the content attribute**, which is the shape `DefaultAttributeObserver` already uses for `img`'s `src`/`srcset`/`sizes`. There is a four-case agreement test for it, and each intermediate attempt is its own commit with the test that killed it.
2. **The version deliberately does not sit on `QueueMutation`**, which was my first instinct and is wrong: the parser builds the tree through `Node.AddNode`/`InsertNode`/`RemoveNode` and queues no record, so a counter there would have been blind to `document.write`. It is advanced from the mutators, and the one imprecision left is a false positive (a detached-but-owned clone advances it) — which costs a consumer a recomputation and can never give it a stale answer.

Suite is 4072 passing in both `prefetched` modes, and **no existing test depended on the old silence**. If there is a specific area you would rather I cut or defend, name it and I will — I would rather land the conformance half cleanly than push the whole thing through.

FlorianRappl:
I agree that it's two problems, but maybe let's look at the performance numbers first before acting (e.g., splitting this into two PRs - we can always do that later).

lahma:
Understood — numbers first, and we decide about splitting afterwards. Rebased onto `devel` past #1345 and #1346; the conflict was the changelog again and both sets of bullets are kept.

What I'll measure and post here:

- **`ElementCreationParserBenchmark.ParsePage`** — the row that matters most for the counter, since it sits on `Node.AddNode`, which is the parser's hot path. If the increment is visible there, `MutationVersion` is not worth it and I'll withdraw that half.
- **`MutationBenchmark`** — `AppendChild`, `SetAttribute`, `ClassListToggle`, `SetCharacterData` against a `CreateElements` control that touches no tree, attribute or character data, so it should not move at all.

Allocation column included, paired A/B with the order alternated, default job, on an idle machine. One caveat I'd rather state up front: the `classList` and boolean-attribute changes make observers fire where they previously did not, so on a document that *has* an observer registered they are expected to cost something — the benchmark measures the no-observer case, which is the one every existing consumer is in today.

lahma:
## Numbers, as asked — and they justify your suspicion

Paired A/B, 6 rounds with the order alternating every round, default job, idle machine, both arms from worktrees differing only by this branch. `MutationBenchmark`:

| row | median Δ | range over 6 rounds | allocation before → after |
| --- | ---: | --- | --- |
| **`ClassListToggle`** | **+526.5 %** | +450.9 … +637.7 (6/6) | **54.69 KB → 171.88 KB** |
| `AppendChild` | +1.8 % | 3/6 | unchanged |
| `SetAttribute` | +0.4 % | 3/6 | unchanged |
| `SetCharacterData` | +1.1 % | 4/6 | unchanged |
| `CreateElements` *(control)* | +1.7 % | 4/6 | unchanged |

Everything that already ran the attribute change steps is flat. `classList` did not, and now does — so the conformance half works, and it costs **6×** plus triple the allocation on one of the hottest operations a page performs. That is not acceptable as it stands, and I would not have known without your asking.

## Most of that cost is pre-existing and avoidable

`Element.AttributeChanged` constructs the record **as an argument**:

```csharp
Owner.QueueMutation(MutationRecord.Attributes(
    target: this, attributeName: localName, attributeNamespace: namespaceUri, previousValue: oldValue));
```

So `MutationRecord.Attributes(...)` allocates on **every** attribute change, and only then does `QueueMutation` test `observers.Length > 0` and discard it. `QueueMutation` additionally does `document.Mutations.Observers.ToArray()` per call. None of this is new — `setAttribute` has always paid it — but `classList` used to skip the whole path, so this branch turns a free operation into one that allocates a record nobody reads.

I am fixing that in this branch: build the record only when the document actually has a mutation observer registered. `QueueMutation`'s own early-out already proves the call is a no-op otherwise, so it changes no semantics — only when the allocation happens. I will re-measure the full six rounds and post the table.

**On splitting**: with this in hand I think the shape of the answer is clearer. The `classList` and boolean-attribute changes are conformance and stand on the specification; `Document.MutationVersion` is a new public member that has to earn itself on `ElementCreationParserBenchmark.ParsePage`, which I will measure separately since the counter sits on `Node.AddNode`. Happy to split along that line whenever you prefer — say the word and I will; otherwise I will bring both halves back with their numbers and you can decide then.

One pre-existing item I am deliberately **not** touching here: `Node.cs` builds `new NodeList()` for childList records before its `!suppressObservers` guard too, the same shape. It is on `devel` today and belongs in its own change.

lahma:
## Corrected numbers after the allocation fixes

Same protocol — paired A/B, 6 rounds, order alternating, default job, idle machine, arms differing only by this branch.

| row | before the fixes | **now** | allocation base → candidate |
| --- | ---: | ---: | --- |
| `ClassListToggle` | +526.5 % | **+263.8 %** (6/6) | **54.69 KB → 54.69 KB** (was 171.88) |
| **`SetAttribute`** | +0.4 % | **−30.6 %** (0/6) | **171.88 KB → 54.69 KB** |
| `AppendChild` | +1.8 % | −4.1 % (2/6) | unchanged |
| `SetCharacterData` | +1.1 % | −3.0 % (1/6) | unchanged |
| `CreateElements` *(control)* | +1.7 % | −0.2 % (2/6) | unchanged |

Three causes, one commit each so any of them can be reverted alone:

1. `Element.AttributeChanged` builds the record only when `Document.HasMutationObservers`. `QueueMutation`'s own early-out already proved the call was a no-op otherwise, so semantics are unchanged — only when the allocation happens.
2. The `Observers.ToArray()` moved behind that guard and off LINQ onto `MutationHost.Snapshot()`. It stays a snapshot: an observer callback may connect or disconnect mid-delivery, so the live list must not be walked.
3. **`Document.AttributeObservers` was typed `IReadOnlyList<>` over an array**, so every `foreach` boxed an enumerator — 32 bytes per attribute change, in `AttributeChanged` and `SetupElement`. That was exactly the residual: after (1) and (2) the row sat at 85.94 KB, which is 31.25 KB over 1000 toggles, 32 bytes each. Typing it as the array it already is closes it precisely.

## What the numbers now say, including the part that is still a cost

**`setAttribute` is 30 % faster and allocates a third as much, for everyone.** It had been paying an unconditional record and a boxed enumerator all along; `classList` simply made that visible by walking the same path.

**`classList` remains ~3.6× its old cost, and allocation is now identical to baseline** — so what is left is *work*, not garbage: the observer dispatch and the attribute change steps it previously skipped entirely. That is the price of the conformance fix, not an inefficiency I can remove; the old number was cheap because it did not do what DOM requires.

Whether that trade is right is your call, and the honest framing is that a page does far more `setAttribute` and `className` writes than `classList` toggles — but a framework that toggles classes in a loop pays this. I am happy to be told it is too much.

**On splitting**: `ElementCreationParserBenchmark` is measuring now, since `MutationVersion`'s counter sits on `Node.AddNode` and the parser path is where it has to justify itself. When that lands you will have both halves with their own numbers and a clean seam to cut on — the conformance half (1)–(3) plus `classList`, and the counter.

Two pre-existing follow-ups deliberately left out, both in the PR body: `Node.cs` builds `new NodeList()` ahead of its `suppressObservers` guard, and `CharacterData.Replace` builds its record unconditionally — the latter puts an allocation on the parser's text path, and it is one line with the new predicate whenever you want it.

lahma:
## The parser path, for `Document.MutationVersion`

This is where the counter has to justify itself: it sits on `Node.AddNode`/`InsertNode`/`RemoveNode`, which is how the parser builds every tree. Same protocol — paired, 6 rounds, alternating order, idle machine.

| row | median Δ | range over 6 rounds | allocation base → candidate |
| --- | ---: | --- | --- |
| `ParsePage` | **−1.60 %** | −4.77 … +2.06 (3/6) | 1,603.16 KB → **1,583.91 KB** |
| `ParseSmallDocument` | +3.30 % | −1.04 … +6.14 (5/6) | 9.72 KB → 9.73 KB |

**`ParsePage` — the representative document — shows no cost, and allocates slightly less** (the reduction is the record and enumerator fixes reaching the parser's own attribute writes, which more than pay for the increment).

**`ParseSmallDocument` I will not claim either way.** It reads +3.3 % with 5 of 6 rounds agreeing, but the range crosses zero and its allocation is unchanged at 9.7 KB, so there is no mechanism for a real cost there — a small parse runs a few dozen `AddNode` calls, which cannot be 3 %. I would rather flag it as unresolved than round it to "no change": if you want it settled I will run it at twelve rounds.

That caution is not theoretical. Elsewhere in this campaign a 3-round paired run produced a 95 % CI *excluding zero*, with full sign agreement, on three separate rows a change could not affect — one of them reading a field from its own object's shape slot. All three reversed at six rounds and one turned out to be faster. A percentile bootstrap over few paired observations reports a confident interval when it has no business having one.

So, both halves now have their numbers:

- **Conformance** (`classList` and the boolean-attribute clear, plus the three allocation fixes): `setAttribute` −30.6 % and a third of the allocation, `classList` ~3.6× and at baseline allocation — that residual is the work DOM requires and which it used to skip.
- **`Document.MutationVersion`**: no measurable cost on the parse it was most at risk of taxing.

Yours to decide whether that lands as one change or two; the seam is clean either way, and I am glad you asked for the numbers rather than the argument — they turned up a 30 % win on `setAttribute` that neither of us was looking for.

FlorianRappl:
I think everything that suppresses mutation observers should also not change the mutation count. If we just include this simple rule then I don't think there should be any slowdown here.

lahma:
That rule is easy to add, but I do not think it buys what you are expecting, and the measurements say why — the counter is not where the cost is.

**The counter is already free.** Three figures from the run above:

- `SetAttribute` bumps it on **every** call and is **−30.6 %** (and allocates a third as much).
- `ParsePage` bumps it once per node — the parser's hot path — and is **−1.6 %**, i.e. no cost, with *less* allocation than before.
- `ClassListToggle`'s allocation is back to **54.69 KB, exactly the `devel` figure**.

Since allocation on that row is now identical to baseline, its residual is *work*, not garbage — and the work is the observer dispatch and attribute change steps that `classList` writes previously skipped altogether. An `Int64` increment cannot be 263 %, and `SetAttribute` getting faster while bumping the same counter is the control that proves it.

**What `suppressObservers: true` means here is narrower than the name suggests.** There are five call sites, and every one is an *aggregate-record* pattern rather than an invisible mutation — `ReplaceAll` removes and inserts each child with the flag set and then queues **one** `ChildList` record for the whole replacement (`Node.cs:264-279`); `ReplaceChild` and the fragment path in `InsertBefore` do the same. The flag says "do not report this step, the enclosing operation reports the whole thing". So a rule keyed on it would have to be careful in two ways:

1. The version is advanced in the raw mutators (`InsertNode`/`AddNode`/`RemoveNode`), which sit **below** the flag. Skipping there without adding a bump at the aggregate `QueueMutation` would make `element.innerHTML = '…'` change the tree **without** changing the version — a silently stale cache, which is the one failure mode this member must never have.
2. If instead the bump moves to `QueueMutation`, the counter stops seeing the **parser** entirely: `HtmlDomBuilder` calls `AddNode` directly (`HtmlDomBuilder.cs:324, 518, 3342, 3362, 3370`) and queues no record at all, so `document.write` and every initial parse would become invisible. That was the finding that made me put the counter on the mutators rather than on `QueueMutation` in the first place.

**So the real question is which contract you want**, and both are defensible — it is your library and I will implement whichever you say:

- **"Anything that changed the tree, an attribute or character data"** — what is there now. It sees the parser, so a consumer can cache a live collection and trust it. It bumps more than once for one logical operation (a replace bumps per child), which the documented contract already covers by saying to compare readings for equality and never for distance.
- **"Exactly what a `MutationObserver` would have been told"** — simpler to explain, and your rule follows from it directly. But it is a weaker guarantee than the name `MutationVersion` implies, and it removes the use case that motivated the member: a binding caching a live collection cannot use a counter that is blind to parsing. If you prefer this one I would want to say so plainly in the XML doc rather than leave a consumer to discover it.

Happy to split the PR at the same time — the conformance half stands on its own and does not depend on which of the two you pick.

FlorianRappl:
My point is not really about the counter (even though this included) but that we have mutation logic in paths triggered by the parser. This should not be. Any mutation logic should only be triggered by actual mutations.

`InnerHtml` explicitly sets `suppressObservers` to `false`. So your (2) is wrong and `InnerHtml` is supposed to increment that counter (as a matter of fact this is not `InnerHtml`, but `ReplaceAll` that will increment this counter - but as `ReplaceAll` is called from `InnerHtml` its the same net effect).

So I think it's not a choice between (1) and (2) - but rather that the parser goes in paths that have implicit no-mutation-change logic, while outside calls have implicit mutation-change-logic. Simplified: `AddNode` etc. never change mutation - its the call to `AddNode` (as you remarked) that makes this decision.

What I suggest is to merge this PR, but to have a follow-up with improvements to get the parse small document closer to the 1.8.1 baseline / net-effect have the `MutationVersion` at 0 (not set at 0 - but just unchanged) *after* parsing is done.

lahma:
That reframing is better than either option I put to you, and you are right that I got `InnerHtml` wrong.

You are correct that `InnerHtml` passes `suppressObservers: false`, so the outer `ReplaceAll` queues its aggregate record. What I should have said is narrower: the version is advanced in `AddNode`/`InsertNode`/`RemoveNode`, which sit *below* the flag, and `ReplaceAll`'s own internal steps pass `true` — so under "skip where suppressed", `innerHTML` loses its bump **unless the bump follows the record rather than the mutator**. That is your point, and following the record is the right answer.

And the underlying principle is the part I had not separated out: **mutation logic should not live in paths the parser drives**. I had treated "the counter must see the parser" as a requirement, when it is really a property of where I happened to put it. Framed your way — the call site decides, not `AddNode` — the parser simply does not make that decision, and `MutationVersion` comes out of a parse unchanged. That is a cleaner contract to document, too.

Happy to take your suggestion: **merge as it stands, and I will open the follow-up** moving the advance to the decision points and bringing the small-document parse back toward its 1.8.1 baseline. I will measure `ParseSmallDocument` and `ParsePage` on it rather than assert it, since `ParseSmallDocument` is the row I flagged as unresolved earlier and this is exactly what should settle it.

One consequence worth recording now, because it is my problem rather than yours and I would rather it be written down than discovered: with the counter unchanged across a parse, a consumer that caches a live collection keyed on it must treat a parser boundary as its own invalidation point — a script running at one boundary can cache a count that the parser then invalidates by appending, without the version moving. In a scripted host that boundary is explicit and cheap to hook, so this costs the consumer a line rather than the design anything. I mention it only so the follow-up's documentation says plainly what the member does and does not promise.

Thanks for pushing back twice on this one; both times it came out better.

FlorianRappl (approved):
LGTM!

Issue #1035, opened 2022-01-21
Custom element in `<head>`


LukeTOBrien:
I came upon this today, I am developing a Static Website Generator and I want to allow for layout templates like:

```html
<html>
  <head>
    <site-include type="partial" name="scripts"></site-include>
  </head>
</html>
```

So here you see the tag `site-include` denotes where a partial page should be included (in this case the scripts partial page in the head).  
In my C# code I then `QuerySelector` the `site-include` and `ReplaceWith` the partial.  
But or course the problem is that when I use the `HTMLParser` the above template gets turne into:

```html
<html>
  <head>
  </head>
  <body>
   <site-include type="partial" name="scripts"></site-include>
  </body>
</html>
```

Is there any option I can set to the `HTMLParser` to allow custom elements inside the `<head>`?  
I believe your parser is working as is correct for browsers, but it is not what I need for my purposes.

FlorianRappl:
Potentially we could add such an option, but then I'd ask why you use an HTML5 conform parser if what you input is not HTML5 confirm. Maybe XML is more what you look for?

Anyway - is there any chance to either use an existing (head-allowed) element or some notation such as the processing elements or comments?

(I don't want to change your SSG - just want to see if there is any other option besides changing a lib that is used underneath)

LukeTOBrien:
> why you use an HTML5 conform parser if what you input is not HTML5

I guess I didn't realise it would do this, I have used AngleSharp in the past and had the idea of using it with this project.  
An option to make the parser non HTML5 conformant would be the solution... But I could look at using XML like you say, the `System.Xml.Linq` namespace has similar manipulation methods, so I will give this a try.

> (I don't want to change your SSG - just want to see if there is any other option besides changing a lib that is used underneath)

At this stage I am just doing unit tests and experimenting to see what direction to go in.

LukeTOBrien:
Hello, just wanted to update here.  
XML didn't work because some HTML elements don't have closing tags which are required for XML.  
However I did find that [HTML Agility Pack](https://html-agility-pack.net/) worked in my case (I guess they don't validate for HTML5). The only thing missing is your fancy CSS selectors, but in this project I don't need them. (In my previous project I switched to AngleSharp because I was using CSS3 selectors like `nth-of-type`).

I will close he issue. Thanks

LukeTOBrien:
[...]

PR #1345, opened 2026-09-12
Three small DOM collection-read fixes: memoized Forms, single-pass id/name lookup, one less child scan

lahma:
## Background

This came out of profiling a JavaScript-interpreter workload (Jint) that repeatedly reads live DOM collections (`for (i = 0; i < c.length; i++) c[i]`). Two ETW captures over that workload pointed at four candidate hotspots in AngleSharp's collection-read paths. Three of them held up under inspection of this code; the fourth turned out to already be fixed here (see "Not included" below).

## What changed

**1. `Document.Forms` now memoizes its collection, like every sibling collection already does**

`IDocument.Forms` is annotated `[DomSameObject]` (`src/AngleSharp/Dom/IDocument.cs`), which means `document.forms` must be reference-stable across reads per the WebIDL extended attribute - exactly the contract `Images`, `Anchors`, `Scripts`, `Plugins`, `Commands`, `Links`, `All` and `Children` already honor via a `??=`-cached field. `Forms` alone built `new HtmlCollection<IHtmlFormElement>(this)` on every read, silently violating its own declared contract. This is a correctness fix with a perf side effect (one fewer allocation per read, and a stable wrapper identity for anything - a host binding, a weak table - keyed on it).

**2. The id/name named-getter lookup now enumerates its sequence once, not twice**

`CollectionExtensions.GetElementById<T>(IEnumerable<T>, string)` backs the named getter on every `HtmlCollection<T>` and on `HtmlFormControlsCollection` (`document.forms["x"]`, `document.images["x"]`, `form.elements["x"]`, ...). It matched `Id` in one pass over the sequence, then matched the `name` attribute in a second pass over the same sequence. For a lazily-built sequence - which `HtmlFormControlsCollection` always is (`root.GetNodes<HtmlFormControlElement>().Where(m => ReferenceEquals(m.Form, form))`, i.e. a full document walk filtered down to one form's controls) - each pass re-evaluates that whole upstream query. A lookup for a name that exists but has no id anywhere in the document walked the entire document twice for one answer.

Folded into one pass: keep scanning for an id match (returns immediately - an id match always wins) while remembering the first name match seen, falling back to it only once the sequence is exhausted. This preserves the exact original tie-break rule (any id beats every name; the first name match in document order wins among name matches) while enumerating once.

**3. The tag/class-name traversal helpers stop double-scanning each element's children**

The private recursive helpers behind `GetElementsByTagName` (both overloads) and `GetElementsByClassName` in `QueryExtensions.cs` guarded recursion with `if (element.ChildElementCount != 0)`. `Element.ChildElementCount` is an unconditional full loop over `ChildNodes` counting `NodeType == Element` - it does not short-circuit. So every element with element children paid for two full scans of its children at every tree level: once to count (the guard), once to match-and-recurse (the call the guard was gating).

Recursing unconditionally removes the guard's scan entirely: the recursive call is itself a single scan, a no-op for a childless element and no more expensive than the guard was for an element with only non-element children. Every element with element children now costs one scan instead of two.

I considered replacing the guard with a short-circuiting "has any element child" test (`Element.FirstElementChild` already is one) rather than dropping it, per the original framing of this class of fix. I rejected that: it still pays for a separate scan before the recursive call's own scan, so it's never cheaper than just recursing, and is strictly worse whenever the first child happens to be an element.

## Not included

A fourth candidate - giving `Element.GetAttribute(string)`'s case fold (`StringExtensions.HtmlLower`) an ASCII fast path that returns the original string when nothing would change - turned out to already be exactly what the code does (since #1035, Dec 2023). `HtmlLower(this StringOrMemory value)` scans for an uppercase ASCII character and returns the input unchanged if none is found; `ReadOnlyMemory<char>.ToString()` then returns the original string reference (not a copy) when the memory spans the whole string. Verified empirically: `ReferenceEquals("class".HtmlLower(), "class")` is `true`, and 100,000 calls with an already-lowercase input allocate 0 bytes. No change needed here.

## Tests

- Added `DocumentFormsReturnsTheSameCollectionInstanceOnRepeatedAccess` (fails against the prior `new HtmlCollection<IHtmlFormElement>(this)` implementation; verified this before committing the fix).
- Added `NamedCollectionAccessPrefersIdMatchOverAnEarlierNameMatch`, locking in the tie-break ordering the single-pass rewrite must preserve.
- No new test for change 3 (behavior-preserving); the full suite covers it.
- Full suite: `dotnet test -c Release src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj -f net10.0` → 4037 passed, 0 failed, 1 skipped (network-dependent). Ran the standard (`prefetched=false`) pass only; none of these changes touch text-source/tokenizer code, so the `prefetched=true` leg is not expected to differ.
- `dotnet build -c Release src/AngleSharp/AngleSharp.Core.csproj` (all target frameworks: `netstandard2.0`, `net8.0`, `net10.0`, `net462`, `net472`) - clean, no warnings.

## Benchmarks

Added `CollectionReadBenchmark` beside `QuerySelectorBenchmark` (`src/AngleSharp.Benchmarks/QuerySelectorBenchmark.cs`), `[MemoryDiagnoser]`, default job, kept as its own class since `QuerySelectorBenchmark` is parameterized over CSS selector text that these rows don't use:

- `QuerySelectorAllControl` - baseline; exercises the selector engine, untouched by this PR, should not move.
- `GetElementsByTagNameSweep` / `GetElementsByClassNameSweep` - over the existing `page.html` fixture.
- `DocumentFormsRead` - isolates the property-getter identity (not `.Length`, which pays for the live query regardless of getter caching).
- `FormElementsNamedLookup` - worst case for the id/name lookup (no `id` attribute anywhere in the document).

I ran this once with `--job short` locally only to confirm all five rows execute; I did not run it for real numbers (shared/contaminated environment) and none are quoted here.

**To be measured**: `dotnet run -c Release --project src/AngleSharp.Benchmarks -f net8.0 -- --filter "*CollectionReadBenchmark*"` on an idle box, default job.


FlorianRappl (approved):
LGTM!

PR #1346, opened 2026-09-12
Cache selector specificity and the specificity-sorted selector list

lahma:
## What the profile said

`Selectors.Specificity` (`src/AngleSharp/Css/Dom/Internal/Selectors.cs`) was `=> ComputeSpecificity()` — an unconditional recompute on every read. `CompoundSelector.ComputeSpecificity` and `ListSelector.ComputeSpecificity` are `O(n)` loops over their own selector list; `ComplexSelector.Specificity` loops all its combinators separately (it does not derive from `Selectors`). In `AngleSharp.Css`, `CssStyleRule.TryMatch` reads `.Specificity` on every successful match — once per matched rule per element — so a cascade over a large document repeats this recomputation across the entire match set.

Separately, `ListSelector.GetMatchingSelector` (`src/AngleSharp/Css/Dom/Internal/ListSelector.cs`) ran `_selectors.OrderByDescending(m => m.Specificity)` on **every call**, allocating an ordered enumerable, its buffer and a key array each time — on top of the specificity recompute above for every key.

## The mechanism

A selector is immutable once a `CssSelectorConstructor` has finished assembling it. `Add`/`Remove` (via the `Selectors` base class) and `ComplexSelector`'s `AppendSelector`/`ConcludeSelector` are the only mutators, and none of them run after parsing hands the selector back to its caller — confirmed by every caller in this repo (`CssSelectorConstructor.GetResult`, `Insert`, `InsertOr`) finishing the mutation before returning, and by the existing `SelectorPriorityTests`, which all read `.Specificity` only after `ParseSelector` returns.

## What I changed

- `Selectors.Specificity` now caches the computed `Priority` in a nullable field, computed lazily on first read.
- `Add`/`Remove` (and the indexer setter, unused today but part of the same surface) invalidate that cache through a new `protected virtual Invalidate()` hook, so a hypothetical premature read can never stick even though nothing exercises that today.
- `ComplexSelector.Specificity` gets the same treatment directly (it isn't a `Selectors` subclass), invalidated from `AppendSelector`/`ConcludeSelector`.
- `ListSelector.GetMatchingSelector` now caches the specificity-sorted array the same way, overriding `Invalidate()` to drop it too. This mirrors the existing precedent in `AngleSharp.Css`'s `CssStyleRule.ISelectorVisitor.List`, which already sorts its own selector list once instead of per match, for the identical reason.

## What I did not change, and why

- **`Selectors.Text` (`Stringify()`) stays uncached.** It allocates a fresh string per read, but nothing on the match path reads it — only parse-time error/serialization code (`CssSelectorConstructor`'s `:not()`/`:host-context()` error text) and the CSSOM's own `cssText` output do. Caching it would add a field with no exercised benefit, so I left it alone per this stream's brief ("judge it on whether anything hot reads it").
- **`GetMatchingSelector` has no internal caller in either `AngleSharp` or `AngleSharp.Css` today.** `CssStyleRule` bypasses it entirely via its own pre-sorted `_selectorList` (built once in `ISelectorVisitor.List`). It's only reachable through the public `IMultiSelector` API (exercised by `SelectorPriorityTests` in this repo). So this half of the change is not expected to move a cascade benchmark — it removes a real per-call allocation for any embedder that does call `GetMatchingSelector` directly, but there is no in-repo hot loop to demonstrate it against.

## Tests

- `dotnet test -c Release src/AngleSharp.Core.Tests/AngleSharp.Core.Tests.csproj -f net10.0` — 4035 passed, 1 skipped (pre-existing, network-dependent), 0 failed.
- No new tests added: this is a caching change behind an existing, already-tested public surface (`ISelector.Specificity`, `IMultiSelector.GetMatchingSelector`) — `SelectorPriorityTests` and the broader `Css/CssSelector*` suites already cover the values these caches return, before and after.

## To be measured

`dotnet run -c Release --project src/AngleSharp.Benchmarks -f net8.0 -- --filter "*QuerySelectorBenchmark*"` — I extended this with a multi-part compound selector (`a.url.fn[href]`) and a selector list (`h1, h2, h3, .toc, .note`) for broader `CompoundSelector`/`ListSelector` matching coverage, but note this is coverage, not a demonstration of the specificity cache: `QuerySelector`/`QuerySelectorAll` never read `.Specificity`. The change this PR is actually about is best measured in `AngleSharp.Css`'s `src/AngleSharp.Performance.Css/CssCascadeBenchmarks.cs` `ComputedStyle` row (see the companion `AngleSharp.Css` PR from this same stream), since that's where `.Specificity` is read on the hot path. I ran the extended `QuerySelectorBenchmark` rows once with `--job short` to confirm they execute; no numbers from that run should be treated as meaningful.

Part of a benchmark-gated performance campaign (stream U3). A companion PR against `AngleSharp.Css` (`perf/selector-matching`) fixes a related pseudo-class matching cost in the same area.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Released on Friday, September 11 2026
 - Fixed `Document.Forms` allocating a new collection instance per read, contradicting its own `[DomSameObject]` contract
 - Improved id/name lookup on collections (`document.forms["x"]`, `form.elements["x"]`, ...) to enumerate the underlying sequence once instead of twice
 - Improved `GetElementsByTagName` / `GetElementsByClassName` traversal to stop double-scanning each element's children
+- Improved selector specificity (`Selectors.Specificity`, `ComplexSelector.Specificity`) to be computed once instead of recomputed on every read
+- Improved `ListSelector.GetMatchingSelector` to sort its selectors once instead of on every call
 - Improved performance of parser attribute duplication check
 - Fixed vulnerability via SVG style serialization (GHSA-cgp3-27rh-pcp2)
 
--- a/src/AngleSharp.Benchmarks/QuerySelectorBenchmark.cs
+++ b/src/AngleSharp.Benchmarks/QuerySelectorBenchmark.cs
@@ -49,7 +49,14 @@ namespace AngleSharp.Benchmarks
             "#title",
             ".note",
             // No match at all: forces a complete traversal even for QuerySelector.
-            "div.no-such-class-anywhere"
+            "div.no-such-class-anywhere",
+            // A compound selector with several simple-selector parts chained on one element -
+            // exercises CompoundSelector.Match/Specificity across all four parts per candidate.
+            "a.url.fn[href]",
+            // A selector list (ListSelector) with branches of differing specificity - exercises
+            // ListSelector.Match plus, for anything that reads it (e.g. IMultiSelector.
+            // GetMatchingSelector, or a host cascade), Selectors.Specificity across every branch.
+            "h1, h2, h3, .toc, .note"
         };
 
         [Benchmark]
--- a/src/AngleSharp/Css/Dom/Internal/ComplexSelector.cs
+++ b/src/AngleSharp/Css/Dom/Internal/ComplexSelector.cs
@@ -16,6 +16,8 @@ namespace AngleSharp.Css.Dom
 
         private readonly List<CombinatorSelector> _combinators;
 
+        private Priority? _specificity;
+
         #endregion
 
         #region ctor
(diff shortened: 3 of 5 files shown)


FlorianRappl (approved):
LGTM!

There is a conflict that needs to be resolved before merging.

lahma:
Rebased onto `devel` — the conflict was with #1345, which has since merged. Both changelog bullets are kept, its three first; the benchmark file merged cleanly (my additions are entries inside `GetSelectors`, its `CollectionReadBenchmark` class sits after and is untouched).

Two notes I owe you:

- **This should have been opened as a draft.** My standing rule now is that a pull request on a repository I don't maintain stays a draft until I've reviewed it on my side; this one predates that.
- **Performance numbers are being measured and are not here yet.** `Selectors.Specificity` is read by `CssStyleRule.TryMatch` on every successful match, so the cascade is where this should show, and that is the row being measured — paired A/B, default job, allocation column, on an idle machine. I'll post the table here. One caveat I'd rather state than have you find: `ListSelector.GetMatchingSelector` has **no internal callers** in either repository (`CssStyleRule` uses its own pre-sorted `_selectorList`), so that half is reachable only through the public `IMultiSelector` API and there is no in-repo hot loop to demonstrate it against. If the specificity half measures as noise I'll say so rather than leave it open on a plausible story.

FlorianRappl (approved):
LGTM!

PR #1347, opened 2026-09-13
Keep mutation logic out of the tree construction path

lahma:
Follow-up to #1344, for this from the review:

> My point is not really about the counter (even though this included) but that we have mutation logic in paths triggered by the parser. This should not be. Any mutation logic should only be triggered by actual mutations. … have the `MutationVersion` at 0 (not set at 0 - but just unchanged) *after* parsing is done.

A parse now leaves `Document.MutationVersion` unchanged.

## What was wrong

`MarkMutated` sat on the raw operations rather than on the algorithms above them:

| Site | What it is |
| --- | --- |
| `Node.AddNode` / `InsertNode` / `RemoveNode` | the raw child list writes `HtmlDomBuilder` drives directly, once per node |
| `Element.AddAttribute`, `IConstructableElement.SetAttributes` | the raw attribute writes, once per element |
| `CharacterData.Replace` | reached from `Node.AppendText` / `InsertText`, i.e. the parser's text path |

So the counter advanced for every node, attribute batch and coalesced text run of a parsed document, and `page.html` came out of the parser several thousand steps from where it started.

## What changed

The advance moved up to the points that decided a mutation happened — `ReplaceAll`, `InsertBefore`, `RemoveChild`, `ReplaceChild`, where the `ChildList` record is queued. The other two families were checked on their merits:

- **`NamedNodeMap.NotifyChanged` was already correct** and is untouched. Only a set or a remove through the map reaches it; the tree builder goes to `FastAddItem`, which now does nothing but add.
- **`CharacterData.Replace` was not**, because `Node.AppendText` / `InsertText` reach it for the parser's text. The replace data steps keep the version and the record; the tree builder now goes through raw `AppendData` / `InsertData` instead. `AppendData` needs no range fix-up at all — an append at the end of the content can move no boundary — which is also why it is the cheap half.

**The advance follows the record, not the `suppressObservers` flag and not the mutator.** `ReplaceAll`, `ReplaceChild` and the fragment path in `InsertBefore` run their inner steps suppressed precisely so the enclosing operation reports once, and `InnerHtml` passes `suppressObservers: false` — so an `innerHTML` write *is* reported, once, by `ReplaceAll`. Reasoning from the inner steps alone concludes the opposite, so there is a test named for it.

It reaches for the owning document rather than the owner, because a document node reports a null `Owner`; `document.AppendChild(...)` and `document.RemoveChild(...)` must still version the document they change. Two tests pin that.

## One behaviour change to look at

The tree builder has two call sites that remove a node already in the tree — the adoption agency (`lastNode.Parent?.RemoveChild(lastNode)`) and the frameset error recovery that drops a body element. Both went through `IConstructableNode`, which passed `suppressObservers: false`, so they queued a mutation record and a transient observer. They sit directly beside the raw `AddNode` and `RemoveNode` the same algorithms use for everything else, and without them `<b><p>mis</b>nested</p>` still left the counter at 1 after a parse. **They are suppressed now.** A `MutationObserver` registered during a `document.write` parse no longer sees those two removals — it never saw any of the parser's *insertions*, so the removals were the odd one out, but it is a visible change and worth a look. The node removed steps and the `IPreRemove` hooks still run; only the observer bookkeeping is skipped.

## What I did not change

`ReplaceAll` still builds its two `NodeList`s unconditionally — its own loops iterate them, so they are not record-only. `Document.Clear()` and `Document.Open()` call it with `suppressObservers: true` and so no longer advance the version; both are documented as discarding the tree without notification, and `document.open` starts a parse, which the XML doc now names as a boundary a consumer has to invalidate on itself. The empty `addedNodes` a fragment `ReplaceChild` reports, and the null `previousSibling` it reports, are pre-existing and left alone.

## Allocations

Not a benchmark — `GC.GetAllocatedBytesForCurrentThread()` deltas, deterministic and load-independent, before vs. after on the same box:

| Operation | before | after |
| --- | --- | --- |
| `AppendChild` | 633 B | 433 B |
| append + remove | 480 B | 256 B |
| character data write | 224 B | 136 B |
| `SetAttribute` | 56 B | 56 B (already gated in #1344) |
| parse `page.html` | 2,017,713 B | 2,017,567 B |
| parse the small document | 11,536 B | 11,536 B |

That comes from `InsertBefore`, `RemoveChild`, `ReplaceChild` and `CharacterData.Replace` now asking `HasMutationObservers` before building a record `QueueMutation` would throw away — the question `Element.AttributeChanged` already asked. **The parse path barely moves on allocations**, because the tree builder rarely reached `CharacterData.Replace` in the first place (the tokenizer coalesces a text run, so `AppendText` usually creates a fresh `TextNode`). What the parse loses is the per-node bookkeeping itself, not an allocation.

## Tests

`MutationVersionTests` grew from 21 to 29 cases. The eight new or inverted ones all fail on `devel`:

- `VersionStandsStillForASmallParse`, `...ForAParseWithAttributesAndText` — a parse leaves it unchanged, through both the attribute batch and the text path
- `VersionStandsStillForAParseThatReparentsNodes` ×3 — the adoption agency, foster-parented text, the dropped body element
- `VersionStandsStillForAConstructionPathInsert` / `...Attribute` — inverted; these pinned exactly the behaviour being removed
- `VersionStandsStillForNodesWrittenByAScriptDuringParsing` — inverted; `document.write` feeds the tokenizer, so it is still a parse

Everything that was already there and is about a *mutation* passes unchanged, including `VersionMovesForAnInnerHtmlWrite`, which now carries a comment saying why it is not obvious. Added alongside: `VersionMovesForATextContentWrite`, `VersionStandsStillForAFragmentParse`, and the two document-node cases.

`dotnet test src/AngleSharp.Core.Tests` is green on `net10.0` in **both** `prefetched` modes (4083 tests) and on `net472`; the solution builds warning-free on all five TFMs.

## Documentation

- `AGENTS.md` gains the rule and the `suppressObservers` trap, in *The DOM*.
- `MutationVersion`'s XML doc now states that a parse does not advance it and that a consumer caching against it must treat a parser boundary as its own invalidation point.
- `docs/general/02-Features.md` said the counter "covers the parser's own construction of the tree"; corrected.
- `CHANGELOG.md` under 1.8.2 — the `Added` line is reworded rather than contradicted, since the feature has not shipped yet.
- `AGENTS.md` *Commands* also gains two things that cost me a wrong step while measuring this: there is no project or solution file at the repository root, so a bare `dotnet build` there is `MSB1003`, and the library's project file is `AngleSharp.Core.csproj` though its directory and assembly are both `AngleSharp`; and the `prefetched` bullet now says how to enter the second mode from the SDK, and that it is matched as the exact string `true`.
- `AGENTS.md` *Performance* gains the calibration rule the A/A below earned, since that section said to measure rather than assert without saying what makes a small measurement trustworthy.

## Measured

Paired A/B against `eb3925b9d` (the `devel` merge of #1344), **six rounds**, both benchmark classes in one
run, the order of the two arms alternating every round so run-order bias lands on each arm equally often.
The unit of observation is the per-round percentage difference; the interval is a percentile bootstrap over
the six, because a per-process distribution is often bimodal and a t-interval would assume it is not. A row
is a win only when the interval excludes zero. `net10.0`, `DefaultJob`, idle box, nothing else running.

| Row | median | 95 % CI | verdict | allocated, base → cand |
| --- | ---: | --- | --- | --- |
| `MutationBenchmark.SetCharacterData` | **−19.54 %** | [−20.56 %, −15.86 %] | faster | 203.13 KB → 117.19 KB |
| `MutationBenchmark.AppendChild` | **−10.29 %** | [−15.02 %, −7.73 %] | faster | 610.30 KB → 414.99 KB |
| `MutationBenchmark.SetAttribute` | +0.47 % | [−2.95 %, +1.97 %] | noise | 54.69 KB → 54.69 KB |
| `MutationBenchmark.ClassListToggle` | +0.16 % | [−3.62 %, +3.71 %] | noise | 54.69 KB → 54.69 KB |
| `MutationBenchmark.CreateElements` | +1.73 % | [+1.11 %, +3.53 %] | see below | 375.00 KB → 375.00 KB |
| `ElementCreationParserBenchmark.ParsePage` | −0.41 % | [−1.97 %, +0.88 %] | noise | 1583.91 KB → 1583.91 KB |
| `ElementCreationParserBenchmark.ParseSmallDocument` | +0.10 % | [−1.15 %, +1.91 %] | noise | 9.73 KB → 9.73 KB |

The per-round deltas, so the spread is visible rather than summarised:

```
SetCharacterData   -18.80  -12.92  -20.08  -19.40  -21.03  -19.67
AppendChild        -11.42   -9.16  -17.71   -6.52  -12.33   -8.93
SetAttribute        +2.03   -0.21   +1.91   +1.15   -1.99   -3.91
ClassListToggle     +0.24   -0.87   +0.09   -6.37   +4.53   +2.89
CreateElements      +1.32   +4.42   +1.19   +2.65   +1.02   +2.14
ParsePage           -0.43   -0.55   +1.41   -3.38   -0.38   +0.35
ParseSmallDocument  +0.49   -0.28   -2.00   +1.38   +2.45   -0.31
```

**The two rows that move are the two whose records are now gated**, and each moves on time and on
allocations together: `SetCharacterData` −42 % allocated, `AppendChild` −32 %. `SetAttribute` is the
control for that gating — it was already asking `HasMutationObservers`, from #1344 — and it does not move
on either axis; `ClassListToggle` is the second control and likewise does not. The allocation figures are
BenchmarkDotNet's `MemoryDiagnoser` column and agree with the `GC.GetAllocatedBytesForCurrentThread()`
deltas reported above.

**The parse path does not regress — and it does not improve either.** Both parser rows read as noise. An
earlier isolated six-round pair of the parser class alone put `ParseSmallDocument` at −1.79 %
[−2.22 %, −0.17 %]; it did not reproduce here, so the honest reading of the two runs together is *no
measurable change*. That is what this change needs the parse to be: the point was to take the bookkeeping
out of construction, not to make construction faster, and one increment per node is far below what this
box resolves on a row dominated by tokenizing.

### `CreateElements`, and why it is reported rather than claimed

`CreateElements` is the control row: `document.CreateElement("span")` in a loop. It reaches
`Document.CreateElement` → the element factory → `Element.SetupElement` (an empty loop for an element with
no attributes) and **none** of the methods this change touches — not `AddNode`, not `InsertBefore`, not
`RemoveChild`, not `AddAttribute`, not `CharacterData`. But it came out +1.73 % with an interval excluding
zero and all six rounds positive, which is too tidy to wave away, so I calibrated instead of arguing.

**A/A: two separately built worktrees of `eb3925b9d` against each other**, same six-round protocol, same
box, `diff -r` across `src/` empty. If the protocol is sound this is all noise, and it is:

| Row | A/A median | A/A 95 % CI | the A/B above | verdict |
| --- | ---: | --- | ---: | --- |
| `CreateElements` | −2.03 % | [−4.09 %, +0.94 %] | +1.73 % | **inside its own spread** |
| `AppendChild` | −2.93 % | [−5.97 %, +1.56 %] | −10.29 % | outside — real |
| `SetCharacterData` | +1.11 % | [−0.78 %, +2.60 %] | −19.54 % | outside — real |
| `SetAttribute` | −0.19 % | [−1.12 %, +1.32 %] | +0.47 % | control, both flat |
| `ClassListToggle` | −0.06 % | [−5.03 %, +1.99 %] | +0.16 % | control, both flat |

So `CreateElements` has a build-to-build spread of roughly ±4 % on this box — wider than the "+1.73 %
regression" the six-round bootstrap reported for it. **It is a false positive of the protocol, not a cost
of this change**, and the two rows that do move are the two that clear their own A/A envelope by a factor
of three and of seven. I have added that rule to the *Performance* section of `AGENTS.md`, since the
section told a contributor to measure without saying what makes a small measurement trustworthy.


FlorianRappl (approved):
LGTM!

PR #1348, opened 2026-09-13
Use IReadOnlyDictionary

mcraiha:
# Types of Changes

## Prerequisites

## Contribution Type

## Description

I assume the mappings should be immutable


FlorianRappl (changes requested):
This is a breaking change and therefore not desired. The description also does not contain any reason why this would be an improvement (e.g., better performance with a measurement to show the expected speed up).

While I agree that most use cases would not mind having it read only the breaking nature of the PR excludes it.

PR #1349, opened 2026-09-13
Advance document mutation version for native selector state

sebastienros:
**Closed for Jint's use case:** [sebastienros/jint#4075](https://github.com/sebastienros/jint/pull/4075) has merged Browser-owned invalidation with the existing AngleSharp packages. It tracks the calls Jint controls and keeps fresh queries for unclassified native writers and asynchronous attachment. Jint no longer needs this upstream mutation-version proposal. The native-API gaps documented below remain relevant to unrestricted native callers.

---

Native control writes can change a selector result while `Document.MutationVersion` stays unchanged. This advances the existing token for checkedness, selectedness, indeterminacy, value/dirty-value/custom-validity state, and focus. The focus transition advances it before callbacks run. No public interface member, observer, cache or layout implementation is added.

The companion stylesheet producer is AngleSharp/AngleSharp.Css#248.

Parser construction is handled by the separate AngleSharp/AngleSharp#1347 follow-up to #1344. This draft adds no notification to the parser's raw node/attribute/text operations. Control cloning copies validity state directly instead of calling a notifying user setter. An isolated integration of #1347 at `b19914329ca5332cbffb60fd387a935366b29c0f` with this draft passes all 41 mutation-version tests, including control parsing at version zero and preservation of user mutations. The Core parser fix remains a separate dependency; consumers must invalidate at parser boundaries, or keep geometry query-local while parsing can resume. No version is reset after parsing.

The blocked consumer is Jint.Browser's repeated geometry queries (sebastienros/jint#3698). With `input:checked { display: none }`, a native host assignment to `IHtmlInputElement.IsChecked` can change whether an element has a box between two same-turn reads. A cache keyed by the existing DOM version currently serves the old result.

`NativeCheckednessInvalidatesACacheWithoutMutationRecords` reproduces this using a native interface, not a wrapper: the selector result changes, serialized markup stays identical, and a document-wide observer receives zero records. Nine new regressions fail on the unmodified base `eb3925b9dfc0731cf249e82b13b7bb19a2022db8`; all 31 mutation-version cases pass with this change. The [rationale and reproducer](https://github.com/sebastienros/AngleSharp/blob/codex/selector-state-version/docs/general/MutationVersionSelectorState.md) explain the other covered states.

Why this belongs in Core: an external wrapper only intercepts callers that opt into it. Existing hosts can hold native interfaces and bypass it; the control implementations are internal and sealed. Replacing the element factory would mean replacing those controls and their DOM behavior. No DOM observer delivery/draining strategy can recover a record that does not exist. Scanning control state on every read can preserve correctness but loses the cheap unchanged-read requirement; DOM serialization cannot even detect checkedness. Under the supported native APIs, the existing mutation points are the narrowest place to provide the missing synchronous signal. Cache policy stays in the consumer. Stylesheets, render-device settings, custom selectors and host-owned state still require their own invalidation policy.

Validation: fresh Release Core suite, 4,084 passed / 1 skipped; library builds for netstandard2.0, net8.0 and net10.0 with no warnings. Windows framework execution was not available locally; the Linux and Windows CI jobs both passed on the updated draft.

Benchmarks use identical source on base `eb3925b9dfc0731cf249e82b13b7bb19a2022db8` and the draft: BenchmarkDotNet 0.15.8, default job with three launches, Release net10.0, Apple M4 Pro, macOS 15.7.9. Native-state rows (`SelectorStateMutationBenchmark`, the first five rows) use .NET 10.0.10 / SDK 10.0.302. The `ElementCreationParserBenchmark` rows use .NET 10.0.11 / SDK 11.0.100-rc.1.26413.103, in two before/after rounds with alternating order. Every comparison uses matching SDK/runtime versions within its row. Runs were sequential. State-write figures are per assignment (1,000 assignments/invocation); parsing and selector rows are controls. These are local producer measurements, with no layout-cache benefit included.

| Operation | Time before (mean ± error) | Time after (mean ± error) | Δ time | Allocation before | Allocation after | Δ allocation |
|---|---:|---:|---:|---:|---:|---:|
| SetChecked | 0.5114 ns ± 0.0038 ns | 0.7963 ns ± 0.0114 ns | +55.71% | 0 B | 0 B | 0.00% |
| SetSelected | 0.5281 ns ± 0.0072 ns | 0.7706 ns ± 0.0081 ns | +45.92% | 0 B | 0 B | 0.00% |
| SetValue | 0.4317 ns ± 0.0134 ns | 0.8033 ns ± 0.0120 ns | +86.08% | 0 B | 0 B | 0.00% |
| MatchChecked | 66.0413 ns ± 0.7050 ns | 69.2681 ns ± 1.1108 ns | +4.89% | 0 B | 0 B | 0.00% |
| ParseControls | 4,279.8407 ns ± 61.2546 ns | 4,171.5979 ns ± 49.9702 ns | -2.53% | 10408 B | 10408 B | 0.00% |
| ParsePage | 1,334.530 μs ± 25.9640 μs | 1,228.997 μs ± 22.7372 μs | -7.91% | 1583.94 KB | 1583.94 KB | 0.00% |
| ParseSmallDocument | 4.299 μs ± 0.1132 μs | 4.880 μs ± 0.1509 μs | +13.51% | 9.73 KB | 9.73 KB | 0.00% |

The final two rows above are parser round 1 (base first, then draft). Round 2 reverses the order (draft first, then base), with three launches per arm again:

| Operation | Time before (mean ± error) | Time after (mean ± error) | Δ time | Allocation before | Allocation after | Δ allocation |
|---|---:|---:|---:|---:|---:|---:|
| ParsePage | 1,286.610 μs ± 18.4267 μs | 1,232.404 μs ± 22.1108 μs | -4.21% | 1583.94 KB | 1583.94 KB | 0.00% |
| ParseSmallDocument | 4.527 μs ± 0.0663 μs | 4.483 μs ± 0.0422 μs | -0.97% | 9.73 KB | 9.73 KB | 0.00% |

`ParseSmallDocument` changed from +13.51% in round 1 to −0.97% in round 2; the initial slowdown did not reproduce. `ParsePage` measured −7.91% and −4.21%, respectively. Parser allocations are unchanged in both rounds. These two rounds are diagnostic, not a calibrated regression gate; they do not establish a parser speedup or prove zero timing overhead.

Δ = (after − before) / before × 100; negative means less time or allocation. Unchanged zero allocation is shown as 0.00%.

Error is BenchmarkDotNet’s reported 99.9% confidence-interval half-width; cross-process/environment variation limits small before/after comparisons. No concurrent local test or benchmark workload was run during measurement.

Reproduce on both worktrees (copy only the new benchmark class to the base):

```sh
RestoreSources=https://api.nuget.org/v3/index.json dotnet run -c Release --project src/AngleSharp.Benchmarks -f net10.0 -- --filter '*SelectorStateMutationBenchmark*' '*ElementCreationParserBenchmark*' --launchCount 3
```

Withdrawn in favor of the merged Jint consumer implementation linked above.


docs/general/MutationVersionSelectorState.md
```diff
@@ -0,0 +1,30 @@
+# Selector state and document mutation versions
+
+`Document.MutationVersion` lets a consumer cache derived results without observing or walking every node. That only works when native state that affects the result also advances the version. A checked input is a small counterexample to DOM-only tracking:
```
FlorianRappl, line 3: This whole thing should be something internal. Is there a reason to make this external / part of the docs?

FlorianRappl (commented):
I am actually not sure if we are going in the right direction here.

The whole mark mutated sounds like an anti-pattern to me. I know where we are coming from here, but the issue with all of that is that it's incomplete. The way we are currently introducing it essentially changes the code in multiple locations and still has some unclosed areas (e.g., I can still add pseudo classes and with the current changes selectors would be cached and not re-evaluated, even though those pseudo classes changed... - this just one of ~6 cases I can immediately think of).

Long story short: I am currently thinking about reverting the whole mutation version change. It's a bandaid that does not work if not reliably implemented (i.e., everywhere). But implementing it everywhere is cumbersome and chances are high we miss something. We need a better mechanism.

sebastienros:
Closing this proposal following the merge of [sebastienros/jint#4075](https://github.com/sebastienros/jint/pull/4075).

Jint now brackets the DOM/control mutations it invokes, handles parser and environment boundaries itself, and keeps fresh queries for unclassified native writers and asynchronous attachment. This resolves Jint's immediate cache-invalidation use case with the current AngleSharp packages. The native-interface gap documented here still applies to unrestricted native callers, but this upstream change is no longer required by Jint.

PR #1350, opened 2026-09-14
Walk the tree once per form.elements read instead of re-running a LINQ query

lahma:
`HtmlFormControlsCollection` held one deferred LINQ query and re-executed it on every member:

```csharp
_elements = root.GetNodes<HtmlFormControlElement>().Where(m => { ... m.Form ... });
...
public Int32 Length => _elements.Count();
public HtmlFormControlElement this[Int32 index] => _elements.GetItemByIndex(index);
```

So `form.elements.length` was a whole-document `NodeEnumerable` walk — a `Stack<EnumerationFrame>`, an
`OfType` iterator, a `Where` iterator and a closure — and `form.elements[i]` was another one. A loop over
`length` and `[i]`, which is what a page does, paid all of it twice per iteration.

This keeps only the root to walk and the form to match against, and drives all four members — the length,
the indexed read, the named read and enumeration — through one `struct` walker and one predicate. The
collection stays exactly as live as it was: nothing is memoized, no result is materialized, every read
still queries the tree as it is at that moment.

The walker is the technique already in `ElementTreeEnumerator` — parent links plus a stack of child
indices, so the stack holds value types and is bounded by the tree's depth rather than its width — with
the index stack rented from `ArrayPool<Int32>`, which is what takes a read to **zero** allocations.

**Why not reuse `ElementTreeEnumerator` itself**, since it is most of this: it yields its root (and a
`<fieldset>` is itself a form control, so it must not appear in its own collection), its element is
untyped and unfiltered, and it allocates its index stack with `new Int32[16]` per walk. Composing on it
would have been about a third of the code and left one array per read. Making *it* pooled would have
fixed `QuerySelector`/`QuerySelectorAll` too, but it would push a disposal contract into the selector
engine, where a missed or doubled `Dispose` hands one buffer to two walkers and corrupts both silently.
That is not a drive-by change, so it is not in this one.

## Semantics

Unchanged, and the equivalences are worth stating because they are what a reviewer would otherwise have
to re-derive:

- `NodeEnumerable` yields **descendants only**, never the starting node, which is what the new walker
  does — a fieldset is excluded from its own collection, as `TheControlsMustRootAtTheFieldsetElement`
  already pinned.
- Descending only into elements finds every element descendant, because a form control's ancestors inside
  an element-rooted subtree are always elements. That makes the walk equivalent to
  `NodeEnumerable.OfType<HtmlFormControlElement>()` without visiting character data.
- The named read reproduces `CollectionExtensions.GetElementById`'s rule exactly — an id match always
  wins, even over a name match seen earlier in tree order.
- The indexed read still throws `ArgumentOutOfRangeException` outside the range, as `GetItemByIndex` did.
- `input type=image` is still excluded; it is a listed element but a submit button rather than a member
  of `form.elements`.

## Tests

Eight new cases in `LiveCollectionTests`, written and run against the **unmodified base first** so they
are equivalence tests rather than tests of the new code. Then each was confirmed load-bearing by breaking
the implementation on purpose in five ways — dropping the image exclusion, letting a name match overwrite
an id match, yielding the root, rooting the walk at the form instead of the document element, returning
`null` instead of throwing, and memoizing `Length` — and each break fired on exactly its own property.
The memoizing break is caught by `HtmlFormLiveCollectionFollowsAppendRemoveAndReparent`, which is the one
that matters most here: it is what says the collection is still live.

`4075 -> 4083` cases, 0 failed, in **both** text-source modes (`prefetched=false` and `prefetched=true`).
The library builds warning-free on all five TFMs.

## Measured

Paired A/B against `eb3925b9d`, **6 rounds**, arms alternating every round, `DefaultJob`, idle box, each
arm from its own worktree — the baseline carrying this PR's new benchmark file untracked, so both arms
run the same rows and differ by exactly `HtmlFormControlsCollection.cs`. The median is of the per-round
percentage differences and the interval is a percentile bootstrap over them.

| Row | median | 95 % CI | allocated, base → cand |
| --- | ---: | --- | --- |
| `FormControlsCollectionBenchmark.FormElementsLength` | **−56.05 %** | [−57.10 %, −54.43 %] | 61.72 KB → **0 B** |
| `FormControlsCollectionBenchmark.FormElementsIndexed` | **−53.83 %** | [−56.00 %, −52.45 %] | 67.19 KB → **0 B** |
| `FormControlsCollectionBenchmark.FormElementsNamed` | **−52.25 %** | [−54.89 %, −51.80 %] | 67.19 KB → **0 B** |
| `FormControlsCollectionBenchmark.FormElementsEnumerate` | **−49.02 %** | [−50.40 %, −44.29 %] | 67.19 KB → 5,600 B |
| `CollectionReadBenchmark.FormElementsNamedLookup` | **−54.01 %** | [−55.21 %, −53.36 %] | 688 B → **0 B** |
| `FormControlsCollectionBenchmark.GetElementsByTagNameControl` | −0.93 % | [−6.03 %, +0.27 %] | unchanged |
| `CollectionReadBenchmark.QuerySelectorAllControl` | +0.94 % | [−2.85 %, +1.63 %] | unchanged |
| `CollectionReadBenchmark.GetElementsByClassNameSweep` | −2.68 % | [−4.19 %, +14.08 %] | unchanged |
| `CollectionReadBenchmark.GetElementsByTagNameSweep` | +0.25 % | [−2.66 %, +23.46 %] | unchanged |
| `CollectionReadBenchmark.DocumentFormsRead` | −0.87 % | [−5.57 %, −0.19 %] | unchanged |

`FormElementsNamedLookup` is **not a row this PR wrote** — it came with #1345 and hits this collection
through a different member. It reads the same magnitude independently, which is the closest thing to a
second opinion a single box can give.

`FormElementsEnumerate` keeps 5,600 B because enumerating through `IEnumerable<IHtmlElement>` boxes the
struct enumerator once per enumeration; that is the interface's cost, not the walk's, and the other three
rows show what the walk itself now costs. `DocumentFormsRead`'s −0.87 % is **not claimed**: its interval
only just excludes zero, which on this box is inside the band where a six-round paired bootstrap cannot
be trusted on its own.

## Follow-up, deliberately not in this PR

`HtmlFormControlElement.Form` is the larger remaining cost and it is untouched here. For a control
carrying `form="id"` it walks to the root and then does `root.ChildNodes.GetElementById(formId)` — the
**recursive** scan — once per candidate per read, and there is no id index anywhere in this DOM
(`Document.GetElementById` is literally that same scan). So `form.elements.length` is O(K·N) for K
explicitly-associated controls. The obvious pre-filter — a candidate with no `form` attribute must have
the form as an ancestor — needs care around `form=""`, a non-form id target and a disconnected root, all
of which `FormOwnerTests` already pins, and it touches files that currently have another PR open on them.
Happy to take it as a separate change.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Released on Friday, September 11 2026
 
 - Fixed `Document.Forms` allocating a new collection instance per read, contradicting its own `[DomSameObject]` contract
 - Improved id/name lookup on collections (`document.forms["x"]`, `form.elements["x"]`, ...) to enumerate the underlying sequence once instead of twice
+- Improved `form.elements` / `fieldset.elements` to walk the tree directly on every read instead of re-running a captured LINQ query, so a length, an indexed read, a named read or an enumeration allocates nothing
```
FlorianRappl: Too much detail:

> Improved performance of `form.elements` / `fieldset.elements` by walking the tree directly (#1350)

FlorianRappl (commented):
This is fine but the changelog should be less detailed - technical details are always to be found in the referenced issue or PR

CHANGELOG.md
lahma: Done, taken verbatim in 51540790b.

I applied the same rule to #1347 while I was at it (c6fcd757c) — its three lines in the same 1.8.2 block were the same shape, so rather than wait to be told twice:

- `Added `Document.MutationVersion`, a synchronous counter of DOM mutations that a parse leaves unchanged (#1344, #1347)`
- `Improved performance of HTML tree construction by keeping mutation logic out of it (#1347)`
- `Improved performance of DOM mutations by building a mutation record only when an observer will consume it (#1347)`

Happy to trim those further if they are still long for your taste.

FlorianRappl (approved):

PR #1351, opened 2026-09-15
Serialize only the case-sensitivity modifier that was specified

lahma:
`ISelector.Text` writes a case-sensitivity modifier that was never specified, and omits one that was:

```csharp
var parser = new CssSelectorParser();

parser.ParseSelector("[type=\"text\"]")!.Text;       // [type="text" i]
parser.ParseSelector("[align=\"center\" s]")!.Text;  // [align="center"]
```

CSSOM § 5.2 appends the modifier only "if the attribute selector has the case-sensitivity flag **present**" — present on the selector, i.e. specified. `[type="text"]` carries no flag; `type` merely happens to be one of the ~44 attribute names HTML matches ASCII case-insensitively. So the serialization is wrong on its own terms, before anyone reads it back.

Cause: `BaseAttrSelector.Modifier` read the resolved `StringComparison`, and `DefaultAttributeSelectorFactory` has already folded that HTML rule into it — by then a specified ` i` and an implied one are the same value, and a specified ` s` is indistinguishable from no flag at all. The factory now tells the selector it just built which flag was written, and `Modifier` reads that.

**Cost: none.** One `Byte` enum field on the internal `BaseAttrSelector`, which lands in existing padding: `AttrMatchSelector` measures **56 bytes/instance on `devel` and 56 with this change** (`GC.GetAllocatedBytesForCurrentThread`, 200k instances, Release/net10.0). One `is` test per selector *construction*, and nothing at all on the matching path — `Match` still reads the same `StringComparison` field it always did.

No public API change: `Creator`, `Register`, `Unregister`, `Create` and `CreateDefault` are untouched, and the single `_creators` table keeps its existing precedence.

Tests in `CssSelectorParserTests` cover the modifier that was not specified, the specified ` s` per operator, and what each serialization then selects.


FlorianRappl:
Important note (and that is part of the W3C spec as well): HTML (and therefore also CSS selectors when those are serialized) are not round-trip safe.

In general being round-trip safe is not a goal of the project (HTML is not as well). So I don't consider this a bug, but having round-trip capability (if it does not add cost) is an enhancement.

src/AngleSharp/Css/DefaultAttributeSelectorFactory.cs
```diff
@@ -72,6 +72,23 @@ public class DefaultAttributeSelectorFactory : IAttributeSelectorFactory, IAttri
             { CombinatorSymbols.Unlike, (name, value, prefix, mode) => new AttrNotMatchSelector(name, value, prefix, mode) },
         };
 
+        // The built-in selectors serialize themselves, so they need the modifier the author wrote
```
FlorianRappl: This is wrong. They don't need that. See my comment about round-trip capability.

src/AngleSharp/Css/DefaultAttributeSelectorFactory.cs
```diff
@@ -143,6 +161,11 @@ private ISelector CreateSelector(String combinator, String name, String value, S
                 _ => insensitiveAttributes.Contains(name),
             };
 
+            if (_builtInCreators.TryGetValue(combinator, out var builtIn))
```
FlorianRappl: Why do we have two lists here (instead of one) and when would `_creators` be hit (this means no match on `_builtInCreators`, but a match on `_creators`)?

lahma: Gone — there is a single `_creators` table again, with its original precedence. You're right that the second one made `_creators` unreachable for the built-ins; it existed only to carry one extra argument the public `Creator` delegate has no room for.

Instead the factory now sets the flag on the selector it just built (`if (selector is BaseAttrSelector attr) attr.DeclareCaseSensitivity(caseSensitivity);`), so `Creator`, `Register`, `Unregister` and the eight `Attr*Selector` constructors are all untouched. A consumer-registered creator returns something that is not a `BaseAttrSelector`, so it is unaffected either way.

src/AngleSharp/Css/DefaultAttributeSelectorFactory.cs
lahma: Deleted along with the table it described. The comment that replaces it justifies the flag by what CSSOM § 5.2 serializes, not by round-tripping.

lahma:
Fair — taking it as an enhancement. I've reworded the changelog line and the description accordingly and dropped the round-tripping argument.

What's left is that the serialization is wrong on its own terms: CSSOM § 5.2 appends the modifier only "if the attribute selector has the case-sensitivity flag present", and `[type="text"]` came back as `[type="text" i]` with no flag ever specified.

On cost: it is one `Byte` enum field on the internal `BaseAttrSelector`, and it lands in existing padding — `AttrMatchSelector` measures 56 bytes/instance on `devel` and 56 with this change (200k instances, `GC.GetAllocatedBytesForCurrentThread`, Release/net10.0). Nothing is added to the matching path; `Match` reads the same `StringComparison` field as before. The second creator table is gone too.

src/AngleSharp/Css/DefaultAttributeSelectorFactory.cs
FlorianRappl: Great - much appreciated!

FlorianRappl (approved):
LGTM!

PR #1352, opened 2026-09-18
Fix broken image download URLs in FAQ

RichardJansma:
The FAQ’s “How to download a picture?” section has two copyable C# examples that request `https://via.placeholder.com/150`. That endpoint closed the connection in my browser check, so the examples cannot download the sample image.

This changes only those two URL strings to `https://placeholder.photo/150.jpg`. I checked that it returns a 150×150 JPEG, matching the examples’ existing `image.jpg` output path. No runtime code or dependencies change.


FlorianRappl (approved):
LGTM - thanks for spotting!

FlorianRappl:
(Also thanks for providing that service - very helpful!)

